PR #17, opened 2015-02-06
Display the "part of" metadata

jamiecobbett:
I based this upon the code which currently does it in Whitehall:
https://github.com/alphagov/whitehall/blob/b1f827a2daa1df50cf69785af6827a621857e815/app/helpers/document_helper.rb#L205-L247

~~Note that the code in Whitehall has many more fields in it, but these aren't
currently sent to content-store. I think this may be based upon an assumption
that Case Studies only have the fields currently being sent to content-store.~~

~~Are there any other "part of" fields which do in fact appear for Case Studies?~~ 

Note: the tests are currently failing on an unrelated pre-existing failure that @danielroseman is working on.

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -19,7 +19,7 @@ class ContentItemPresenter
   end
 
   def part_of
-    []
+    links("related_policies") + links("worldwide_priorities") + links("world_locations")
   end
 
   def history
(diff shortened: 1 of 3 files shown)


edds:
:+1: Looks good assuming the failing tests are from the other problem

jamiecobbett:
The test failure is because government-frontend doesn't have support for the "coming_soon" format described in [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/tree/master/formats) - it will do, but not yet. This same test is currently failing on master: https://ci-new.alphagov.co.uk/view/Publishing/job/govuk_government_frontend/25/console

PR #18, opened 2015-02-09
Add Worldwide Organisations to "From" metadata

jamiecobbett:
Case Studies can have Worldwide Organisations, so we should display them.

The test failure is an unrelated pre-existing one already happening on master.


edds:
:+1:

PR #19, opened 2015-02-09
Improve README

jamiecobbett:
I followed our styleguide:
https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md


dsingleton:
:clap: big improvement.

Worth switching from `.rdoc` to `.md` at the same time? (matches the convention of other READMEs and I think GH will do better diffing with `md`)

jamiecobbett:
Renamed to `.md`. Test failure is due to an unrelated failure already happening on master.

PR #22, opened 2015-02-09
Only run contract tests for specific formats

heathd:
The government-frontend does not (yet) support the 'coming_soon' format.

Only attempt to run contract tests for formats which we currently
support.

As we evolve the contract testing approach, we may want to develop a
convention for frontends to declare which formats they support. For now
it seems reasonable to hard code it.

### DIFF
--- a/test/support/govuk_content_schema_examples.rb
+++ b/test/support/govuk_content_schema_examples.rb
@@ -41,6 +41,12 @@ module GovukContentSchemaExamples
 
   module ClassMethods
     def govuk_content_schema_examples
+      all_govuk_content_schema_examples.each_with_object({}) do |(filename, data), hash|
+        hash[filename] = data if supported_format?(data)
+      end
+    end
+
+    def all_govuk_content_schema_examples
       govuk_content_schema_example_files.each_with_object({}) do |file_path, hash|
         filename = File.basename(file_path)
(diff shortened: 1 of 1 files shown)


test/support/govuk_content_schema_examples.rb
```diff
@@ -41,6 +41,19 @@ def stub_example_content_items_in_content_store
 
   module ClassMethods
     def govuk_content_schema_examples
+      supported_formats = %w{
```
danielroseman: Probably should be `supported_examples` - not all the files here refer to distinct formats.

heathd: :+1:

tekin:
I think we have to be careful here as the hardcoding has been made against specific *examples* and not the *formats* themselves. This means that we've kind of broken the point of having the schema and examples as a way to verify that changes to formats do not break government-frontend. For example, if another "type" of case study is added to govuk-content-schema examples (say a case study for a previous administration), then that will not automatically be tested against government-frontend and a potentially breaking change will not be flagged. Therefore I think this change should work at the format level (i.e. case_study and unpublishing supported, but coming_soon not), and not at the level of individual example files.

heathd:
@tekin that makes sense, I'll modify it to restrict by format

PR #24, opened 2015-02-20
Move presenter tests to presenter tests folder

jamiecobbett:


alext:
Will these be picked up by the standard rake tasks?  (This isn't rspec...)

alext:
I just checked, and they don't get picked up...

On Master:
```
$ be rake
Run options: --seed 49300

# Running:

....................

Finished in 0.611270s, 32.7187 runs/s, 52.3500 assertions/s.

20 runs, 32 assertions, 0 failures, 0 errors, 0 skips
```

On this branch:
```
$ be rake
Run options: --seed 37255

# Running:

.........

Finished in 0.401127s, 22.4368 runs/s, 32.4087 assertions/s.

9 runs, 13 assertions, 0 failures, 0 errors, 0 skips
```

jamiecobbett:
@alext thanks - I've added a test task to run them and include them in the default test task.

alext:
I wonder whether it would be simpler to put these in `test/unit/presenters`.  That might make more sense given they're unit tests of the presenters.

Having said that, this is starting to get into bikeshedding, so I'm happy for this to me merged as is.  I'll let you make the call..

PR #26, opened 2015-02-23
Support Coming Soon content items

jamiecobbett:
Coming Soon content items are used for scheduled publishing. It is important in
scheduled publishing that all users see the new version as quickly as possible.
For content that is replacing existing content, this is easy: we simply wind in
the cache headers leading up to the publish time. For new content, this is
harder, because our stack doesn't respect the cache time for HTTP responses
with the 404 status. So instead, we return a 200 status and display a "Coming
Soon" page with the same cache header winding.

Whitehall:
screen shot 2015-02-23 at 15 08 30

government-frontend:
screen shot 2015-02-23 at 15 08 40

### DIFF
--- a/app/assets/stylesheets/views/_coming-soon.scss
+++ b/app/assets/stylesheets/views/_coming-soon.scss
@@ -1,0 +1,6 @@
+.coming-soon {
+  .summary {
+    @include core-24;
+    margin-bottom: $gutter-half;
+  }
+}
--- a/app/views/content_items/coming_soon.html.erb
+++ b/app/views/content_items/coming_soon.html.erb
@@ -1,0 +1,15 @@
+<%= content_for :page_class, "coming-soon" %>
+<%= content_for :title, "Coming soon - GOV.UK" %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title', title: 'Coming soon' %>
+
+    <p class="summary">
+      This document will be published on<br />
+      <%= content_tag(:time, class: 'publish_time', datetime: @content_item.iso8601_publish_time) do %>
+        <%= @content_item.formatted_publish_date + " at " + @content_item.formatted_publish_time %>
+      <% end %>
+    </p>
+  </div>
+</div>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -7,5 +7,6 @@
 
 // pages specific view imports
 @import "views/case-studies";
+@import "views/coming-soon";
 @import "views/unpublishing";
 
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -23,6 +23,7 @@ private
     case content_item['format']
       when 'case_study' then ContentItemPresenter.new(content_item)
       when 'unpublishing' then UnpublishingPresenter.new(content_item)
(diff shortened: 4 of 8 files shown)


app/assets/stylesheets/views/_coming-soon.scss
```diff
@@ -0,0 +1,4 @@
+.summary {
```
edds: This will effect all `.summary` elements in the application rather than just the coming soon ones. The way we try and do this is to use the `page_class` as the root selector in the view Sass file and then nest everything inside that selector. You can see this in [`views/_case-study.scss`](https://github.com/alphagov/government-frontend/blob/coming_soon/app/assets/stylesheets/views/_case-studies.scss) where the outermost selector is `.case-studies {` and all the page elements are nested within it. Hope that makes sense.

jamiecobbett: Yep, perfect sense, thanks. Rebased.

app/views/content_items/coming_soon.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= content_for :page_class, "coming-soon" %>
+<%= content_for :title, "Coming soon - GOV.UK" %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title', title: 'Coming soon' %>
+
+    <p class="summary">
+      This document will be published on<br />
+      <%= content_tag(:abbr, class: 'publish_time', title: @content_item.iso8601_publish_time) do %>
```
edds: Whitehall was using an older (pre HTML5) way to markup dates and times. You should be able to use the [time tag](http://www.w3.org/TR/html-markup/time.html) now. Which would make this line:

```erb
<%= content_tag(:time, class: 'publish_time', datetime: @content_item.iso8601_publish_time) do %>
```

edds:
Is the cache header handled in another part of the stack? Or am I missing something?

heathd:
it would be good to have a test for expiry time getting passed through (I guess this is why you were asking about integration tests earlier?)

danielroseman:
@edds it's done in the controller for all formats, see app/controllers/content_items_controller.rb#L6

app/views/content_items/coming_soon.html.erb
jamiecobbett: Nice. Does it make sense to duplicate the iso8601 datetime in a `title` attribute so that it still gets displayed on hover? Or is that something nobody is really interested in?

edds: The `title` was just so we followed the [microformat datetime pattern](http://microformats.org/wiki/datetime-design-pattern). We shouldn't need to add the title anymore.

jamiecobbett: OK, thanks for explaining it. Rebased.

edds:
:+1: Thank you for fixing those things @jamiecobbett

jamiecobbett:
Does it matter that the font sizing and bottom margin is different? Or is that desirable?

edds:
Ideally there would probably be more space at the bottom of the page. But I feel that is a concern of the report a problem button really.

edds:
The font-weight change is good though, the non-bold is also a historic thing.

jamiecobbett:
I've added a commit to add test coverage for expiry headers.

jamiecobbett:
I think this is ready to merge - any objections?

PR #27, opened 2015-02-25
Add the available languages block

danielroseman:
This hasn't gone into a shared component because there is a bit too much logic for a simple standalone template.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,8 +4,12 @@
 @import 'colours';
 @import 'measurements';
 @import 'typography';
+@import 'shims';
 
 // pages specific view imports
 @import "views/case-studies";
 @import "views/unpublishing";
 
+// HELPERS
+// Styles for comment page elements
+@import "helpers/available-languages";
--- a/app/views/shared/_available_languages.html.erb
+++ b/app/views/shared/_available_languages.html.erb
@@ -1,0 +1,15 @@
+<% if translations.length > 1 %>
+  <div class="available-languages">
+    <ul>
+      <% translations.each.with_index do |translation, i| %>
+        <li class="translation<%= ' last' if i == translations.length-1 %>">
+          <% if translation["locale"] == I18n.locale.to_s %>
+            <span><%= native_language_name_for(translation["locale"]) %></span>
+          <% else %>
+            <%= link_to native_language_name_for(translation["locale"]), translation["web_url"] %>
+          <% end %>
+        </li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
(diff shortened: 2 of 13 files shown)


test/helpers/available_languages_helper_test.rb
```diff
@@ -0,0 +1,56 @@
...
+    assert_equal ['es', 'ar'], locale_codes(non_english)
+  end
+
+  test "native_language_name_for returns native locale names" do
+    locales = ['en', 'es', 'ar'].map {|l| native_language_name_for(l)}
+    assert_equal ["English", "Español", "العربية"], locales
+  end
+
+
+  private
```
jamiecobbett: Our styleguide says that `private` should be outdented: https://github.com/alphagov/styleguides/blob/master/ruby.md#classes

danielroseman: Fixed.

app/views/shared/_available_languages.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if non_english_translations(translations).any? %>
```
edds: Not sure this is correct. There are times when we might have content which is translated into more than one language and neither of them will be english. I think this really just wants to check for more than one translation.

app/views/shared/_available_languages.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <div class="available-languages">
+    <ul>
+      <% sorted_locales(translations).each.with_index do |translation, i| %>
```
edds: Can the translations come back from the presenter already in the correct order?

app/views/shared/_available_languages.html.erb
danielroseman: This is the logic in Whitehall, which does actually prevent extra translations on documents where the main language is not English. However you're right there is no reason to restrict it that way here, so changed to simply check the number of translations.

app/views/shared/_available_languages.html.erb
danielroseman: Moved the bulk of the logic to the presenter.

app/assets/stylesheets/application.scss
```diff
@@ -4,8 +4,12 @@
...
 @import 'typography';
+@import 'shims';
 
 // pages specific view imports
 @import "views/case-studies";
 @import "views/unpublishing";
 
+// HELPERS
+// Styles for comment page elements
+@import "helpers/_available_languages";
```
edds: You don't need the leading underscore on the import line as Sass infers it as it is trying to load a partial file. The leading underscore just indicates to Sass that the file shouldn't be compiled by itself. It can just read `@import "helpers/available_languages";`.  To match the file naming convention of the other files in this project can the helper be named `helpers/_available-languages.scss` (with a dash for word separation).

edds:
One super minor style comment. Other than that this looks good to me :+1:. Thank you for making those changes the template.

app/assets/stylesheets/application.scss
danielroseman: @edds Done.

PR #28, opened 2015-03-04
Add government navigation

jamiecobbett:
~~DON'T MERGE UNTIL https://github.com/alphagov/static/pull/553 has been merged and deployed.~~

Government pages need the government navigation which appears in the banner.

I had to update Slimmer to get the improved test behaviour of components
which allows the test to check that the component had been called and had been
called with the correct parameters.

Screenshot with static on @tommyp's branch:

screen shot 2015-03-04 at 16 23 33


danielroseman:
:+1: Good to merge once the static one is deployed.

PR #34, opened 2015-03-27
Use 'long' title format.

danielroseman:
Fix issue with title size, noted in alphagov/whitehall#2032 (cc @dsingleton)


dsingleton:
:+1:

PR #35, opened 2015-03-27
Don't show caption if it is false.

danielroseman:
A large number of case studies contain the boolean value "false" as the image caption, and this was displaying underneath the picture. This change just hides that value; a separate change to Whitehall will be needed to ensure no value is sent to content-store if there is an empty caption.

### DIFF
--- a/app/views/content_items/case_study.html.erb
+++ b/app/views/content_items/case_study.html.erb
@@ -37,10 +37,10 @@
 
 <div class="grid-row">
   <div class="column-third">
-    <% if @content_item.image %>
+    <% if (image = @content_item.image) %>
       <figure class="sidebar-image">
-        <img src="<%= @content_item.image["url"] %>" alt="<%= @content_item.image["alt_text"] %>">
-        <figcaption><%= @content_item.image["caption"] %></figcaption>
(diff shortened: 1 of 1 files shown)


alext:
Doh!...

Is this worth a test?

app/views/content_items/case_study.html.erb
```diff
@@ -40,7 +40,7 @@
     <% if @content_item.image %>
       <figure class="sidebar-image">
         <img src="<%= @content_item.image["url"] %>" alt="<%= @content_item.image["alt_text"] %>">
-        <figcaption><%= @content_item.image["caption"] %></figcaption>
+        <figcaption><%= @content_item.image["caption"].presence %></figcaption>
```
edds: This should really be a guard around the `<figcaption>` if there is no caption rather than rendering an empty one.

danielroseman: Done in ffb6977.

jamiecobbett:
I think I fixed the source of the bad data here: https://github.com/alphagov/whitehall/pull/2026. We discussed republishing all affected editions, but decided not to, as we would republish everything soon anyway.

PR #37, opened 2015-04-17
Show placeholder image for items without images

vinayvinay:
https://trello.com/c/VzKdQ6xm

Whitehall doesn't send image_data for a content item if:

- the edition has no image assigned, and
- the edition's organisation also doesn't have a default image

see: alphagov/whitehall@ab30110

this allows the frontend app to decide what to do in such a scenario.
we're choosing to display a placeholder image.

this required changing the content item presenter to have `view_context`.

we should prefer this method for using view helper methods over
mixing-in related modules in the presenter. [asset URL helper
methods](https://github.com/rails/rails/blob/v4.1.5/actionview/lib/action_view/helpers/asset_url_helper.rb#L194) depend on `view_context` to access configuration and
request information, for example request protocol.

as a result, `*_url` methods don't work in a presenter when
mixed-in, because they don't have access to `config` where it
looks for the asset host. without knowing asset_host it returns
a path instead of returning the full URL, and to make it worse,
does that silently.

hat-tip: @heathd

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -21,7 +21,7 @@ private
 
   def present(content_item)
     case content_item['format']
-      when 'case_study' then ContentItemPresenter.new(content_item)
+      when 'case_study' then ContentItemPresenter.new(content_item, view_context)
       when 'unpublishing' then UnpublishingPresenter.new(content_item)
       when 'coming_soon' then ComingSoonPresenter.new(content_item)
       else raise "No support for format \"#{content_item['format']}\""
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -55,7 +55,7 @@ class ContentItemPresenter
   end
 
   def image
-    content_item["details"]["image"]
+    content_item["details"]["image"].presence || placeholder_image_data
   end
 
   def archived?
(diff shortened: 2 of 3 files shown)


tekin:
One alternative to passing in the view context would be to move the logic into the template:

```ruby
<% image = @content_item.image || placeholder_image %>
```
Where `placeholder_image` is a view helper that returns an appropriate data structure. This feels to me like putting the responsibility for deciding what is shown to a user when there is no image associated with a piece of content in the right place - the view. That said, the line is fairly blurred between responsibilities  between presenters and views. I'm just not a fan of passing in the view context to presenters as it introduces a fairly hairy piece of coupling.

edds:
I would probably go even further than @tekin's suggestion and probably just change the view to be:

```erb
<% if (image = @content_item.image) %>
  <figure class="sidebar-image">
    <img src="<%= image["url"] %>" alt="<%= image["alt_text"] %>">
    <% if image["caption"].present? %><figcaption><%= image["caption"] %></figcaption><% end %>
  </figure>
<% else %>
  <div class="sidebar-image"><%= image_tag 'placeholder.jpg', alt: "" %></div>
<% end %>
```

That makes it really clear as to where the application concerns (the placeholder) and the content store concerns (the case study with provided image) lay. It also means that you would be able to not use a figure element when just using a decorative image as oppose to an image which will relate to the content. This should also be indicated by the empty `alt` attribute.

vinayvinay:
you're right, the line is fairly blurred; there's view-helper stuff `content_tag` and `link_to` already being used there. however, i like the idea of content item presenter being one place where most stuff relating to presenting the content item can be found. another thing to bear in mind, this logic will be common across different formats. i'll wait for more inputs on this one to decide.

vinayvinay:
just saw @edds's comment. if we want to implement his suggestion, we'll definitely have to move the logic out from the presenter. i'll make that change. thanks for your inputs!

vinayvinay:
replaced by https://github.com/alphagov/government-frontend/pull/38

PR #38, opened 2015-04-20
Show placeholder image for items without image data

vinayvinay:
https://trello.com/c/VzKdQ6xm

Whitehall doesn't send image_data for a content item if:

- the edition has no image assigned, and
- the edition's organisation also doesn't have a default image

see: alphagov/whitehall@ab30110

this allows the frontend app to decide what to do in such a scenario.
we're choosing to display a placeholder image. the approach to do so
is decided based on suggestions on: #37.


tekin:
:+1:

edds:
This LGTM. Just need to add the placeholder image.

vinayvinay:
@edds: i was under the impression that because we have [configured an asset server](https://github.com/alphagov/government-frontend/blob/master/config/environments/production.rb#L37) it will work without adding the placeholder image to this repository. i don't completely understand how assets are served, so it'll be helpful if you could explain a bit further why we need the image here.

edds:
@vinayvinay setting an asset server ensures that all asset urls are pointing at our asset CDN. On top of that we also set [an asset prefix](https://github.com/alphagov/government-frontend/blob/8276822218e88839d5dc8bfd29bb1b6a9bb2c935/config/application.rb#L39) which prefixes the path for assets to be within the `government-frontend` folder. The configuration of our CDN and servers means that the assets in that folder will be the assets generated from this application (the minified CSS/JS and images which end up in the public folder when running `rake assets:precompile`). We try where possible to keep all apps responsible for all the specific assets they are going to need. This lets us have confidence in removing assets when they are no longer used. So in this case this application will need to provide its own placeholder image to display in the sidebar.

vinayvinay:
@edds: thanks for the explanation, i've added the placeholder.jpg, same as Whitehall, and rebased.

edds:
Thank you for persisting with this. :+1:

PR #41, opened 2015-05-20
Prevent exception if item has no first_public_at.

danielroseman:
Draft ContentItems that have never been published will not have a
first_public_at date in their details; previously the presenter was
choking in this case.

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -87,7 +87,11 @@ private
   end
 
   def any_updates?
-    DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(content_item["details"]["first_public_at"])
+    if (first_public_at = content_item["details"]["first_public_at"]).present?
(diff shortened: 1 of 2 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -87,7 +87,11 @@ def display_time(timestamp)
   end
 
   def any_updates?
-    DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(content_item["details"]["first_public_at"])
+    if (first_public_at = content_item["details"]["first_public_at"]).blank?
+      false
+    else
+      DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
+    end
```
alext: This conditional feels backwards (you're using the assignment in the `if` in the `else` block.  It might be better to turn this around, and use `.present?` instead of `.blank?`

danielroseman: Good point, updated.

PR #51, opened 2015-06-29
Pass on 403 errors received from content-store.

danielroseman:
Catches the HttpForbidden error that is now raised by gds-api-adapters,
and renders a 403 response.


.gitignore
```diff
@@ -14,3 +14,5 @@
 # Ignore all logfiles and tempfiles.
 /log/*.log
 /tmp
+
+/public/government-frontend
```
alext: Did you mean to add this?  It seems unrelated to this PR.

app/controllers/content_items_controller.rb
```diff
@@ -9,6 +9,8 @@ def show
     else
       render text: 'Not found', status: :not_found
     end
+  rescue GdsApi::HTTPForbidden => e
+    render text: e.message, status: 403
```
alext: This might be better handled with a `rescue_from` call at controller level.  See https://github.com/alphagov/collections/blob/master/app/controllers/subtopics_controller.rb#L6 for an example (in that example, `error_404` is a method defined in `ApplicationController`).

.gitignore
danielroseman: Sorry, removed.

app/controllers/content_items_controller.rb
danielroseman: Done.

PR #55, opened 2015-08-26
Add analytics meta tags component

edds:
So that the correct analytics data is sent add the analytics meta tag
component.


app/views/layouts/application.html.erb
```diff
@@ -7,6 +7,7 @@
   <!--[if IE 7]><%= stylesheet_link_tag "application-ie7" %><script>var ieVersion = 7;</script><![endif]-->
   <!--[if IE 8]><%= stylesheet_link_tag "application-ie8" %><script>var ieVersion = 8;</script><![endif]-->
   <%= csrf_meta_tags %>
+  <%= render partial: 'govuk_component/analytics_identifiers', locals: { content_item: @content_item.content_item } %>
```
jamiecobbett: I think this should be `analytics_meta_tags`? `analytics_identifiers` being the original name in the branch. https://github.com/alphagov/static/pull/640/files

PR #56, opened 2015-08-26
Update name of component

edds:
The component name has been changed to make it slightly more generic.
This needs reflecting here.


jamiecobbett:
Tested successfully in dev :+1:

PR #58, opened 2015-10-14
Support new `service_manual_guide` format for content items

AndrewVos:
This is feature flagged in at the moment, so as not to break production.


jamiecobbett:
Are you planning to add a content schema? https://github.com/alphagov/govuk-content-schemas

KushalP:
@jamiecobbett there's a work in progress branch here: https://github.com/alphagov/govuk-content-schemas/compare/service_manual_publisher

We're going to wait until we understand a little more about the schemas we need before submitting the PR.

PR #61, opened 2015-10-26
Service manual - "Guide page" - initial frontend work

gemmaleigh:
This pull request:

- Uses the classname `.service-manual-guide` to prevent styles affecting other formats.

- Creates new components
  - a breadcrumb `.breadcrumb`
  - a contents list `.contents-list`

- Duplicates existing GOV.UK components and modifies them
  - metadata `.metadata`
  - page title `.page-title`

- Creates new styles for the main content area - headings, lists and paragraph text.
  - page content `.prose`

This PR uses ems for margin and padding, so these values will change with the values set by the typography mixins.

A consistent naming convention is used for child elements of components, to avoid "leaky" styles,  `.componentname-childelementname`.


app/assets/stylesheets/views/_service-manual-guide.scss
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+  }
+}
+
+
+// Modify existing component: 'govuk_component/title'
+
+.govuk-title {
+  margin-top: 1.875em; // 30/16
```
alextea: Could you just have `30/16` as the value here, removing the need for the comment?

gemmaleigh: Absolutely, I was just trying to make it clear where the em values had come from.

Ideally, I'd like to use a [px-to-em function, like the one in elements](https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_helpers.scss#L16), I'll create a PR to add this to the front end toolkit.

// Return ems from a pixel value
@function em($px, $base: 19) {
  @return ($px / $base) + em;
}

So instead it would be `margin-top: em(30,16);`

app/views/content_items/service_manual_guide.html.erb
```diff
@@ -1,11 +1,73 @@
...
+        Topic
+      </p>
+      <h1>
+        <%= @content_item.title %>
+      </h1>
+    </div>
+
+  </div>
+  <div class="column-third">
+    <div class="govuk-community-contact">
```
dsingleton: I not sure this is worth turning into a component - and we shouldn't re-use the naming convention for something that isn't.

dsingleton:
Just to clarify, is this a prototyping / discussion PR, or to be merged?

dsingleton:
:+1: to prototyping component additions / updates within the app, but it'd be good to start seeing them tested  `static` / the component guide, and most importantly against other GOV.UK Frontend apps

Especially for things like the margin additions, which we could test and roll out separately.

app/views/content_items/service_manual_guide.html.erb
gemmaleigh: This already is a component, I was going to use the existing component here: http://govuk-component-guide.herokuapp.com/components/title/fixtures/with_context/preview. 

The only suggestion here was to change .context to .govuk-title-context.

gemmaleigh:
It would be good to get this merged.

I hadn't planned to update the component guide just yet (as this will affect other GOV.UK front end apps) - but have commented where the service manual guide view either modifies an existing component, or creates a new one.

tadast:
> Just to clarify, is this a prototyping / discussion PR, or to be merged at some point?

@dsingleton This is [feature-flagged](https://github.com/alphagov/government-frontend/pull/58/files) and we'd like to get this PR merged so that non-devs can play around with publisher and frontend on preview.

bradwright:
This functionality, and the feature flag functionality, don't have any tests.

As we discussed, this is a production application, and it needs to have production code in it, which means tests at the very least. It should also mean that any lingering `TODOs` are removed.

We also discussed deploying branches instead of master as a way around this, so I'm not sure I understand what the urgency is to have this merged.

dsingleton:
> @dsingleton This is feature-flagged and we'd like to get this PR merged so that non-devs can play around with publisher and frontend on preview.

I don't think I have the context on what this work is trying to achieve to be able to review, but from an outside perspective it seems like a building a prototype inside a production application, which isn't clearly indicated as being a prototype.

As @bradwright mentioned, it's possible to deploy banches to preview, or, tbh if it's easy to talk to the draft content store, then this could be run as a [prototype app on heroku](https://github.com/dsingleton/govuk-prototyping-rails).

I'm wary of mixing prototype behaviours in to a production app, where they overlap with production styles, and effectively forking components and breaking the abstraction. I'm not keen on this being merged in it's current form.

I'm not sure github/text is the best way to discuss this. It might be better to get everyone together offline.

app/assets/stylesheets/views/_service-manual-guide.scss
KushalP: That PR is now merged and can be rolled into this.

app/assets/stylesheets/views/_service-manual-guide.scss
```diff
@@ -0,0 +1,145 @@
...
+
+.govuk-community-contact {
+  margin-top: 1.875em; // 30/16
+}
+
+
+// Modify existing component: 'govuk_component/metadata'
+
+.govuk-metadata {
+  margin-bottom: 2em;
```
dsingleton: This is going to leak into Case Study pages, affecting their components.

| before | after |
| --- | --- |
|  screenshot 2015-10-28 11 16 31  | screenshot 2015-10-28 11 16 27 |

They'll potentially be other components this leaks into, eg, margins for `govuk-title`

app/views/content_items/service_manual_guide.html.erb
dsingleton: This was a comment on `govuk-community-contact`

KushalP:
> Just to clarify, is this a prototyping / discussion PR, or to be merged at some point?

It's to be merged in, but we welcome discussion or ways to improve it :smile: 

> As we discussed, this is a production application, and it needs to have production code in it, which means tests at the very least. It should also mean that any lingering TODOs are removed.

Commits 1b8ed26 and 43297f5 from @AndrewVos add tests around the feature flag functionality. The aim is to merge this PR into master rather than deploy against a branch. This work is the start of taking what exists as a prototype into something that can run in production.

> I'm not sure github/text is the best way to discuss this. It might be better to get everyone together offline.

Agreed. I'll try to grab some of your time today.

dsingleton:
> Commits 1b8ed26 and 43297f5 from @AndrewVos add tests around the feature flag functionality. The aim is to merge this PR into master rather than deploy against a branch. This work is the start of taking what exists as a prototype into something that can run in production.

From a frontend perspective, to go into production, this is going need;

- A separate CSS file, to avoid any risk of service manual CSS affecting production formats (as is currently happening, and would be easy to reintroduce, even if fixed)
- Renaming any component-like classes (to something `govuk-prototype-title`) and duplicating CSS locally (rather than hot-linking static) so as not to violate the abstraction of components, and make clear they're not actually components.
- Adding [asset precompile tests](https://github.com/alphagov/frontend/blob/master/jenkins.sh#L29) to `jenkins.sh` (optional extra)

I think those are the main blockers for me.

KushalP:
> Adding asset precompile tests to jenkins.sh (optional extra)

Good spot. Got a PR for that here: https://github.com/alphagov/government-frontend/pull/62

tadast:
I've got some more changes related to this branch (replacing placeholder data with real one) https://github.com/alphagov/government-frontend/compare/service-manual-guide-frontend...service-manual-guide-frontend-data should I wait till this gets merged or put them in this PR?

KushalP:
Chatted with @tadast about this. He's going to wait for @gemmaleigh to finish with this PR and then submit that afterwards.

KushalP:
>  A separate CSS file, to avoid any risk of service manual CSS affecting production formats (as is currently happening, and would be easy to reintroduce, even if fixed)

The use of a top-level `.service-manual-guide` class prefix achieves this separation.

I'm happy with the rest of the changes in this PR. @dsingleton can you do a once-over please?

PR #63, opened 2015-10-29
Use real data when rendering service manual frontend guides

tadast:
We had a couple of placeholders on the guides page. Real data is being published now, so we can display it on the page.


KushalP:
There was one small thing I picked up about how general `Publisher` is, but the rest of this looks good.

Once you've changed that, I'm happy with this PR.

tadast:
@KushalP I have renamed the field and added another commit to show `<title>`. The build should fix itself once we merge https://github.com/alphagov/govuk-content-schemas/pull/150

PR #64, opened 2015-11-03
Add phase banner to service manual guide page

gemmaleigh:
- Import alpha beta partial from the front end toolkit
- Use the phase banner mixin from the front end toolkit
- Use placeholder for the Beta banner

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -5,6 +5,7 @@
 @import 'measurements';
 @import 'typography';
 @import 'shims';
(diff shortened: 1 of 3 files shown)


jamiecobbett:
Can't whether this is shown be determined by `phase` attribute in the content item?

jamiecobbett:
Also, it should probably use the [beta banner govuk_component](https://github.com/alphagov/static/blob/master/app/views/govuk_component/beta_label.raw.html.erb).

gemmaleigh:
Hi @jamiecobbett, of course. I'll close this and use the component instead.

PR #66, opened 2015-11-10
Render phase labels

tadast:
If a content item responds to `#phase` and it has a value of either 'alpha' or 'beta' render a component with the corresponding label. ~~Currently only Service Manual Guide presenter responds to `#phase` so only this format will be affected. I chose to not expose `phase` in ContentItemPresenter (used by Case Studies atm), because I don't know if the team that owns this format expects the `phase` field value to trigger this feature.~~

~~I'm not particularly happy with the helper test: Rails helpers fail when `render` is invoked on them in test environment. The default Rails test does not have tools for stubbed expectations, so I ended up overwriting `render` method in the test class to serve my purpose.~~

~~Alternatives I can think of: do not use a helper, just have a conditional in the application layout. Test setup would be a bit more difficult for this alternative though.~~

~~Any other ideas?~~

Update: see updated commit and its note.

Closes #65 

/cc @jamiecobbett

### DIFF
--- a/test/integration/phase_label_test.rb
+++ b/test/integration/phase_label_test.rb
@@ -1,0 +1,44 @@
+require 'test_helper'
+
+class PhaseLabelTest < ActionDispatch::IntegrationTest
+  test "Beta phase label is displayed for a Service Manual Guide in phase 'beta'" do
+    ENV["FLAG_ENABLE_SERVICE_MANUAL"] = 'yes'
+    guide_sample = JSON.parse(GovukContentSchemaTestHelpers::Examples.new.get('service_manual_guide', 'basic_with_related_discussions'))
+    guide_sample.merge!("phase" => "beta")
+    content_store_has_item("/service-manual/agile", guide_sample.to_json)
+
+    visit "/service-manual/agile"
+
+    assert_has_phase_label "beta"
+    ENV.delete("FLAG_ENABLE_SERVICE_MANUAL")
+  end
+
+  test "Alpha phase label is displayed for a Case Study in phase 'alpha'" do
+    case_study = JSON.parse(GovukContentSchemaTestHelpers::Examples.new.get('case_study', 'case_study'))
+    case_study.merge!("phase" => "alpha")
+    content_store_has_item("/government/case-studies/get-britain-building-carlisle-park", case_study.to_json)
+
+    visit "/government/case-studies/get-britain-building-carlisle-park"
+
+    assert_has_phase_label "alpha"
+  end
+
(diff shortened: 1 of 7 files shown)


jamiecobbett:
The implementation is exactly what I was hoping for :+1: 

Regarding the testing, what about making a controller or integration/acceptance test and stubbing out the request to content-store?

tadast:
To write a functional/acceptance test we'd need to stub a lot more than that to do it nicely i.e. avoid using any existing controllers, presenters and views. Unless we say it's cool to have a test for Service Manual Guide only.

dsingleton:
> I chose to not expose phase in ContentItemPresenter (used by Case Studies atm), because I don't know if the team that owns this format expects the phase field value to trigger this feature.

We'll move `phase` to `ContentItemPresenter` at some point. I think it'd be a quick conversation with Core (who own case-studies) to warn/agree, then we put it there from the start.

Otherwise, LGTM :+1:

tadast:
Closing this with an intention to reopen with an improved version of code

dsingleton:
@tadast and I had a chat offline and agreed it would be better if the phase logic was usable by any ContentItem, not just service manual content, as all ContentItems can have a phase and we'll calling it for the whole app, not just service manual content.

We agreed `ContentItemPresenter` should have the phase logic with `ServiceManualGuidePresenter` should inherit from it. We also discussed a rule-of-thumb that `ContentItemPresenter` should deal all base fields (eg, anything not in `details`). `ContentItemPresenter` already has some exceptions to this for Case Studies, that aren't worth moving out yet - Core will take a look at that once we start on World Location News, and with more examples we'll have a better idea what a good presenter structure looks like, but in the mean time having `ContentItemPresenter` provide the base fields seems like a good step.

tadast:
@dsingleton @jamiecobbett have a look at updated code please.

test/integration/phase_label_test.rb
```diff
@@ -0,0 +1,44 @@
+require 'test_helper'
+
+class PhaseLabelTest < ActionDispatch::IntegrationTest
+  test "Beta phase label is displayed for a Service Manual Guide in phase 'beta'" do
+    ENV["FLAG_ENABLE_SERVICE_MANUAL"] = 'yes'
```
jamiecobbett: This would be more graceful and less fragile using [ClimateControl](https://github.com/thoughtbot/climate_control) - what do you think?

tadast: It would. But it's a temporary feature flag, is it really worth adding a new gem dependency for that?

PR #67, opened 2015-11-18
Base presenter

tadast:
Content item presenter was used to render `CaseStudy` format and had a lot of logic that's specific to this format. In order to make `ContentItemPresenter` useful for rendering other content types, I've extracted generic functionality to `ContentItemPresenter` and moved all case-study-specific functionality to a new `CaseStudyPresenter` class. Then I've made other presenters inherit from `ContentItemPresenter` to remove logic duplication.

__Note: this affects case studies, "unpublishing" pages, coming soon pages and service manual guides.__ It would be nice to thoroughly review the changes in case I've broken anything (I'm not familiar with formats other than Service Manual Guide).

This refactoring has been triggered by a discussion in #66. Once this is merged I'll reintroduce the change there that benefits from this refactoring.

/cc @dsingleton @jamiecobbett

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -1,80 +1,16 @@
 class ContentItemPresenter
-  include ActionView::Helpers::UrlHelper
-
-  attr_reader :content_item, :title, :description, :body, :format, :format_display_type, :locale
+  attr_reader :content_item, :title, :description, :format, :locale
 
   def initialize(content_item)
     @content_item = content_item
-
     @title = content_item["title"]
     @description = content_item["description"]
-    @body = content_item["details"]["body"]
     @format = content_item["format"]
-    @format_display_type = content_item["details"]["format_display_type"]
     @locale = content_item["locale"] || "en"
-  end
-
-  def from
-    links("lead_organisations") + links("supporting_organisations") + links("worldwide_organisations")
-  end
-
-  def part_of
-    links("document_collections") + links("related_policies") + links("worldwide_priorities") + links("world_locations")
   end
 
   def available_translations
     sorted_locales(@content_item["links"]["available_translations"])
-  end
-
-  def history
-    return [] unless any_updates?
-    content_item["details"]["change_history"].map do |item|
-      {
-        display_time: display_time(item["public_timestamp"]),
-        note: item["note"],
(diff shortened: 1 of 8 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -1,104 +1,22 @@
...
-        time: content_tag(:time, display_time(notice["withdrawn_at"]), datetime: notice["withdrawn_at"]),
-        explanation: notice["explanation"]
-      }
-    end
-  end
-
 private
 
   def sorted_locales(translations)
     translations.sort_by { |t| t["locale"] == I18n.default_locale.to_s ? '' : t["locale"] }
```
jamiecobbett: Should `translations` be `available_translations`?

tadast: `available_translations` are passed in as a parameter named `translations`, so it's all cool I think. 

Not sure why they are passed in as a parameter though: it's a private method and the object already knows how to access this data. Perhaps I'll add another commit to address this.

tadast: I've tried merging `#available_locales` and `#sorted_locales` methods into one locally, but looking at the new version I don't think it's an improvement readability-wise. I'll leave it like this for now.

app/presenters/content_item_presenter.rb
```diff
@@ -1,104 +1,22 @@
...
-
-  def any_updates?
-    if (first_public_at = content_item["details"]["first_public_at"]).present?
-      DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
-    else
-      false
-    end
-  end
-
-  def links(type)
```
boffbowsh: This feels like it'll be something that's common to other content types, not just case studies. Perhaps it should at least be a mixin?

jamiecobbett: I can see what you're getting at, but I'm a bit wary of that because the name is so generic. I had to read the code closely to figure out what it returns (an array of HTML anchor elements). 

What if we renamed the method to `link_tags`? Then it is clear what it returns and it doesn't clash with the top-level property of the content item.

tadast: I agree with @jamiecobbett, but I would like to have a separate PR for that.

PR #68, opened 2015-12-16
Enable service manual guide rendering, remove feature flag

tadast:
We're deploying service manual publisher to production and it no longer makes
sense to have this feature-flagged.

We're not planning to publish any content yet, but we will need preview
functionality on the draft stack.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -22,14 +22,11 @@ private
   end
 
   def present(content_item)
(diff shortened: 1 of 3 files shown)


AndrewVos:
lgtm

AndrewVos:
:ship:

fofr:
The service manual publisher readme has a reference to this flag which should probably be removed too:
https://github.com/alphagov/service-manual-publisher/blob/master/README.md#development

tadast:
Well spotted! Will remove it now.

PR #69, opened 2016-01-19
Render "Take part" format

fofr:
https://trello.com/c/P1IYhe0C/242-take-part-migrate-front-end-work-large

* Add Take Part template, styles and presenter
  * Use `format` rather than `format_display_type` for gathering title context
  * Copy markup patterns from case study where similar/the same
* Don’t port the strange “underline” take part headings, the take part contents are purposefully left out, and the description has been moved above the image to match other formats.
* Extract styles and markup common to Case Studies and Take Part into partials and mixins
* When an image is omitted it will show a generic government placeholder image

# Before
screen shot 2016-01-19 at 16 22 09

# After
screen shot 2016-01-19 at 15 31 13

cc @dsingleton @boffbowsh @benlovell @gpeng

### DIFF
--- a/app/views/shared/_sidebar_image.html.erb
+++ b/app/views/shared/_sidebar_image.html.erb
@@ -1,0 +1,10 @@
+<% if image %>
+  <figure class="sidebar-image">
+    <img src="<%= image["url"] %>" alt="<%= image["alt_text"] %>">
+    <% if image["caption"].present? %><figcaption><%= image["caption"] %></figcaption><% end %>
+  </figure>
+<% else %>
+  <div class="sidebar-image">
+    <%= image_tag 'placeholder.jpg', alt: '' %>
+  </div>
+<% end %>
--- a/test/presenters/take_part_presenter_test.rb
+++ b/test/presenters/take_part_presenter_test.rb
@@ -1,6 +1,21 @@
 require 'test_helper'
 
 class TakePartPresenterTest < ActiveSupport::TestCase
-  include ActionView::Helpers::UrlHelper
 
+  test 'presents the basic details of a content item' do
+    assert_equal take_part['description'], presented_take_part.description
+    assert_equal take_part['format'], presented_take_part.format
+    assert_equal take_part['locale'], presented_take_part.locale
+    assert_equal take_part['title'], presented_take_part.title
+    assert_equal take_part['details']['body'], presented_take_part.body
+  end
+
+private
+  def presented_take_part(overrides = {})
+    TakePartPresenter.new(take_part.merge(overrides))
+  end
+
+  def take_part
+    govuk_content_schema_example('take_part', 'take_part')
+  end
 end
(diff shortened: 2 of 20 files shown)


boffbowsh:
:+1: for ones and zeroes, I'll let someone else review the shapes and colours

dsingleton:
LGTM.

*Note, @fofr and I discussed this offline*: As we add more formats we're going to need a pattern to avoid duplication of presenters, templates, and SASS. I'm not sure the approach of composable partials (and shared SASS mixins) will be right long term, but it will work for this format and the next few simple ones.

As we add more formats we'll have a better idea of the duplication pattern and can re-evaluate the approach. But in the mean time i'm happy to press on with this approach to get us moving :+1:

test/presenters/take_part_presenter_test.rb
```diff
@@ -0,0 +1,23 @@
+require 'test_helper'
+
+class TakePartPresenterTest < ActiveSupport::TestCase
+  include ActionView::Helpers::UrlHelper
```
benlovell: Is this needed?

test/presenters/take_part_presenter_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  test 'presents the basic details of a content item' do
+    assert_equal take_part['description'], presented_take_part.description
+    assert_equal take_part['format'], presented_take_part.format
+    assert_equal take_part['locale'], presented_take_part.locale
+    assert_equal take_part['title'], presented_take_part.title
+    assert_equal take_part['details']['body'], presented_take_part.body
+  end
+
+private
+
```
benlovell: We should drop this line per the style guide.

test/presenters/take_part_presenter_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  def presented_take_part(overrides={})
```
benlovell: Can you add spaces around the `=`, per the style guide?

I am quickly becoming the human embodiment of our style guide. :blush:

app/views/shared/_sidebar_image.html.erb
```diff
@@ -0,0 +1,11 @@
+<% image ||= false %>
```
benlovell: `nil` is falsey, so I think this may be unnecessary.

test/presenters/take_part_presenter_test.rb
fofr: You're right, removed and rebased.

test/presenters/take_part_presenter_test.rb
fofr: Updated this test file and the existing case study ones to match style guide.

test/presenters/take_part_presenter_test.rb
fofr: Done :+1:

app/views/shared/_sidebar_image.html.erb
fofr: I've removed this line. I think the problem this solves is if `image` isn't passed in as a local variable. Given this is a sidebar_image partial, that shouldn't ever happen.

benlovell:
:sparkler: :heart_eyes_cat:

PR #71, opened 2016-01-21
Use govspeak component for video embed

gemmaleigh:
- Move the govspeak styles for the service manual to a new partial
- Rename the prose text wrapper

This enables the govspeak youtube embedding to work, while maintaining the styles for service manual content pages.


gemmaleigh:
Closing for now, as this overrides other govspeak components in a way we don't want. Will re-open later.

PR #72, opened 2016-01-21
Enable videos to be embedded in Service Manual guide pages

gemmaleigh:
This functionality already exists as part of the govspeak component.

Use the `.govuk-govspeak` class to ensure that video links are expanded to the accessible video player. 

Also add a modifier, `.govuk-govspeak--service-manual` to set consistent spacing between HTML elements within the govspeak wrapper. 

cc. @tadast

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -23,5 +23,6 @@
   @import "components/page-title";
   @import "components/metadata";
   @import "components/page-contents";
+  @import "components/govspeak-service-manual";
   @import "views/service-manual-guide";
 }
--- a/app/assets/stylesheets/components/_govspeak-service-manual.scss
+++ b/app/assets/stylesheets/components/_govspeak-service-manual.scss
@@ -1,46 +1,51 @@
 .govuk-govspeak--service-manual {
 
-  margin-top: 1em;
-  margin-bottom: 2em;
+  padding-top: 1em;
+
+  margin-bottom: (30/19)+em;
   @include media(tablet) {
-    margin-top: 2em;
-    margin-bottom: 4em;
+    margin-bottom: (45/19)+em;
+  }
+
+  > h2:first-of-type {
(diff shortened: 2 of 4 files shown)


tadast:
Looks good to me :+1:

bradwright:
This doesn't appear to actually be using the component, but rather re-using the CSS class.

Correct usage here: http://govuk-component-guide.herokuapp.com/components/govspeak

but seems like:

```erb

<%= render partial: 'govuk_component/govspeak', locals: {
  content: @content_item.body.html_safe
} %>
```

is actually what you want.

tadast:
Using the component does not let us add an extra CSS class and the existing styling does not match our needs (see `.govuk-govspeak--service-manual`). We could use a component and wrap it in an extra div + change the CSS selectors to be more specific, but it's also not ideal.

bradwright:
I feel like it'd be better to extend the component to meet your needs (ie adding an extra class) than do what you're doing here. You've not future-proofed the implementation here.

dsingleton:
The duplication of the component was originally at my request, vs adding application CSS to override component CSS, which violates the component abstractions (setting bad precedent) and risk affecting production formats in government-frontend.

This was agreed with @KushalP when service manual was initial added to government frontend as a prototype, as it's clearly separated the prototype/exploratory work from production formats and components.

At the time, and best I can tell now, the desired govspeak changes weren't easy to integrate with components, as they are intentionally different from what components have defined. We really should be standardising the service manual to match the rest GOV.UK, or clearly defining an extension to the govspeak component, as `rich_govspeak` hints at.

dsingleton:
(I think this needs a face to face conversation to resolve, and to establish if this is still a prototype embedded in a production app)

fofr:
Will service manual be moved out of government frontend?

gemmaleigh:
Closing as #77 fixes this, we're using the govspeak component as-is. 

I'll create a separate PR to increase the spacing within the govspeak component.

PR #73, opened 2016-01-22
Make app work on Heroku

boffbowsh:
Set ruby version in Gemfile
Fix ruby versions for binstubs
Log to stdout
Enable debug logging
Serve static assets if set

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -20,6 +20,7 @@ Rails.application.configure do
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
+  # `RAILS_SERVE_STATIC_ASSETS` is an heroku-specific environment variable
   config.serve_static_assets = ENV["RAILS_SERVE_STATIC_ASSETS"] =~ /(true|yes)/i
 
   # Compress JavaScripts and CSS.
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -20,7 +20,7 @@ Rails.application.configure do
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = false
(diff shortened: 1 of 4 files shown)


config/environments/production.rb
```diff
@@ -20,7 +20,7 @@
   # config.action_dispatch.rack_cache = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this).
-  config.serve_static_assets = false
+  config.serve_static_assets = ENV["RAILS_SERVE_STATIC_ASSETS"] =~ /(true|yes)/i
```
fofr: It feels like this line and the ones below could benefit from a couple of comments explaining purpose. Without the Heroku context they might be a bit opaque.

config/environments/production.rb
```diff
@@ -81,7 +81,11 @@
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
-  config.logstasher.enabled = true
-  config.logstasher.logger = Logger.new("#{Rails.root}/log/#{Rails.env}.json.log")
-  config.logstasher.suppress_app_log = true
+  if ENV["DYNO"].present?
```
benilovj: If this app used 12 factor logging [like Publishing API](https://github.com/alphagov/publishing-api/blob/master/config/environments/production.rb#L53-L57), this conditional shouldn't be necessary. It'd be good to move to this approach anyway, as per [the accepted RFC](https://gov-uk.atlassian.net/wiki/display/TECH/RFC+26%3A+12-Factor+Rails+Apps).

benilovj: I've created a PR for this: https://github.com/alphagov/government-frontend/pull/80

config/environments/production.rb
benlovell: Addressed in 3ed2427

benlovell:
:+1: on @boffbowsh's and @benilovj's bits.

benilovj:
:+1:

PR #75, opened 2016-01-25
Migrate Statistics Announcements

dsingleton:
> Migrating the **Statistics Announcement** format to use Content Store, and moving it from whitehall to government-frontend.

- https://trello.com/c/9pMsgEJK/179-statistics-announcement-front-end-work-medium

## What we did

Each commit maps to a feature of the format, more or less, so that'd a good way to read this PR.

| Before | After |
| --- | --- |
| before | after |

This included some intentional product/design changes (all agreed with Jen):
- Use "from" rather than "organisations" label for metadata, to be consistent with other Whitehall document pages
- Use "part of" rather than "topic" label for metadata, to be consistent with other Whitehall document pages
- Changed the order of metadata items, so "from" / "part of". This is due to the behaviour of the [metadata component](http://govuk-component-guide.herokuapp.com/components/metadata), which will always list "standard" (eg, "from" / "part of"), before non-standard "other" items - intentionally, to enforce some consistency of metadata across GOV.UK pages. The downside of the order change is that the dates are _potentially_ less visible, not being at the top of the metadata, however, thats a reckon, and if dates are important they should be somewhere more prominent than in the metadata anyway. Jen was happy with the order changing for now, and that we may look at this again when rationalising formats.
- Altered the announcement "summary" presentation to match the description presentation on other formats. This avoids the long line length issue on the original, but has some issues for very long summaries. Looking at the average lengths of published summaries in production shows that 99% are less than 400 characters, which looks reasonable, and outliers (like the example in the screenshot) will look a bit weird, but no worse than the original. This allows it to be consistent with other formats in the app, and avoid adding unneeded format specific design. 

## What we didn't do

There were a few things we intentionally skipped, some of them make sense to do later, when we have a better idea of duplication patterns and some are smaller/non-blocking issues we felt could be done in follow up PRs, to allow the bulk of the migration to be reviewable sooner. 

We didn't do much on the duplication between the new and Case Study presenter. They're pretty similar, as are their tests, mostly in common field accessors, generating link markup for metadata. We could have generalised this code, but with only two examples it didn't feel right yet and we know we'll be revisiting it for the next few formats.

From the image comparison there is some significant whitespace difference, that matches the existing case studies layout. If we want to alter the whitespace that we should do that independently from migrating a format.

We had to add a new translation key, but were unable to run `rake translation:regenerate` to generate the empty keys for other locales. The same task fails on master, so not tackling that in this PR.

## What we think would help next time

General reckons we noted as we went a long.

- Feels like we're missing a layer of abstraction in the Presenters, things like generating the metadata markup for `links` applies to 90% of formats we'll migrate and should "free" to new formats. Either adding to the base `ContentItemPresenter` or creating a `DocumentContentItemPresenter` that sits above individual WH formats.
- We found writing the presenter unit tests a bit of a pain, quite a lot of duplication in set, and we were unsure how much of the example content items to rely on. In the end we stubbed quite a few field, rather than relying on schema examples that could change independently.


fofr:
Could we fix the padding/indentation on the notice styles?

fofr:
> "if dates are important"

They are possibly the most important reason for viewing this page.

dsingleton:
> Could we fix the padding/indentation on the notice styles? I think it's reasonable to do as part of this PR as it's a regression for this format type.

I was planning to do it in a separate PR, once `wraith` is usable to test it, along with a couple of other FE tweaks that would benefit from a proper visual diff.

test/presenters/statistics_announcement_presenter_test.rb
```diff
@@ -0,0 +1,128 @@
...
+    links = [
+      link_to('Policy Area', '/orgs/policy-area')
+    ]
+    assert_equal links, item.part_of
+  end
+
+  test 'presents release_date' do
+    item = presented_statistics_announcement({
+      "details" => {
+        "display_date" => "About Midday on Tuesday"
```
fofr: As discussed on Slack, it would be good to see this be a more realistic example, like “April to May 2016”

config/locales/en.yml
```diff
@@ -172,6 +172,9 @@ en:
       statistical_data_set:
         one: Statistical data set
         other: Statistical data sets
+      statistics_announcement:
+        one: Statistics release announcement
+        other: Statistics release announcement
```
fofr: I think this should be `Statistics release announcements`

dsingleton:
> They are possibly the most important reason for viewing this page.

That was rhetorical, and so the difference of being at the top or the bottom of the metadata isn't in making a major change. Given, the aim of this story/PR is to migrate the format without straying into re-inventing the format, or adding too much customisation.

config/locales/en.yml
dsingleton: Thanks, I forgot to push the fix.

app/views/content_items/statistics_announcement.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, "#{@content_item.title}" %>
```
fofr: Should the format be in the title?

Current titles look like:
`<title>Provisional Accident and Emergency Quality Indicators for England - October 2015, by provider   - Statistics release announcement - GOV.UK</title>`

fofr:
>  In the end we stubbed quite a few field, rather than relying on schema examples that could change independently.

When working on Take Part I thought that was an intentional coupling. @jamiecobbett do you have thoughts on this, it looks like you worked on a lot of the original tests.

A few comments, but generally I'm happy with approach and details in this PR about decisions made :+1: :100:. Good work @dsingleton and @binaryberry

dsingleton:
> I think this should be Statistics release announcements
>  Add a line note



> When working on Take Part I thought that was an intentional coupling. @jamiecobbett do you have thoughts on this, it looks like you worked on a lot of the original tests.

I don't have a super strong opinion on it, but it does feel a bit brittle and reminds me of copy changes breaking integration tests in the past.

jamiecobbett:
@dsingleton I agree does feel a bit brittle, but it feels like a reasonable compromise. You should find out that you've broken these tests as soon as you open a Pull Request to change the schema examples.

dsingleton:
> @dsingleton I agree does feel a bit brittle, but it feels like a reasonable compromise. You should find out that you've broken these tests as soon as you open a Pull Request to change the schema examples.

I don't think being made aware of tests failing is the issue, more the faff of having to update another repo (or more) when making an example change. But thats not a problem we've had yet - so should I be taking that approach, using examples as fixtures with no modifications, for this PR?

Gemfile
```diff
@@ -27,4 +27,5 @@ group :test do
   gem 'capybara'
   gem 'webmock', '~> 1.18.0', :require => false
   gem 'govuk-content-schema-test-helpers', '1.1.0'
+  gem 'pry'
```
benlovell: Add `pry-byebug` instead for even more wealth, greatness and the ability to advance up and down stack frames.

app/presenters/statistics_announcement_presenter.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def release_date_and_status
+    return "#{release_date} (#{state})" unless cancelled?
+    release_date
+  end
+
+  def other_metadata
+    if cancelled?
+    {
```
benlovell: Can we indent this hash one level, I thought I was looking at JavaScript for a second :eyeglasses:

app/presenters/statistics_announcement_presenter.rb
```diff
@@ -0,0 +1,58 @@
...
+      "Proposed release" => release_date,
+      "Cancellation date" => cancellation_date
```
benlovell: Can we add a trailing comma here, it helps with diffs when other folks add further entries plus it's recommended in the style guide.

app/presenters/statistics_announcement_presenter.rb
```diff
@@ -0,0 +1,58 @@
...
+    }
+    else
+      {"Release date" => release_date_and_status}
```
benlovell: Can you add whitespace around the hash braces here?

test/presenters/statistics_announcement_presenter_test.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    item = presented_statistics_announcement({
+      "details" => {
+        "format_sub_type" => "national"
+      }
+    })
+    assert item.national_statistics?
+  end
+
+private
```
benlovell: It's not generally worth using access modifiers in the tests.

Gemfile
binaryberry: Thanks, updated now!

app/presenters/statistics_announcement_presenter.rb
binaryberry: Oh dear! Amended now :-)

app/presenters/statistics_announcement_presenter.rb
binaryberry: True, amended now!

app/presenters/statistics_announcement_presenter.rb
binaryberry: Done, thanks!

test/presenters/statistics_announcement_presenter_test.rb
binaryberry: Thanks, addressed now. Opening a separate PR to remove them in other format's tests.

dsingleton:
@jamiecobbett Added a commit (https://github.com/alphagov/government-frontend/commit/35679644b3621e035a06230d5312095d2b02aaa6) to real content schema examples, rather than stubbing out values.

boffbowsh:
Can we get a better logo while we're at it? It offends my retina [screens]

boffbowsh:
1s and 0s look good to me :+1:

benlovell:
:sparkles: :sparkling_heart: :blowfish:

jamiecobbett:
:100:

PR #76, opened 2016-01-26
DO NOT MERGE Html sanitizer CVE fix

thehenster:
DO NOT MERGE - The tests pass but I've been unable to poke this in development mode as I've only been here a couple of days and my VM doesn't run properly. If you could poke this for me that would be great.

Upgrade Rails and Slimmer gem related to the below security releases.

http://weblog.rubyonrails.org/2016/1/25/Rails-5-0-0-beta1-1-4-2-5-1-4-1-14-1-3-2-22-1-and-rails-html-sanitizer-1-0-3-have-been-released/


boffbowsh:
I've merged this as testing on integration is easiest in that case, @thehenster

PR #77, opened 2016-01-26
Use GOV.UK Govspeak component for rendering SM guide body

tadast:
Use the component because it has support for automatic youtube video link
embedding. It modifies the styling of text slightly, which is not ideal, but we
decided it's better than overwriting the components styles.

The long term plan is to list the discrepancies and update the component with
improvements that have been discovered through Service Manual user research.

[Compare. Left: Before. Right: After.](https://cloud.githubusercontent.com/assets/218239/12582503/7b16c5d6-c435-11e5-99d0-fd774309a372.png)

### DIFF
--- a/app/assets/stylesheets/views/_service-manual-guide.scss
+++ b/app/assets/stylesheets/views/_service-manual-guide.scss
@@ -2,49 +2,9 @@
   margin-top: 1.875em; // 30/16
 }
 
-.prose {
-
-  margin-top: 1em;
-  margin-bottom: 2em;
+.govspeak-wrapper {
+  padding-top: 0.9375em;
   @include media(tablet) {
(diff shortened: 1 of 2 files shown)


tadast:
@gemmaleigh if you look at the image on the right [here](https://cloud.githubusercontent.com/assets/218239/12582503/7b16c5d6-c435-11e5-99d0-fd774309a372.png) it looks like the body is no longer aligned with the sidebar. Could you fix it please?

gemmaleigh:
@tadast that's been fixed - there's now a parent wrapper `.govspeak-wrapper` aligning the main content area.

tadast:
Cool, let's :shipit: if you're happy

fofr:
:+1:

PR #78, opened 2016-01-26
Fetch content owner info from the links hash

tadast:
As a part of an effort to reference everything using content_ids allow
Service manual guides to use information about the service manual guide content
owner from the links hash (see [1]) but fall back to the old way when the links
hash is not present.

This will allow us to migrate to the new format without breaking anything.

The fallback code should be removed once the migration process is complete.

[1] - https://github.com/alphagov/govuk-content-schemas/pull/219

### DIFF
--- a/app/presenters/service_manual_guide_presenter.rb
+++ b/app/presenters/service_manual_guide_presenter.rb
@@ -13,8 +13,12 @@ class ServiceManualGuidePresenter < ContentItemPresenter
   end
 
   def content_owner
-    content_owner_data = content_item["details"]["content_owner"]
-    ContentOwner.new(content_owner_data["title"], content_owner_data["href"])
+    content_owner_data = Array(content_item["links"] && content_item["links"]["content_owners"]).first
+    if content_owner_data.present?
+      ContentOwner.new(content_owner_data["title"], content_owner_data["base_path"])
+    else
+      content_owner_fallback
+    end
   end
 
   def related_discussion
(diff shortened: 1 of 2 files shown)


app/presenters/service_manual_guide_presenter.rb
```diff
@@ -13,8 +13,12 @@ def initialize(content_item)
   end
 
   def content_owner
-    content_owner_data = content_item["details"]["content_owner"]
-    ContentOwner.new(content_owner_data["title"], content_owner_data["href"])
+    content_owner_data = Array(content_item["links"].try(:[], "content_owners")).first
```
boffbowsh: I'm not sure `.try(:[], key)` is a good pattern here. I think this might read better:

```ruby
if content_item["links"] && content_item["links"]["content_owners"].present?
  content_owner_data = content_item["links"]["content_owners"]
  ContentOwner.new(content_owner_data["title"], content_owner_data["base_path"])
else
  content_owner_data = content_item["details"]["content_owner"]
  ContentOwner.new(content_owner_data["title"], content_owner_data["href"])
end
```

boffbowsh:
Style nitpick, otherwise :+1:

app/presenters/service_manual_guide_presenter.rb
tadast: `Array(...).first` here is also important :) I could rewrite it as 
```
Array(content_item["links"] && content_item["links"]["content_owners"]).first
``` 
but I don't find it any more readable :confused: 

http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig would be ideal here :)

tadast:
@boffbowsh I've added a fixup commit with an alternative syntax. Should I keep and squash it or revert back?

thehenster:
It looked better to me. Squashed.

PR #79, opened 2016-01-26
Remove access modifier in tests

binaryberry:
As per @benlovell's [comment](https://github.com/alphagov/government-frontend/pull/75#discussion-diff-50845288R109) on PR #75, they're not worth using in tests.


benlovell:
:dancers: :dancers: :dancers: :sparkler:

PR #80, opened 2016-01-26
Log to STDOUT and STDERR in production mode.

benilovj:
To bring this more in line with 12-facgtor principals, this changes the
logging to go to STDOUT(json), and STDERR(everything else).

Precedent: https://github.com/alphagov/publishing-api/pull/109
Needed Puppet change: https://github.com/alphagov/govuk-puppet/pull/3998

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -53,6 +53,7 @@ Rails.application.configure do
 
   # Use a different logger for distributed setups.
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr))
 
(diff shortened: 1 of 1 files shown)


boffbowsh:
Nice :+1:  How do we coordinate the deployment of this repo with the puppet change? Which needs to happen first? Does it matter if there's a large window between them?

benilovj:
If I'm reading the [app class in puppet](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/app.pp#L340-L352) correctly, STDOUT and STDERR output gets piped into upstart logs. The puppet change would then make those those logs be slurped by the regular logging mechanism.

The app should be deployed before puppet, because the puppet change cleans up the old logfile.

PR #83, opened 2016-01-29
Add per example integration testing

dsingleton:
Government Frontend already has _some_ [integration tests](https://github.com/alphagov/government-frontend/blob/master/test/contracts/govuk_content_schemas_test.rb), that ensure none of the content examples for supported formats return a non-200. This gives some coverage, but when working on Statistics Announcement @binaryberry and I felt that we were missing proper integration coverage, and that it was hard to approach adding a format by TDD.

This PR adds integration tests that be run per format/example, specifically asserting the behaviour of the different examples. The commits show a slight evolution of how I approached this, so read them if you want to know the decisions leading to the overall diff.

I've added some quite basic tests for Take Part and Case Study to illustrate usage, and make sure the code wasn't tightly coupled with Statistics Announcements. They're not exhaustive, but at least new formats can follow this pattern.

The overall pattern is an integration test per format, with a test per example. Each example should exercise some combination of features of that format and should be tested. Simpler formats, like Take Part, have few variations and only have one test.

I've also tried to reduce the boilerplate in the actual tests to make them easier to read, hopefully.

**Thoughts:**

- There's a growing collection of component based asserts in `test_helper`, which have some duplicated logic, but not _quite_ enough to generalise it _yet_. Also, as that list grows it'll make less sense to move that out in to a `component_asserts` type module, but again, not there yet.
- For formats with many examples, there is a lot of duplication in "basic" behaviours, like titles, descriptions, etc. I wonder if this might be better done in test per-format that runs across all examples for that format, testing the common behaviours, leaving the per-example tests to deal with example specific logic. Or we could only include the common behaviours tests in the first/happy-case example and omit them in other examples.
- Some of the component test helpers would be useful in other apps, and longer term I think we'll want to extract them out into `slimmer`, but in a way thats aware of different test frameworks.

### DIFF
--- a/test/integration/case_study_test.rb
+++ b/test/integration/case_study_test.rb
@@ -18,9 +18,7 @@ class CaseStudyTest < ActionDispatch::IntegrationTest
 
     within ".withdrawal-notice" do
       assert page.has_text?('This case study was withdrawn'), "is withdrawn"
-      within shared_component_selector("govspeak") do
-        assert_equal @content_item["details"]["withdrawn_notice"]["explanation"], JSON.parse(page.text).fetch("content")
-      end
+      assert_has_component_govspeak(@content_item["details"]["withdrawn_notice"]["explanation"])
     end
   end
 end
--- a/test/integration/case_study_test.rb
+++ b/test/integration/case_study_test.rb
@@ -1,0 +1,26 @@
+require 'test_helper'
+
+class CaseStudyTest < ActionDispatch::IntegrationTest
+  test "translated case study" do
+    setup_and_visit_content_item('translated')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+
+    assert page.has_css?('.available-languages')
+  end
+
+  test "withdrawn case study" do
+    setup_and_visit_content_item('archived')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+
+    within ".withdrawal-notice" do
+      assert page.has_text?('This case study was withdrawn'), "is withdrawn"
+      within shared_component_selector("govspeak") do
+        assert_equal @content_item["details"]["withdrawn_notice"]["explanation"], JSON.parse(page.text).fetch("content")
+      end
+    end
+  end
+end
--- a/test/integration/statistics_announcement_test.rb
+++ b/test/integration/statistics_announcement_test.rb
@@ -4,20 +4,20 @@ class StatisticsAnnouncementTest < ActionDispatch::IntegrationTest
   test "official statistics" do
     setup_and_visit_content_item('official_statistics')
 
-    assert_has_component_title('Diagnostic imaging dataset for September 2015')
+    assert_has_component_title(@content_item["title"])
   end
 
   test "national statistics" do
     setup_and_visit_content_item('national_statistics')
 
-    assert_has_component_title('UK armed forces quarterly personnel report: 1 October 2015')
+    assert_has_component_title(@content_item["title"])
     assert page.has_css?('.national-statistics-logo img')
   end
 
   test "cancelled statistics" do
     setup_and_visit_content_item('cancelled_official_statistics')
 
-    assert_has_component_title('Diagnostic imaging dataset for September 2015')
+    assert_has_component_title(@content_item["title"])
     assert page.has_text?('Statistics release cancelled'), "is cancelled"
   end
 
@@ -28,7 +28,7 @@ class StatisticsAnnouncementTest < ActionDispatch::IntegrationTest
   end
 
   def setup_and_visit_content_item(name)
-    JSON.parse(get_content_example(name)).tap do |item|
+    @content_item = JSON.parse(get_content_example(name)).tap do |item|
       content_store_has_item(item["base_path"], item.to_json)
       visit item["base_path"]
     end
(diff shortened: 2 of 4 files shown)


test/integration/case_study_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    assert page.has_css?('.available-languages')
+  end
+
+  test "withdrawn case study" do
+    setup_and_visit_content_item('archived')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+
+    within ('.withdrawal-notice') do
```
benlovell: Can we remove this space between the method name and parentheses?

test/integration/statistics_announcement_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_helper'
+
+class StatisticsAnnouncementTest < ActionDispatch::IntegrationTest
+
```
benlovell: Can we remove this whitespace line for consistency?

test/integration/statistics_announcement_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    assert page.has_css?('.national-statistics-logo img')
+    assert_has_component_metadata_pair("Release date", "January 2016 (provisional)")
+  end
+
+  test "cancelled statistics are rendered correctly" do
+    setup_and_visit_content_item('cancelled_official_statistics')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+    within ('.cancellation-notice') do
```
benlovell: Same here with the space.

test/test_helper.rb
```diff
@@ -20,4 +21,37 @@ class ActionController::Base
...
+      visit item["base_path"]
+    end
+  end
+
+  def get_content_example(name)
+    GovukContentSchemaTestHelpers::Examples.new.get(schema_format, name)
+  end
+
+  def schema_format
+    # Override this method if your test file doesn't match the convention
```
benlovell: This may be super-niggly but perhaps this comment should go above the method signature since that is what it applies to?

benlovell:
I just did my best impression of rubocop, but that aside, this looks pretty awesome. :+1:

test/integration/case_study_test.rb
```diff
@@ -0,0 +1,26 @@
...
+      assert page.has_text?('This case study was withdrawn'), "is withdrawn"
+      within shared_component_selector("govspeak") do
+        assert_equal @content_item["details"]["withdrawn_notice"]["explanation"], JSON.parse(page.text).fetch("content")
```
jamiecobbett: > JSON.parse(page.text).fetch("content")

How is that working?

test/integration/statistics_announcement_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    assert_has_component_metadata_pair("Release date", "20 January 2016 9:30am (confirmed)")
+  end
+
+  test "national statistics are rendered correctly" do
+    setup_and_visit_content_item('national_statistics')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+    assert page.has_css?('.national-statistics-logo img')
+    assert_has_component_metadata_pair("Release date", "January 2016 (provisional)")
```
jamiecobbett: It feels like this test and the one above have a slightly odd relationship between their names and the last assertion. Is it genuinely a difference between the formats, or is it because each example happens to demonstrate one of provisional vs confirmed?

test/integration/statistics_announcement_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  test "cancelled statistics are rendered correctly" do
+    setup_and_visit_content_item('cancelled_official_statistics')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+    within ('.cancellation-notice') do
+      assert page.has_text?('Statistics release cancelled'), "is cancelled"
+      assert page.has_text?(@content_item["details"]["cancellation_reason"]), "displays cancelleation reason"
```
jamiecobbett: It might be safer to use `@content_item["details"].fetch("cancellation_reason")` so that we at least know if the `cancellation_reason` key is removed.

jamiecobbett:
A couple of comments, otherwise :+1:

test/integration/case_study_test.rb
dsingleton: In test mode `slimmer` doesn't fetch components from `static`. Instead it inserts a placeholder element that contains a JSON representation of the arguments passed to the component partial. eg:


```
<%= render 'govuk_component/title', title: "hello world" %>
```

Is rendered as 
```
<test-govuk-component data-template="govuk_component-title">{ "title": "Hello World" }</test-govuk-component>
```

test/integration/statistics_announcement_test.rb
dsingleton: The later. The examples happen to be a mix, presumably intentionally so we have both cases covered without creating a combinatorial set of examples.

I think it might be clearer if the test name was just the example name, to indicate it's exercising that _example_.

dsingleton: I've changed the test names, which I think is enough for now. I think we can improve the readability of these kind of tests as we add more.

dsingleton:
Addressed linting concerns with some rebase fixing. 

@jamiecobbett are you happy with the answers/changes? It'd be good to get this merged. I've got a branch to add a missing statistics feature that's based on the work here.

boffbowsh:
LGTM :+1:

PR #84, opened 2016-01-29
Add Ruby linting

benlovell:
Lint the Ruby in the `app`, `config`, `test` directories and the `Gemfile`.


benbarnett:
:+1:

dsingleton:
:+1:  to having linting.

When we've added linting to [other apps](https://github.com/alphagov/whitehall/blob/master/jenkins.sh#L55) we've tended to use the setting that only lints _changes_, rather than the whole app. This avoids "noise" being added to git history when any existing violations are fixed.

There are a smaller number of changes needed on this app than some of the others, so it _might_ be worth doing. But IMO the main benefit we want is future violations to be enforces by a computer, rather than a well meaning human, which we don't need to fix existing violations to get.

benlovell:
@dsingleton I'm not sure I see the benefit overall. If the whole project is passing (as it does now) and CI fails the build unless any further additions also pass, there are simply no means to merge code which would normally introduce violations. Does that make sense, or am I missing something?

dsingleton:
> @dsingleton I'm not sure I see the benefit overall. If the whole project is passing (as it does now) and CI fails the build unless any further additions also pass, there are simply no means to merge code which would normally introduce violations. Does that make sense, or am I missing something?

No, that makes sense. The downside is muddying git history with the autocorrect commits, vs not adding to git history and only linting the _diff_ of a change.

I'm not entirely against adding the autocorrect commits now, but it's worth thinking about the trade off.

PR #85, opened 2016-01-29
Support Statistics announcements with changed date

dsingleton:
This applies when a 'confirmed' announcement has it's release date
edited, which should not normally happen. When it does it must be
disclosed, and the reason explained.

diagnostic imaging dataset for september 2015 statistics release announcement gov uk

In WH this disclosure is done as a separate headed section, with
a second 'metadata'-style block to format the date and reason.

Long term this may not be the right pattern for the page, and may
be consolidated into the main metadata block, or otherwise be
made more consistent with other document formats.

To get the feature migrated in the short term i've included a second
call to the metadata component, which causes some complexity in the
integration test, where there is now two metadata components when
asserting. In order to avoid the conflict, i've scoped the metadata
asserts for the `release_date_changed` example only, as the others
don't have multiple metadata blocks.

This does add a slightly spurious class to distinguish the first/
main metadata block, but it feels better than adding a lot of
complexity to the tests or test helpers. If this is a pattern we
encounter in more formats in the future, then we should look at
adding support to the test helpers for cases like this.

### DIFF
--- a/app/assets/stylesheets/views/_statistics_announcement.scss
+++ b/app/assets/stylesheets/views/_statistics_announcement.scss
@@ -9,4 +9,11 @@
   .cancellation-notice {
     @include notice;
   }
+
+  .release-date-change-notice {
+    h2 {
+      font-weight: bold;
+      margin-bottom: $gutter-half;
+    }
+  }
 }
--- a/test/presenters/statistics_announcement_presenter_test.rb
+++ b/test/presenters/statistics_announcement_presenter_test.rb
@@ -21,6 +21,10 @@ class StatisticsAnnouncementPresenterTest < ActiveSupport::TestCase
     assert_equal '20 January 2016 9:30am', statistics_announcement.release_date
   end
 
+  test 'presents previous_release_date' do
+    assert_equal '19 January 2016 9:30am', statistics_announcement_date_changed.previous_release_date
+  end
+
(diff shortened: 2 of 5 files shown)


test/presenters/statistics_announcement_presenter_test.rb
```diff
@@ -58,6 +62,11 @@ class StatisticsAnnouncementPresenterTest < ActiveSupport::TestCase
     assert statistics_announcement_national.national_statistics?
   end
 
+  test 'knows if the release date has changed' do
+    assert statistics_announcement_date_changed.release_date_changed?
+    assert_not statistics_announcement_national.release_date_changed?
```
benlovell: `refute` reads a little nicer, I think.

benlovell:
The Ruby bits look good to me. My comment shouldn't stop you from merging if you don't agree with it.

app/assets/stylesheets/views/_statistics_announcement.scss
```diff
@@ -9,4 +9,11 @@
   .cancellation-notice {
     @include notice;
   }
+
+  .release-date-change-notice {
+    h2 {
```
fofr: It's frustrating we need to do this.

gpeng:
Looks good... I've no preference on the refute/assert_not comment and am happy to merge if you want to respond.

PR #87, opened 2016-02-01
Add wraith for visual regression testing

dsingleton:
[Example output](https://dl.dropboxusercontent.com/u/262911/wraith/shots/gallery.html)

Add Wraith so that regressions can be easily spotted. Runs against a
set of supported examples. Run while pointing government-frontend at
the dummy content store, and an instance of static.

Usage:
On master run: wraith history test/wraith/config.yaml
On local branch run: wraith latest test/wraith/config.yaml

Ran `bundle update` to get a version of nokogiri that would work with
`wraith`.

Configuration generated with `wraith setup` and combined with our
existing wraith usage, [1]. With modern versions of wraith you don't
need a `snap.js` file, as wraith will take care of the `phantomjs`
intergration, however we still need to set some cookies to prevent
intermitent UI (like the survey, or cookie bar) from interfering
with the diff results.

[1] https://github.com/alphagov/design-principles/pull/246


fofr:
:+1: Looks good to me.

benlovell:
:ship:

dsingleton:
Summary of a brief offline discussion. This still needs some work to be robust, and there are some underlying issues in wraith that concern me, _but_, this is useful as it is, and we have something we can improve on when we need

PR #88, opened 2016-02-01
Prevent widows in descriptions

fofr:
Stop widows by putting a `&nbsp;` between the last two words in a description

## Before
screen shot 2016-02-01 at 16 21 34

## After
screen shot 2016-02-01 at 16 21 43

### DIFF
--- a/app/helpers/typography_helper.rb
+++ b/app/helpers/typography_helper.rb
@@ -1,0 +1,5 @@
+module TypographyHelper
+  def nbsp_between_last_two_words(text)
+    text.sub(/\s([\w\.\?\!]+)$/, '&nbsp;\1').html_safe
+  end
+end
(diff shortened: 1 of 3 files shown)


app/helpers/typography_helper.rb
```diff
@@ -0,0 +1,5 @@
+module TypographyHelper
+  def nbsp_between_last_two_words(text)
+    text.sub(/\s([\w\.\?\!]+)$/, '&nbsp;\1').html_safe
```
dsingleton: It would be good to have a test that would assert what happens if the `text` input included markup, eg:
```
assert_equal "this &lt;b&gt;&nbsp;&lt;/b&gt; that", nbsp_between_last_two_words("this <b>and</b> that")
```

I think there's a bug here, even if not, it would be good for the test to cover that case if the implementation changed, and to self-document.

PR #89, opened 2016-02-01
Improve whitespace on withdrawn/cancelled notices

dsingleton:
👀   &nbsp;   **[Before, after and diff of affected pages](https://dl.dropboxusercontent.com/u/262911/government-frontend-notice-spacing/index.html)** &nbsp;  👀

The mixin that powers both of these had poor padding of elements
inside the blue box, and between the title and p/govspeak.

Use standard gutters, tweaked for smaller screens to match spacing
on smaller devices. Use the responsive margin mixin to ensure the
space following the notice looks even compared to other parts of
the page.

Tweak the HTML for case studies, as the notice appeared between
grid columns, but not marked up as a column, so in single-column
(mobile) mode it clung to the edge of the page. Instead, break it
out in to it's own grid row, as it is for statistics announcement
and it will flow to the edge of the grid normally.

### DIFF
--- a/app/assets/stylesheets/helpers/_notice.scss
+++ b/app/assets/stylesheets/helpers/_notice.scss
@@ -1,17 +1,18 @@
 @mixin notice {
   clear: both;
-  margin: 15px 0;
-  padding: 15px 10px;
+  @include responsive-bottom-margin;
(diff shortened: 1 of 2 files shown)


fofr:
:+1:

PR #90, opened 2016-02-02
Service Manual - Topic pages

gemmaleigh:
This PR adds a subsection module for subtopics on a topic page.  The title, description and links underneath for each of the topic pages will later be replaced with an accordion, to hide the list of links.

Also add styles for the related content area to link to the community pages.


gemmaleigh:
cc. @dsingleton, this PR restructures the Sass files as we discussed.

app/views/content_items/topic.html.erb
```diff
@@ -0,0 +1,58 @@
+<%= content_for :page_class, "service-manual" %>
```
thehenster: Maybe this should be "topic" now the this isn't service-manual specific.

AndrewVos:
lgtm

fofr:
Some of the sass in this will not pass the linting checks, which were only just added. https://github.com/alphagov/government-frontend/pull/101

app/assets/stylesheets/application.scss
```diff
@@ -3,22 +3,23 @@
 // govuk_frontend_toolkit imports
 @import 'colours';
 @import 'measurements';
-@import 'typography';
 @import 'shims';
+@import 'typography';
 
 // helpers for common page elements
-@import "helpers/responsive-bottom-margin";
```
fofr: Order was important here as `responsive-bottom-margin` is referenced in later helpers, eg https://github.com/alphagov/government-frontend/blob/master/app/assets/stylesheets/helpers/_sidebar-with-body.scss#L7

and
https://github.com/alphagov/government-frontend/blob/master/app/assets/stylesheets/helpers/_notice.scss#L3

I'm surprised the `notice` styles aren't erroring in tests.

AndrewVos:
@fofr Not seeing any lint output now with:

```
bundle exec govuk-lint-sass app
```

app/assets/stylesheets/application.scss
thehenster: Hi,

I think we're getting away with it because notice (and description should suffer too) are `@mixin`s and aren't getting evaluated until they're used. The first time it's used is the first view import `@import "views/case-studies";`.

The tests pass and it works in the flesh but I think it would read better as it was ordered originally. I'll add a PR to reorder them again.

Thanks

PR #92, opened 2016-02-05
Render "Topical event about page" format

fofr:
* Purposefully alter the topical event about page to look more like other formats. Move the description up beneath the title, introduce a breadcrumb, add spacing beneath body, remove underlined heading styles in main content
* Port `dash-list` styles from Whitehall for contents rendering
* Add helper for generating a list of contents based on `h2` headings within format's body
* To keep the design consistent with what’s currently live, introduce an offset class which is used only when there is no sidebar.
* Indicate when a topical event is archived in the breadcrumb

https://trello.com/c/P4KCL1UJ/267-3-topical-event-about-page-front-end-work-medium

# Before
ebola-old

# After
ebola

---

# Before
somme-old

# After
somme


app/controllers/content_items_controller.rb
```diff
@@ -23,8 +23,9 @@ def load_content_item
   def present(content_item)
     case content_item['format']
     when 'case_study' then CaseStudyPresenter.new(content_item)
-    when 'take_part' then TakePartPresenter.new(content_item)
     when 'statistics_announcement' then StatisticsAnnouncementPresenter.new(content_item)
+    when 'take_part' then TakePartPresenter.new(content_item)
+    when 'topical_event_about_page' then TopicalEventAboutPagePresenter.new(content_item)
```
benilovj: It feels to me like there's enough duplication here to use convention over configuration (ie to derive the presenter class name from the format name):

```ruby
def present(content_item)
  presenter_name = content_item['format'].classify + 'Presenter'
  presenter_class = Object.const_get(presenter_name)
  presenter_class.new(content_item)
rescue NameError
  raise "No support for format \"#{content_item['format']}\""
end
```

benlovell: Much nicer :sparkles: 

`String#constantize` wraps the `Module.const_get` dance (but doesn't respect lexical context) if you go with that approach.

app/presenters/extracts_headings.rb
```diff
@@ -0,0 +1,10 @@
+module ExtractsHeadings
+  def extract_headings_with_ids(html)
+    [].tap do |headings|
```
benilovj: I think you can get rid of one level of nesting with the following:

```ruby
module ExtractsHeadings
  def extract_headings_with_ids(html)
    Nokogiri::HTML(html).css('h2').map do |heading|
      id = heading.attribute('id')
      { text: heading.text, id: id.value } if id
    end.compact
  end
end
```

app/presenters/topical_event_about_page_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+  include ExtractsHeadings
+  include ActionView::Helpers::UrlHelper
+
+  def body
+    content_item["details"]["body"]
+  end
+
+  def contents
+    extract_headings_with_ids(body).map do |heading|
+      link_to(heading[:text], "##{heading[:id]}")
```
benilovj: Maybe `"#" + heading[:id]` will be a bit easier to read...

app/views/shared/_sidebar_contents.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if contents.any? %>
+<nav role="navigation">
```
benilovj: should this be indented (due to the `if` statement above)?

test/test_helper.rb
```diff
@@ -45,6 +45,20 @@ def assert_has_component_govspeak(content)
     end
   end
 
+  def assert_has_component_breadcrumbs(breadcrumbs)
+    within shared_component_selector("breadcrumbs") do
+      assert_equal breadcrumbs, JSON.parse(page.text).deep_symbolize_keys.fetch(:breadcrumbs)
+    end
+  end
+
+  def assert_has_contents(contents)
```
benilovj: this method name sounds very generic but does something very specific...

test/presenters/extracts_headings_test.rb
```diff
@@ -0,0 +1,30 @@
...
+  include ExtractsHeadings
+
+  test "extracts a single h2" do
+    html = '<h2 id="custom">A heading</h2>'
+    assert_equal [{text: "A heading", id: 'custom'}], extract_headings_with_ids(html)
+  end
+
+  test "ignores headings without an id" do
+    html = '<h2>John Doe</h2>'
+    assert_equal [], extract_headings_with_ids(html)
```
benilovj: Personally, I think `assert_empty extract_headings_with_ids(html)` would read better...

test/integration/topical_event_about_page_test.rb
```diff
@@ -0,0 +1,54 @@
...
+    assert_has_contents([
+      {text: "Response in the UK", id: "response-in-the-uk"},
+      {text: "Response in Africa", id: "response-in-africa"},
+      {text: "Advice for travellers", id: "advice-for-travellers"}
+    ])
+  end
+
+  test "breadcrumbs show whether a topical event is archived" do
+    @content_item = JSON.parse(get_content_example("topical_event_about_page")).tap do |item|
+      content_store_has_item(item["base_path"], item.to_json)
```
benilovj: Is there much point doing the `tap` call, as opposed to just a sequential statement?

```ruby
@content_item = JSON.parse(get_content_example("topical_event_about_page"))
content_store_has_item(@content_item["base_path"], @content_item.to_json)
```

Gemfile
```diff
@@ -23,6 +23,7 @@ end
 group :development, :test do
   gem 'govuk-lint'
   gem 'wraith', '~> 3.1'
+  gem 'timecop', '0.8.0'
```
benlovell: Rails has `travel_to` built in since `4.1` which has feature-parity with `timecop` and doesn't require yet another dependency.

app/presenters/extracts_headings.rb
benlovell: :heart: aside: why is there no `Enumerable#compact_map`?

app/presenters/topical_event_about_page_presenter.rb
benlovell: Hmm, I'm not sure.

app/presenters/topical_event_about_page_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def breadcrumbs
+    parent = topical_event
+    title = archived_topical_event? ? "#{parent['title']} (Archived)" : parent["title"]
+
+    [
+      {title: "Home", url: "/"},
```
benlovell: A general style point: can we add spaces inside of the hash braces throughout?

benilovj: I was wondering – how come linting doesn't pick that up?

test/test_helper.rb
fofr: Good point, I'll update to `contents-list`

app/presenters/topical_event_about_page_presenter.rb
benlovell: Oh: https://github.com/alphagov/govuk-lint/blob/05da1e32636f7fd695a2488d486ad78b8b7ad113/configs/rubocop/other-style.yml#L63

Maybe I'll change it and see what people think...

benlovell: alphagov/govuk-lint#26 :boom:

app/presenters/extracts_headings.rb
fofr: @benilovj This doesn't meet our style criteria:
```app/presenters/extracts_headings.rb:6:5: C: Style/MethodCalledOnDoEndBlock: Avoid chaining a method call on a do...end block.  (https://github.com/bbatsov/ruby-style-guide#single-line-blocks)
    end.compact```

Comment (benlovell) [inline app/presenters/extracts_headings.rb]: Using braces, since this block returns a value, should be OK.

File: app/presenters/extracts_headings.rb [hunk 10]
```diff
@@ -0,0 +1,8 @@
+module ExtractsHeadings
+  def extract_headings_with_ids(html)
+    Nokogiri::HTML(html).css('h2').map { |heading|
+      id = heading.attribute('id')
+      { text: heading.text, id: id.value } if id
+    }.compact
```
benilovj: To bring it into the house style, you can change this to:

```ruby
headings = ...
headings.compact
```

app/presenters/topical_event_about_page_presenter.rb
fofr: Fixed in f56cc2e

Gemfile
fofr: Now removed

app/controllers/content_items_controller.rb
fofr: Nice, I've updated.

app/presenters/extracts_headings.rb
fofr: Updated to use braces.

app/views/shared/_sidebar_contents.html.erb
fofr: Added indentation

test/test_helper.rb
fofr: Done in 38ef266

test/presenters/extracts_headings_test.rb
fofr: I agree, updated

test/integration/topical_event_about_page_test.rb
fofr: Good point, removed.

fofr:
Updated and rebased. I've taken on most of Jake's suggestions, removed Timecop, upgraded gouvk-lint and fixed the style offences.

benlovell:
:sparkling_heart: :rocket:

PR #94, opened 2016-02-09
Customise the phase banner message for service manual guides

gemmaleigh:
Using the phase label helper, set the message for service manual guide pages. 

http://govuk-component-guide.herokuapp.com/components/beta_label

cc. @dsingleton.

### DIFF
--- a/app/helpers/phase_label_helper.rb
+++ b/app/helpers/phase_label_helper.rb
@@ -1,7 +1,9 @@
 module PhaseLabelHelper
-  def render_phase_label(presented_object)
+  def render_phase_label(presented_object, message)
     if presented_object.respond_to?(:phase) && %w(alpha beta).include?(presented_object.phase)
-      render partial: "govuk_component/#{presented_object.phase}_label"
+      render partial: "govuk_component/#{presented_object.phase}_label", locals: {
+        message: message
+      }
     end
   end
 end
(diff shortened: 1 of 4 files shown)


dsingleton:
Broadly :+1: to the idea.

Implementation feedback:
- Message should probably be optional in the `render_phase_label` definition
- This change should update the [accompanying tests](https://github.com/alphagov/government-frontend/blob/use-beta-banner-for-guides/test/integration/phase_label_test.rb), and make use of the [more precise component asserts](https://github.com/alphagov/government-frontend/blob/use-beta-banner-for-guides/test/test_helper.rb#L36-L40) to ensure the right args are being passed, as passing an empty `message` argument [may not work as expected](https://github.com/alphagov/static/blob/master/app/views/govuk_component/alpha_label.raw.html.erb#L5).

app/helpers/phase_label_helper.rb
```diff
@@ -1,7 +1,9 @@
 module PhaseLabelHelper
-  def render_phase_label(presented_object)
+  def render_phase_label(presented_object, message)
```
boffbowsh: For backwards compatibility, I think this should be defaulted to `nil` or `""`.

AndrewVos:
Thanks!

PR #100, opened 2016-02-15
Configure asset host in development mode

fofr:
When run under `govuk_setenv`, this configures the asset host to point at the assets-origin on the dev VM.  This will give us better dev-prod parity, and enable running this behind the router in dev.

See also: https://github.com/alphagov/finder-frontend/commit/d8f25e9afc1fefbd28adb30ccc358622f8bf5f23

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -31,4 +31,8 @@ Rails.application.configure do
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  if ENV['GOVUK_ASSET_ROOT'].present?
+    config.asset_host = ENV['GOVUK_ASSET_ROOT']
+  end
 end


config/environments/development.rb
```diff
@@ -31,4 +31,8 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  if ENV['GOVUK_ASSET_ROOT'].present?
```
benlovell: The `present?` isn't strictly necessary here.

boffbowsh:
It'd be better to use `Plek.new.asset_root` here

fofr:
Any updates here should probably also be reflected in https://github.com/alphagov/finder-frontend/commit/d8f25e9afc1fefbd28adb30ccc358622f8bf5f23 and https://github.com/alphagov/frontend/commit/77bb2cb3a05a8a87b23008f6fcc8b7f5ff9b88c5

dsingleton:
> It'd be better to use Plek.new.asset_root here

That doesn't have quite the right behaviour in all dev situations, unfortunately.

If `GOVUK_ASSET_ROOT` is not set then it defaults to `static` host as the `asset_host` to serve a FE app's assets.

On the VM within `bowl` this will work, as `govuk_setenv` will set `GOVUK_ASSET_ROOT`. Outside of `bowl` it will try and load the FE app's assets from the `static domain`, and 404s. Running a FE app outside the VM would have the same broken behaviour.

boffbowsh:
Fair enough :+1:

PR #107, opened 2016-02-22
Run the ruby and sass linters when the default rake task is run

AndrewVos:
This should help us stop committing code that will fail the linter under CI.

### DIFF
--- a/lib/tasks/lint.rake
+++ b/lib/tasks/lint.rake
@@ -1,0 +1,11 @@
+desc "run ruby and sass linters"
+task :lint do
+  if ENV["JENKINS"]
+    sh "govuk-lint-ruby --format html --out rubocop-${GIT_COMMIT}.html --format clang Gemfile app test config"
+  else
+    sh "govuk-lint-ruby --format clang Gemfile app test config"
+  end
+  sh "govuk-lint-sass app"
+end
+
+Rake::Task[:default].enhance [:lint]
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -42,16 +42,6 @@ git clone git@github.com:alphagov/govuk-content-schemas.git tmp/govuk-content-sc
 
 bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment --without development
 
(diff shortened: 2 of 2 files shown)


AndrewVos:
It just occurred to me that this will run the linters twice, because in jenkins.sh `rake` gets executed.

Is there a reason why linting does not happen on master?

jamiecobbett:
I think your commit should say "run ruby and sass linters as part of the default rake task". Additionally, does this overwrite the default rake task that runs all tests?

AndrewVos:
@jamiecobbett It doesn't appear so, but I'll switch to another way so it's more explicit.

```
[~/.../government-frontend] run-linter-by-default ∞ rake
govuk-lint-ruby --diff --format clang Gemfile app test config
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.8-compliant syntax, but you are running 2.1.4.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.

0 files inspected, no offenses detected
govuk-lint-sass app
Run options: --seed 8519

# Running:

....................................................................................................

Finished in 0.890031s, 112.3557 runs/s, 212.3523 assertions/s.

100 runs, 189 assertions, 0 failures, 0 errors, 0 skips

```

lib/tasks/lint.rake
```diff
@@ -0,0 +1,12 @@
...
+task :lint do
+  if ENV["JENKINS"]
+    sh "govuk-lint-ruby --format html --out rubocop-${GIT_COMMIT}.html --format clang Gemfile app test config"
+  else
+    sh "govuk-lint-ruby --format clang Gemfile app test config"
+  end
+  sh "govuk-lint-sass app"
+end
+
+Rake::Task[:default].enhance [:lint]
```
AndrewVos: @jamiecobbett more explicit form ^

boffbowsh: Are both lines 2 and 12 needed?

AndrewVos: @boffbowsh Nope! Can't believe I missed that! Thanks.

PR #111, opened 2016-02-24
Render HTML publication format

fofr:
This is an _all the basics_ PR, it doesn't include the following features:
* a contents link that moves as you scroll
* organisation logos
* <del>working fractions (if changes are necessary)</del>

Does include:
* Rendering of HTML publication content using the new govspeak-html-publication component. Depends on https://github.com/alphagov/static/pull/744
* Rendering of the blue HTML publication header, with a contents list that is provided by the content item
* Add files for styles, template, presenter and tests
* Use `HtmlPublication` rather than `HTMLPublication` as that’s what Rails expects.

Tests also depend on https://github.com/alphagov/govuk-content-schemas/pull/250 (hence the test fails)

https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large

html-pub-format


app/presenters/html_publication_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+  def body
+    content_item["details"]["body"]
+  end
+
+  def contents
+    content_item["details"]["headings"].html_safe
+  end
+
+  def format_sub_type
+    content_item["links"]["parent"][0]["format_sub_type"]
```
benlovell: Maybe a private helper method for `parent` that does the reaching in could simplify this?

app/presenters/html_publication_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+    # This assumes that a translation doesn't need the date to come beforehand.
+    if content_item["details"]["first_published_version"]
+      "#{I18n.t('content_item.metadata.published')} #{timestamp}"
+    else
+      "#{I18n.t('content_item.metadata.updated')} #{timestamp}"
+    end
+  end
+
+  def parent_base_path
+    content_item["links"]["parent"][0]["base_path"]
```
benlovell: Said helper method could be used here also.

dsingleton:
I'm picking up this work from @fofr while he's away, it's already 95% there.

After the changes in the schema/examples fractions now Just Work™, as the SCSS already existed in the [Govspeak Component](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/govuk-component/_govspeak.scss#L323-L334). Though there isn't a component fixture for fractions, so i'll add one and double check print behaves as we expect.

I've added the helper @benlovell suggested, which reduces some duplication :+1: and updated the schema to avoid some weird HTML whitespace issues, which @benlovell is looking at more deeply

I'm happy with leaving org-logos and the fixed contents link to follow up PRs.

I want to have a look at the integration testing, as it's only covering one of the schema examples and I want to make sure we're not missing any cases. After that I think this is :ship:-able.

app/assets/stylesheets/mixins/_white-links.scss
```diff
@@ -0,0 +1,30 @@
+// TODO (@bfirsh): refactor frontend_toolkit external link mixin to work with
```
dsingleton: This is in a few places now, is it viable to move into FET?

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -0,0 +1,84 @@
+.html-publication {
+
+  .govuk-title .context {
```
dsingleton: I don't think we should be relying on the internal markup of a component, that can change independently of this application.

The risk of this approach is that changes to the component, in markup/selectors, or explicitly setting colours on elements that don't have them could break this page without static being aware of the dependency. And it's generally a pattern we try and avoid. 

This feels more like a mode of the component and the responsibility for a white links/different margin version should live within the component, not the app. 

Alternatively, we treat it as a non standard page title, and don't use the component at all, setting _all_ the styles locally.

app/assets/stylesheets/mixins/_white-links.scss
fofr: Follow up PR needed I think.

fofr:
@dsingleton Title component changes made in bd99c4f

app/assets/stylesheets/mixins/_white-links.scss
dsingleton: :+1:

fofr: White external links aren't needed for publications, they were added for consultations. Removing this and the images from this PR.

dsingleton:
Decided against adding the additional integration tests for other content examples - they vary that much and where they do tends to be inside markup blocks that are rendered as is, which we're less concerned about testing in this app.

When we add departmental logo support that would be probably deserve an interrogation test for that example.

PR #117, opened 2016-03-02
Colourise minitest output

AndrewVos:
image


boffbowsh:
I don't think this is really necessary.

PR #118, opened 2016-03-02
Working groups

jamiecobbett:
## Here is the "short" example:
short-before
short-after

<hr>

## "with_policies" example
with_policies-before
with_policies-after

<hr>

## "long" example
long-before
long-after

### DIFF
--- a/app/views/content_items/working_group.html.erb
+++ b/app/views/content_items/working_group.html.erb
@@ -1,0 +1,45 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, @content_item.title %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        title: @content_item.title %>
+  </div>
+</div>
+
+<% unless @content_item.description.blank? %>
+  <%= render 'shared/description', description: @content_item.description %>
+<% end %>
+
+<div class="grid-row sidebar-with-body">
+  <% if @content_item.contents.any? %>
+    <div class="column-third">
+      <%= render 'shared/sidebar_contents', contents: @content_item.contents %>
+    </div>
+  <% end %>
+  <div class="column-two-thirds <% unless @content_item.contents.any? %>offset-one-third<% end %>">
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction %>
+
+    <% @additional_body = capture do %>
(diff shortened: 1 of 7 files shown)


fofr:
Can you add some screenshots of before and after?

test/test_helper.rb
```diff
@@ -52,9 +59,13 @@ def assert_has_component_breadcrumbs(breadcrumbs)
...
 
   def assert_has_contents_list(contents)
+    assert page.has_css?(".dash-list"), "Failed to find an element with a class of dash-list"
     within ".dash-list" do
       contents.each do |heading|
-        assert page.has_css?("a[href=\"##{heading[:id]}\"]", text: heading[:text])
+        selector = "a[href=\"##{heading[:id]}\"]"
+        text = heading.fetch(:text)
+        assert page.has_css?(selector), "Failed to find an element matching: #{selector}"
+        assert page.has_css?(selector, text: text), "Failed to find an element matching #{selector} with text: #{text}"
```
fofr: :+1: :100:

test/test_helper.rb
```diff
@@ -39,12 +39,19 @@ def assert_has_component_title(title)
     end
   end
 
-  def assert_has_component_govspeak(content)
-    within shared_component_selector("govspeak") do
+  def assert_has_component_govspeak(content, index: 1)
```
fofr: Nice pattern. I think we could re-use this pattern on one of the stats announcements which has multiple components of the same type, metadata I think. cc @dsingleton

fofr:
Looks good to me – pending a visual diff with screenshots, I like that you didn't need to write any new styles :+1:

jamiecobbett:
@fofr screenshots added

test/test_helper.rb
dsingleton: That _is_ nice.

app/views/content_items/working_group.html.erb
```diff
@@ -0,0 +1,47 @@
...
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction %>
+
+    <% @additional_body = capture do %>
+      <% if @content_item.policies.any? %>
+        <h2 id="policies">Policies</h2>
+        <ul>
+          <% @content_item.policies.each do |policy| %>
+            <%= content_tag(:li) do %>
```
fofr: This could be `<li>`

PR #122, opened 2016-03-15
Use sticky_footer_html in html publications.

tvararu:
Uses the new module developed in https://github.com/alphagov/static/pull/757. Do not merge until that PR is ready.

Relevant Trello ticket: https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large.

Previous implementation to play around with, on live Whitehall: http://www.gov.uk/government/publications/dwp-single-departmental-plan-2015-to-2020/dwp-single-departmental-plan-2015-to-2020.

Looks just as before (red added for emphasis, first screenshot is "hiding" it as per the desired behaviour):

screen shot 2016-03-15 at 17 12 07
screen shot 2016-03-15 at 17 12 09
screen shot 2016-03-15 at 17 12 21

### DIFF
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -91,4 +91,17 @@
     }
   }
 
+  .back-to-content {
+    @include core-19;
+    display: block;
+    padding-bottom: $gutter-half;
+
+    &:before {
+      content: "\2191";
+    }
+
+    @media print {
+      display: none;
+    }
+  }
 }
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -28,4 +28,5 @@
     locals: {
       content: @content_item.body,
       direction: page_text_direction,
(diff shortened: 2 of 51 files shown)


app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -91,4 +91,17 @@
...
+  .back-to-content {
+    @include core-19;
+    display: block;
+    padding-bottom: $gutter-half;
+
+    &:before {
+      content: "\2191";
+    }
+
+    @media print {
```
fofr: Static also has a `dont-print` class we could use for this:
https://github.com/alphagov/static/blob/31f79a61ca5535ec805c02ff690616823cc32ad9/app/assets/stylesheets/helpers/_print-base.scss#L85

tvararu: I'd like `@include dont-print;` mixin, and also have it exposed through `.dont-print`. Best of both worlds. :+1:

fofr: Sounds good to me.

PR #123, opened 2016-03-16
Display the organisation logos in HTML publications.

tvararu:
This is part of migrating `html_publications` from `whitehall` to `government-frontend`.

Worth mentioning: in the presenter, organisations are sorted because the `content_item["links"]` are not guaranteed to come in a meaningful order.

Relevant Trello ticket: https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large

Looks like this:

screen shot 2016-03-16 at 16 40 48

### DIFF
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -4,9 +4,28 @@
   content_for :simple_header, true
 %>
 
-<p class="see-more-about">
-  <%= link_to t('html_publication.see_more', document_type: t("content_item.format.#{@content_item.format_sub_type}", count: 1)), @content_item.parent_base_path %>
-</p>
+<div class="publication-external">
+  <ol class="organisation-logos">
+    <% @content_item.organisations.each do |organisation| %>
+      <li class="organisation-logo">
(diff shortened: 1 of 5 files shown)


app/views/content_items/html_publication.html.erb
```diff
@@ -4,9 +4,28 @@
...
-</p>
+<div class="publication-external">
+  <ol class="organisation-logos">
+    <% @content_item.organisations.each do |organisation| %>
+      <li class="organisation-logo">
+        <%= render partial: 'govuk_component/organisation_logo', locals: {
+          organisation: {
+            name: organisation["logo"]["formatted_title"],
+            url: organisation["base_path"],
+            brand: organisation["brand"],
```
fofr: What brand gets passed for the Prime Minister's Office? Whitehall currently uses "cabinet-office", which would give a blue border. However the component needs to be passed "executive-office".

I mentioned this to @boffbowsh before going away, just checking it's not been missed. See https://github.com/alphagov/static/pull/754/commits/f9e9e426dc94c3368b7dafedfc0d419bf09292f3 for reference.

PR #127, opened 2016-03-23
Improve HTML publications RTL rendering.

tvararu:
Part of the addressing the final unintended changes introduced with migrating HTML publications in https://trello.com/c/cJGrtnc1/286-5-html-publications-migration-front-end-work-large.

This fixes the rendering for the header and the see more about link. Changes are grouped in this PR because they are very similar.

### Before

screen shot 2016-03-23 at 19 14 13

### After

screen shot 2016-03-23 at 19 14 32

### DIFF
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -31,6 +31,21 @@
           right: $gutter-half;
         }
       }
+
+      &.direction-rtl {
+        direction: rtl;
+        text-align: start;
+
+        // Chrome (and probably other browsers) does a weird thing where it changes
+        // the font kerning ever so slightly when RTL rendering latin characters,
+        // which will cause the glyphs to become slightly wider, which will cause
+        // them to collapse over multiple rows even though they do have enough space to fit.
+        // It's reproducible only for specific character strings, such as "See More".
+        // This is a sheepish fix, by giving the element much more space than it needs.
+        .see-more-about {
+          width: 100%;
+        }
+      }
     }
   }
 
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -56,6 +56,11 @@
 
     background: $light-blue;
     padding: $gutter $gutter $gutter / 3;
+
+    &.direction-rtl {
+      direction: rtl;
+      text-align: start;
+    }
   }
 
   .in-page-navigation {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -24,6 +24,7 @@
         @include core-19;
         padding-top: $gutter-one-third;
         position: static;
+        width: 100%;
```
tvararu: Right, this one is weird. We shouldn't need it, but some very weird Chrome-specific rendering bug appears when you don't have it, for ONLY SOME strings of characters, when `direction: rtl` is flipped on. May as well post pictures:

### Before

screen shot 2016-03-23 at 19 17 46

### After

screen shot 2016-03-23 at 19 17 54

app/views/content_items/html_publication.html.erb
```diff
@@ -3,6 +3,14 @@
   content_for :title, @content_item.title
   content_for :simple_header, true
 %>
+<%
+  publication_header_direction_class = ""
+  publication_header_direction_class = " direction-#{page_text_direction}" if page_text_direction
+%>
+<%
```
fofr: The separate `%> <%` blocks seem unnecessary here.

app/views/content_items/html_publication.html.erb
```diff
@@ -3,6 +3,14 @@
...
+  see_more_about_direction_class = ""
+  see_more_about_direction_class = " direction-#{page_text_direction}" if page_text_direction
```
fofr: Would a single `direction_class` used for both circumstances be better?

tvararu: Can do, changing to that.

tvararu:
Updated. :white_check_mark:

fofr:
Code LGTM.

Like another PR I reviewed I think the commits could give a bit more context around what and why (ie why it wasn't doing that already, why see more needs some extra styling). Some guidance on how we structure commit messages here: https://github.com/alphagov/styleguides/blob/master/git.md

> Makes HTML publications more compliant with right to left languages.

"Compliant" isn't plain english and feels a bit like one of our words to avoid:
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#words-to-avoid

tvararu:
:+1: 

I decided that the extra styling hack could do with a proper comment, as that's a more likely place for it to be noticed should somebody ask in the future "why is this necessary?"

PR #130, opened 2016-04-01
Communities in service manual topic sidebar

thehenster:
List the communities related to the guides included in a topic in the sidebar and invite visitors to join the community.

https://trello.com/c/rZlsL4Am/152-add-communities-sidebar-to-the-topic-pages

Related content schemas PR: https://github.com/alphagov/govuk-content-schemas/pull/277

PR #131, opened 2016-04-01
Fix display of contact details on policy groups

fofr:
### Don't show contact in contents when no email

When an email is excluded it’s sent as an empty string in the content item, rather than being omitted. The template checks for `email.present?` but the contents logic doesn’t, which creates a discrepancy.

screen shot 2016-04-01 at 14 27 25

--

### Use single govspeak component

When using separate components the vertical rhythm of content could break. We apply special rules to the first element in govspeak, eg removing headings from the first H2.

The second govspeak component would remove the margin from the “Contact details” heading, so its content would appear closer to the body than it should.

Rather than wrap that content and give it some margin, concatenate the two bodies and pass to a single component.

#### Before
screen shot 2016-04-01 at 14 27 11

#### After
screen shot 2016-04-01 at 14 26 55

### DIFF
--- a/app/assets/stylesheets/views/_topical-event-about-page.scss
+++ b/app/assets/stylesheets/views/_topical-event-about-page.scss
@@ -3,7 +3,6 @@
   @include sidebar-with-body;
 
   .offset-one-third {
-    // scss-lint:disable SpaceAroundOperator
-    margin-left: percentage(1/3);
+    margin-left: percentage(1 / 3);
   }
 }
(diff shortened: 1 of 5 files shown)


fofr:
Test failures appear unrelated: `ServiceManualTopicTest` cc @AndrewVos

AndrewVos:
@fofr Fixed in https://github.com/alphagov/government-frontend/pull/130, apologies, turns out a govuk-content-schema PR shouldn't have been merged without that one.

AndrewVos:
@fofr Fixed in https://github.com/alphagov/government-frontend/pull/130, apologies, turns out a govuk-content-schema PR shouldn't have been merged without that one.

fofr:
Tests are passing now.

tvararu:
:+1:

PR #132, opened 2016-04-05
Generator for adding formats

fofr:
Based on @dsingleton's work for components: https://github.com/alphagov/static/pull/444

Usage: `rails generate format [format_name]`

Adds the following files:
* View specific stylesheet (and appends include to application.scss)
* Format erb template with title, page class, title component and
description
* Format presenter
* Format presenter test
* Format integration test

Based around these early commits when adding formats:
* https://github.com/alphagov/government-frontend/pull/69/commits/8be8c0e52628ebc8b3fa4810401c7a850d999ee4 Take part
* https://github.com/alphagov/government-frontend/pull/111/commits/b33a8009c10a295af4e1e117676fd7d06a09ca9b HTML publications
* https://github.com/alphagov/government-frontend/pull/92/commits/5f504db8f1400cf3d0843470fc2e85fec881731f Topical event about page


lib/generators/format/format_generator.rb
```diff
@@ -0,0 +1,20 @@
+require 'json'
+
+class FormatGenerator < Rails::Generators::NamedBase
+  source_root File.expand_path('../templates', __FILE__)
+
+  def copy_format_files
+    format_name = file_name
```
boffbowsh: It'd be good to call `underscore` on this. Rails generators for e.g. model accept both forms and turn them into the correct thing, e.g. `rails g model FooBar` is equivalent to `rails g model foo_bar`.

fofr: Good point, added in 5792915

boffbowsh:
:+1: but will let @dsingleton look first and merge

boffbowsh:
Nah, let's merge it

PR #133, opened 2016-04-07
Show service manual guide change history at the bottom of the page

AndrewVos:
https://trello.com/c/qzHbv9S9/181-show-change-notes-and-page-history-at-the-bottom-of-guidance-page

## On initial page load

image

## After clicking that link

image

### DIFF
--- a/app/views/content_items/service_manual_guide.html.erb
+++ b/app/views/content_items/service_manual_guide.html.erb
@@ -76,3 +76,43 @@
 
   </div>
 </div>
+
+<% if @content_item.change_history.any? %>
+  <div class="grid-row change-history" data-module="toggle">
+    <div class="column-one-third">
+      &nbsp;
+    </div>
+    <div class="column-two-thirds">
+      <div>
+        Published by:
+      </div>
+
+      <% @content_item.content_owners.each do |content_owner| %>
+        <div class="change-history-published-by">
+          <%= link_to content_owner.title, content_owner.href %>
+        </div>
+      <% end %>
+
+      <div>
+        Last update:
+      </div>
+
+      <% first_change, *rest_of_changes = @content_item.change_history %>
+
+      <%= render "shared/change_history_item", change_history_item: first_change %>
+
+      <% if rest_of_changes.any? %>
+        <a href="#change-history-hidden" data-expanded="false" data-controls="change-history-hidden">
+          + full page history
+        </a>
+
+        <div id="change-history-hidden" class="js-hidden">
+          <% rest_of_changes.each do |change_history| %>
+            <%= render "shared/change_history_item", change_history_item: change_history %>
+          <% end %>
+        </div>
+      <% end %>
+    </div>
+
+  </div>
+<% end %>
--- a/test/integration/service_manual_guide_test.rb
+++ b/test/integration/service_manual_guide_test.rb
@@ -16,4 +16,48 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
       refute page.has_content?('Published by')
     end
   end
+
+  test "service manual guide shows latest change history" do
+    setup_and_visit_content_item('with_change_history')
+    within ".change-history" do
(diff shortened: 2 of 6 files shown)


app/views/content_items/service_manual_guide.html.erb
```diff
@@ -76,3 +76,45 @@
...
+          <%= link_to content_owner.title, content_owner.href %>
+        </div>
+      <% end %>
+
+      <div>
+        Last update:
+      </div>
+
+      <% first_change = @content_item.change_history.first %>
+      <% rest_of_changes = @content_item.change_history[1..-1] %>
```
dougdroper: first_change, *rest_of_changes = @content_item.change_history

test/integration/service_manual_guide_test.rb
```diff
@@ -16,4 +16,48 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
+      assert_includes timestamps, "09 October 2017"
+    end
+  end
+
+  test "service manual guide hides older change history" do
+    setup_and_visit_content_item('with_change_history')
+
+    module_container = find(".change-history")
+    assert_equal "toggle", module_container["data-module"]
+    assert_equal "toggle", module_container["data-module"]
```
thehenster: The same assertion twice here.

test/integration/service_manual_guide_test.rb
```diff
@@ -16,4 +16,48 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
+
+    hidden_container = find(".change-history #change-history-hidden")
```
thehenster: IDs should be unique so you shouldn't need the `.change-history`.

test/integration/service_manual_guide_test.rb
```diff
@@ -16,4 +16,48 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
+    assert_equal "js-hidden", hidden_container["class"]
+
+    expand_link = find(".change-history a", text: "+ full page history")
+    assert_equal "false", expand_link["data-expanded"]
+    assert_equal "change-history-hidden", expand_link["data-controls"]
+
+    within ".change-history #change-history-hidden" do
```
thehenster: Same here with the ID being all you need.

test/integration/service_manual_guide_test.rb
```diff
@@ -16,4 +16,48 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
+      notes = all(".change-history-note").map(&:text)
+      assert_includes notes, "This is another change"
+
+      reasons = all(".change-history-reason").map(&:text)
+      assert_includes reasons, "This is why we made this change and it has a second line of text"
+
+      timestamps = all(".change-history-public-timestamp").map(&:text)
+      assert_includes timestamps, "09 January 2016"
+    end
+  end
```
thehenster: Personally I'm unsure of this business of checking that some elements have certain classes rather than checking the classes are doing what we want them to do. eg. use a real browser driver such an phantomjs or selenium, rather than rack-test, and check something isn't visible rather than check `js-hidden` is present.

Maybe these things are tested elsewhere, we can trust it, and testing like this is sensible or perhaps we just don't have to bother testing it here?

Maybe @dsingleton or @fofr could guide us here?

thehenster:
I think, but let's check, that we are to add this to service-manual-frontend rather than here. Probably my fault. I should have been clearer that we're a) not extending the component and b) adding our custom version to our own frontend.

This will effectively block us on this feature until the new frontend is deployed but this is a lesser evil than adding this to government-frontend.

Right @dsingleton? (I've also pinged you on a question I have in the code for this PR too)

AndrewVos:
Why would we do this in our frontend, it's far from being deployed?

boffbowsh:
We (Core Formats) are going to help you get it deployed soon.

PR #134, opened 2016-04-18
Detailed guides migration

mgrassotti:
Detailed guides are now rendered from Government Frontend.
Trello: https://trello.com/c/k6H9gNKF/316-6-detailed-guides-migration-front-end-work-small

Please note that some differences are still in place, comparing the 2 rendered versions. 
I.e., using this example:
https://www.gov.uk/guidance/salary-sacrifice-and-the-effects-on-paye

**Part of** - footer
There are some differences about the positioning of the link "Full page history" and some character style differences. 
This is related to some changes of [document_footer](https://github.com/alphagov/static/blob/master/app/views/govuk_component/document_footer.raw.html.erb) in [static](https://github.com/alphagov/static). The PR will follow.
image

### DIFF
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -28,7 +28,7 @@ class DetailedGuidePresenter < ContentItemPresenter
   end
 
   def part_of
-    links("document_collections") + links("related_policies") + links("worldwide_priorities") + links("world_locations")
+    links("document_collections") + links("related_policies") + links("worldwide_priorities") + links("world_locations") + links("topics")
   end
 
   def history
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -1,2 +1,14 @@
 class DetailedGuidePresenter < ContentItemPresenter
+  include ExtractsHeadings
+  include ActionView::Helpers::UrlHelper
+
+  def body
+    content_item["details"]["body"]
+  end
+
+  def contents
+    extract_headings_with_ids(body).map do |heading|
+      link_to(heading[:text], "##{heading[:id]}")
+    end
+  end
 end
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -11,4 +11,71 @@ class DetailedGuidePresenter < ContentItemPresenter
       link_to(heading[:text], "##{heading[:id]}")
     end
   end
+
+  def context
+    parent["title"]
+  end
+
+  def breadcrumbs
+    [
+      { title: "Home", url: "/" },
+      { title: context, url: parent["base_path"] }
+    ]
+  end
+
+  def from
+    links("lead_organisations") + links("supporting_organisations") + links("worldwide_organisations")
+  end
+
+  def part_of
+    links("document_collections") + links("related_policies") + links("worldwide_priorities") + links("world_locations")
+  end
+
+  def history
+    return [] unless any_updates?
+    content_item["details"]["change_history"].map do |item|
+      {
+        display_time: display_time(item["public_timestamp"]),
+        note: item["note"],
+        timestamp: item["public_timestamp"]
+      }
+    end
+  end
+
+  def published
+    display_time(content_item["details"]["first_public_at"])
+  end
+
+  def updated
+    display_time(content_item["public_updated_at"]) if any_updates?
+  end
+
+  def short_history
+    if any_updates?
+      "#{I18n.t('content_item.metadata.updated')} #{updated}"
+    else
+      "#{I18n.t('content_item.metadata.published')} #{published}"
+    end
+  end
+
+private
+
+  def parent
+    content_item["links"]["parent"][0]
+  end
+
+  def any_updates?
+    if (first_public_at = content_item["details"]["first_public_at"]).present?
+      DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
+    else
+      false
+    end
+  end
+
+  def links(type)
+    return [] unless content_item["links"][type]
+    content_item["links"][type].map do |link|
+      link_to(link["title"], link["base_path"])
+    end
+  end
 end
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -1,9 +1,17 @@
 class DetailedGuidePresenter < ContentItemPresenter
   include ExtractsHeadings
   include ActionView::Helpers::UrlHelper
+  include CommonSections
 
   def body
     content_item["details"]["body"]
+  end
+
+  def breadcrumbs
+    [
+      { title: "Home", url: "/" },
+      { title: context, url: parent["base_path"] }
+    ]
   end
 
   def contents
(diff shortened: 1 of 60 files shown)


mgrassotti:
This is related to https://github.com/alphagov/govuk-content-schemas/pull/279

app/views/content_items/detailed_guide.html.erb
```diff
@@ -15,11 +15,11 @@
...
         from: @content_item.from,
-        updated: @content_item.updated,
-        history: @content_item.short_history,
-        published: @content_item.published,
-        direction: page_text_direction,
-        part_of: @content_item.part_of
+        other: {
+          "First published" => @content_item.published,
+          "Last updated" => @content_item.updated,
+          "Part of" => @content_item.part_of #doing this to preserve the ordering but if part_of, at the same level as from and other
```
fofr: I wonder if the ordering needs to be preserved.

app/presenters/detailed_guide_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+      DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
+    else
+      false
+    end
+  end
+
+  def links(type)
+    return [] unless content_item["links"][type]
+    content_item["links"][type].map do |link|
+      link_to(link["title"], link["base_path"], class: "#{type}-link")
```
fofr: Why do we need the dynamic `class: "#{type}-link"` classes?

app/presenters/detailed_guide_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def breadcrumbs
+    [
+      { title: "Home", url: "/" },
+      { title: context, url: parent["base_path"] }
+    ]
+  end
+
+  def from
+    links("lead_organisations") + links("supporting_organisations") + links("worldwide_organisations")
```
fofr: Some of these bits were copied from case studies (published, from, part of, history), could the shared parts be refactored into helpers/modules, maybe like https://github.com/alphagov/government-frontend/blob/master/app/presenters/extracts_headings.rb

app/views/content_items/detailed_guide.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  </div>
+</div>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/metadata',
+        from: @content_item.from,
+        other: {
+          "First published" => @content_item.published,
+          "Last updated" => [@content_item.updated,"<a class=\"see-all-updates\" href=\"#history\">see all updates</a>"].join(", "),
```
fofr: The `see-all-updates` class isn't needed here

app/presenters/detailed_guide_presenter.rb
mgrassotti: @fofr 👍
One question: the `part_of` method includes `links("topics")` for detailed_guides but not for case_study. Should they be different?

boffbowsh:
The breadcrumbs at the top aren't quite right, this will take some changes in here, content store and content schemas:

The actual breadcrumbs at the top should be 
> Home > Business tax > PAYE

The Business tax pseudo-breadcrumb link next to "Guidance" should disappear as part of this, the only breadcrumbs we should have on this page are the ones provided by the breadcrumb component.

The `parent` of this page should be PAYE, which the content store will need to recursively resolve to also include Business tax. The "Part of" metadata should just render whatever's in the `topics` field. This will start out as just Business tax and PAYE, but Finding Things will be changing this later, we don't need to worry about it on Core.

This is all a bit complicated, so shout if anything is unclear.

app/presenters/detailed_guide_presenter.rb
fofr: Is Case study effectively the same? If they are then the content items should match. Thoughts @boffbowsh?

mgrassotti:
The top of the page looks like this now:

image

benilovj:
From the screenshots on the PR description, it looks like the "Is there anything wrong with this page?" widget alignment is off ([case studies](https://www.gov.uk/government/case-studies/doing-business-in-spain) are an example of how it should look like).

mgrassotti:
thanks @benilovj, this indeed appears only when the window size is not full. But reducing the size of the case studies example, it seems to be the same... isn't it?

image

benilovj:
> thanks @benilovj, this indeed appears only when the window size is not full. But reducing the size of the case studies example, it seems to be the same... isn't it?

Ah, ok, so it's an existing issue. We should still fix it at some point, but it shouldn't block this PR...

app/presenters/detailed_guide_presenter.rb
```diff
@@ -16,6 +16,7 @@ def breadcrumbs
       e = e["parent"] && e["parent"].first
     end
 
+    res << { title: "Home", url: "/" }
```
mgrassotti: Manually adding the "Home" link as this is not present in the parent hierarchy.

fofr: Good spot 👍

test/presenters/detailed_guide_presenter_test.rb
```diff
@@ -0,0 +1,40 @@
...
+    assert_equal "/topic/business-tax", presented_example.breadcrumbs[1][:url]
+    assert_equal "PAYE", presented_example.breadcrumbs[2][:title]
+    assert_equal "/topic/business-tax/paye", presented_example.breadcrumbs[2][:url]
+  end
+
+  def format_name
+    "detailed_guide"
+  end
+
+  def example
```
fofr: `example` and `presented_example` could be avoided by updated the `example_content_item` and `presented_example_content_item` calls in the test helper to default to `format_name` if none is provided

app/views/content_items/detailed_guide.html.erb
```diff
@@ -6,7 +6,7 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <%= render 'govuk_component/title',
-        context: "#{@content_item.context} — guidance",
+        context: "Guidance",
```
fofr: This should use the translated string.

app/views/content_items/detailed_guide.html.erb
```diff
@@ -0,0 +1,48 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, @content_item.title %>
+
+<%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        context: "Guidance",
```
fofr: The context should be translated. Other formats use:
```erb
context: t("content_item.format.#{@content_item.format_display_type}", count: 1)
```

fofr:
One last comment regarding translations, otherwise :+1:

fofr:
According to https://docs.google.com/spreadsheets/d/1AWsPFXUhJzqCBR0r0iag1UZ1bavneu8ZBABZzKFZ3XY/edit#gid=1910524167, detailed guides support history mode, withdrawal, political status and publishing government.

This PR doesn't include that feature – I think we should get this merged and work on the missing features on a new branch. This will also be a consideration for the content schema and publishing stories.

app/views/content_items/detailed_guide.html.erb
mgrassotti: It is now:

image

mgrassotti: Ah, that's why "Detailed Guidance" was renamed in "Detailed Guide" in the content schema example, probably to match with the Whitehall model name. I can amend the translation inside en.yml. Could it create problems somewhere else?

config/locales/en.yml
```diff
@@ -73,7 +73,7 @@ en:
       correspondence:
         one: Correspondence
         other: Correspondences
-      detailed_guidance:
```
fofr: Presumably all the other language keys also need updating?

mgrassotti: Yes, you're right. I quickly substituted them, we can go back eventually. In Whitehall I found both terms used, although "detailed_guide" seems to be used more frequently.

PR #135, opened 2016-04-22
Add withdrawn notice to detailed guides

fofr:
See also https://github.com/alphagov/govuk-content-schemas/pull/288

* Extract withdraw presenter logic into module, use in case studies and detailed guides
* Add withdrawn presenter and integration tests for both formats
* Extract withdrawn template into re-usable partial
* Extract withdrawn styles into re-usable helper mixin

Includes some minor bug fixes:
* Fix margins on withdrawal notices
* Avoid an empty "Part of" from showing in metadata component (example has no links)
* Avoid an empty breadcrumb from showing (example has no parent)
* Update titles to use average length, so they appear smaller and match what's currently live

Includes some refactoring:
* Updates presenter tests to all extend PresenterTest – DRY up loading the content item example and presenting the example item, and allows tests to share helper methods

https://trello.com/c/P2CLsRtp/383-6-detailed-guides-migration-front-end-work-withdrawing-documents-small

Compare screenshot below with live example: https://www.gov.uk/guidance/eu-rules-on-the-use-of-chemicals

screen shot 2016-04-22 at 11 38 53

### DIFF
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -1,7 +1,9 @@
 <%= content_for :page_class, @content_item.format.dasherize %>
 <%= content_for :title, @content_item.page_title %>
 
-<%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>
+<% if @content_item.breadcrumbs.any? %>
+  <%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>
+<% end %>
 
 <div class="grid-row">
   <div class="column-two-thirds">
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -7,7 +7,8 @@
   <div class="column-two-thirds">
     <%= render 'govuk_component/title',
         context: t("content_item.format.#{@content_item.format}", count: 1),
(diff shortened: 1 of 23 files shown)


fofr:
Tests passing now https://github.com/alphagov/govuk-content-schemas/pull/288 has been merged

app/views/content_items/detailed_guide.html.erb
```diff
@@ -20,10 +20,10 @@
   <div class="column-two-thirds">
     <%= render 'govuk_component/metadata',
         from: @content_item.from,
+        part_of: @content_item.part_of,
         other: {
```
mgrassotti: when https://github.com/alphagov/static/pull/774 will be merged then this could become:
```
<%= render 'govuk_component/metadata',
        from: @content_item.from,
        part_of: @content_item.part_of,
        first_published: @content_item.published,
        last_updated: @content_item.updated,
        see_updates_link: true
%>
```
right?

boffbowsh:
This needs rebasing

fofr:
@boffbowsh Rebased.

app/views/content_items/detailed_guide.html.erb
boffbowsh: I've merged that PR. Will this block be compatible?

fofr: Updated and published are in `other`, so they are compatible.

PR #139, opened 2016-05-04
Add minimum viable publication rendering

fofr:
Render the following example accurately:
https://www.gov.uk/government/publications/d17-9ly-veolia-es-uk-limited-environmental-permit-issued

Built in conjunction with:
https://github.com/alphagov/govuk-content-schemas/pull/298
(tests will fail until that is merged)

Intentionally, but temporarily, omits from this example:
* Format subtype
* Excluded nations

Further features and examples required for:
* History notice
* Withdrawn notice
* Ministers, topical events, related statistical data sets and other links

publication-format-preview

Part of:
https://trello.com/c/v85kPiKN/366-9-publications-migration-mvp-content-schema-examples-and-front-end-work-medium

### DIFF
--- a/app/presenters/linkable.rb
+++ b/app/presenters/linkable.rb
@@ -1,10 +1,17 @@
 module Linkable
   def from
-    links("lead_organisations") + links("supporting_organisations") + links("worldwide_organisations")
+    links("lead_organisations") +
+      links("organisations") +
+      links("supporting_organisations") +
+      links("worldwide_organisations")
   end
 
   def part_of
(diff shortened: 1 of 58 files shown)


app/presenters/linkable.rb
```diff
@@ -1,10 +1,17 @@
 module Linkable
   def from
-    links("lead_organisations") + links("supporting_organisations") + links("worldwide_organisations")
+    links("lead_organisations") +
```
boffbowsh: You'll need to run all this through `uniq`, as some orgs or policies may appear in more than one of these. Same for `part_of`

PR #140, opened 2016-05-05
Display applicable nations for detailed guides.

tvararu:
This implements a missing feature for the detailed guides migration.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/300 for the tests to pass, I think.

### After ("Applies to" displays)

screen shot 2016-05-05 at 16 46 25

cc @fofr

### DIFF
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -43,7 +43,34 @@ class DetailedGuidePresenter < ContentItemPresenter
     links("related_mainstream")
   end
 
+  def applies_to
+    return nil if !national_applicability
+    all_nations = national_applicability.values
+    applicable_nations = all_nations.select { |n| n["applicable"] }
+    inapplicable_nations = all_nations - applicable_nations
+
+    applies_to = applicable_nations.map { |n| n["label"] }.to_sentence
+
+    if inapplicable_nations.any?
+      nations_with_alt_urls = inapplicable_nations.select { |n| n["alternative_url"].present? }
+      if nations_with_alt_urls.any?
+        additional_guidance_links = nations_with_alt_urls
+          .map { |n| link_to(n['label'], n['alternative_url'], rel: :external) }
+          .to_sentence
+        # TODO: Translation for `see detailed guidance for` below?
+        additional_guidance = " (see detailed guidance for #{additional_guidance_links})"
+        applies_to += additional_guidance
+      end
+    end
+
+    applies_to
+  end
+
 private
+
+  def national_applicability
+    content_item["details"]["national_applicability"]
+  end
 
   def parent
     if content_item["links"].include?("parent")
--- a/test/integration/detailed_guide_test.rb
+++ b/test/integration/detailed_guide_test.rb
@@ -48,4 +48,16 @@ class DetailedGuideTest < ActionDispatch::IntegrationTest
       assert page.has_text?('This guidance was published under the 2010 to 2015 Conservative and Liberal Democrat coalition government')
     end
   end
+
+  test 'detailed guide that only applies to a set of nations' do
+    setup_and_visit_content_item('national_applicability_detailed_guide')
+
+    assert_has_component_metadata_pair('Applies to', 'England')
+  end
+
+  test 'detailed guide that only applies to a set of nations, with alternative urls' do
+    setup_and_visit_content_item('national_applicability_alternative_url_detailed_guide')
+
+    assert_has_component_metadata_pair('Applies to', 'England, Scotland, and Wales (see detailed guidance for <a href="http://www.dardni.gov.uk/news-dard-pa022-a-13-new-procedure-for" rel="external">Northern Ireland</a>)')
+  end
 end
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -23,7 +23,10 @@
         part_of: @content_item.part_of,
         first_published: @content_item.published,
         last_updated: @content_item.updated,
-        see_updates_link: true
+        see_updates_link: true,
+        other: {
+          'Applies to' => @content_item.applies_to
+        }
     %>
   </div>
 </div>
--- a/test/presenters/detailed_guide_presenter_test.rb
+++ b/test/presenters/detailed_guide_presenter_test.rb
@@ -66,4 +66,20 @@ class DetailedGuidePresenterTest < PresenterTest
 
     assert presented.historically_political?
   end
+
+  test 'content can apply only to a set of nations' do
+    example = schema_item('national_applicability_detailed_guide')
+    presented = presented_item('national_applicability_detailed_guide')
+
+    assert example['details'].include?('national_applicability')
+    assert_equal presented.applies_to, 'England'
+  end
+
+  test 'content can apply only to a set of nations with alternative urls' do
+    example = schema_item('national_applicability_alternative_url_detailed_guide')
+    presented = presented_item('national_applicability_alternative_url_detailed_guide')
+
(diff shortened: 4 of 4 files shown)


fofr:
Can you add screenshots?

tvararu:
I took one but I forgot to include it!

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,24 @@ def related_mainstream
     links("related_mainstream")
   end
 
+  def other
```
fofr: This should probably be `other_metadata`

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,24 @@ def related_mainstream
...
+    hash = {}
+    if !applies_to_all_nations
+      hash["Applies to"] = applies_to
```
fofr: `Applies to` _might_ need translating. eg into Welsh.

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,24 @@ def related_mainstream
...
+    end
+    hash
+  end
+
+  def applies_to_all_nations
```
fofr: I'd rename this to `applies_to_all_nations?`

app/presenters/detailed_guide_presenter.rb
tvararu: ❗ I knew I had to do this but completely forgot, good catch.

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,24 @@ def related_mainstream
     links("related_mainstream")
   end
 
+  def other
+    hash = {}
+    if !applies_to_all_nations
```
fofr: Would `unless` work better here?

tvararu: Yeah, *fine*. 😛 

(not a huge fan of `unless`)

app/presenters/detailed_guide_presenter.rb
tvararu: `government-frontend` has no translations. I think we should actually send in the translated keys in the schema, e.g.:

```diff
    "applicable_nations": {
      "all_nations": false,
-     "nations": ["england"]
+     "nations": ["England"]
    }
```

tvararu: Of which, by the way, we have none. 🎱

tvararu: I've coalesced my thoughts in this commit: https://github.com/alphagov/govuk-content-schemas/pull/300/commits/55c6fef7af0850a5e7bdd4e8b0adaefee4e6ee18

In short, they're going to be send down translated, so nothing to do on this end.

tvararu:
Updated to the latest schema changes, external link looks wonky:

screen shot 2016-05-09 at 15 13 01

tvararu:
screen shot 2016-05-09 at 15 25 55

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,33 @@ def related_mainstream
     links("related_mainstream")
   end
 
+  def national_applicability
```
fofr: Should this be private?

test/integration/detailed_guide_test.rb
```diff
@@ -48,4 +48,25 @@ class DetailedGuideTest < ActionDispatch::IntegrationTest
       assert page.has_text?('This guidance was published under the 2010 to 2015 Conservative and Liberal Democrat coalition government')
     end
   end
+
+  test 'detailed guide that only applies to a set of nations' do
+    setup_and_visit_content_item('national_applicability_detailed_guide')
+
+    within shared_component_selector('metadata') do
```
fofr: Does `assert_has_component_metadata_pair` not work here?

fofr:
Mentioned offline: The external link fix should live in the static component.

test/integration/detailed_guide_test.rb
tvararu: I had a comment about that and I removed it; but no.

That helper function purposefully removes `other`: `args.merge!(args.delete("other")) if args.key?("other")`.

I could create another helper function [but didn't seem worth it for 2 uses](http://c2.com/cgi/wiki?ThreeStrikesAndYouRefactor).

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,6 +43,33 @@ def related_mainstream
...
+    applies_to = applicable_nations.map { |n| n["label"] }.to_sentence
+
+    if inapplicable_nations.any?
+      nations_with_alt_urls = inapplicable_nations.select { |n| n["alternative_url"] && !n["alternative_url"].empty? }
+      if nations_with_alt_urls.any?
+        additional_guidance_links = nations_with_alt_urls
+          .map { |n| "<a class=\"country\" rel=\"external\" href=\"#{n['alternative_url']}\">#{n['label']}</a>" }
+          .to_sentence
+        # TODO: Translation for `see detailed guidance for` below?
+        additional_guidance = " (see detailed guidance for #{additional_guidance_links})"
```
fofr: This "detailed guidance" bit won't apply to all formats that need to use this pattern.

tvararu: Yeah, that's something to abstract once this becomes a helper, which can be done now or later for the publications frontend.

test/integration/detailed_guide_test.rb
tvararu: I am mistaken. 🎱

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,8 +43,35 @@ def related_mainstream
...
+    applicable_nations = all_nations.select { |n| n["applicable"] }
+    inapplicable_nations = all_nations - applicable_nations
+
+    applies_to = applicable_nations.map { |n| n["label"] }.to_sentence
+
+    if inapplicable_nations.any?
+      nations_with_alt_urls = inapplicable_nations.select { |n| n["alternative_url"] && !n["alternative_url"].empty? }
+      if nations_with_alt_urls.any?
+        additional_guidance_links = nations_with_alt_urls
+          .map { |n| "<a rel=\"external\" href=\"#{n['alternative_url']}\">#{n['label']}</a>" }
```
fofr: You could probably use the Rails link_to helper to build this link, that might be cleaner.
See: https://github.com/alphagov/government-frontend/blob/master/app/presenters/linkable.rb#L21-L23

tvararu:
External link icon fixed in https://github.com/alphagov/static/pull/785.

app/presenters/detailed_guide_presenter.rb
```diff
@@ -43,8 +43,35 @@ def related_mainstream
...
+  def applies_to
+    return nil if !national_applicability
+    all_nations = national_applicability.values
+    applicable_nations = all_nations.select { |n| n["applicable"] }
+    inapplicable_nations = all_nations - applicable_nations
+
+    applies_to = applicable_nations.map { |n| n["label"] }.to_sentence
+
+    if inapplicable_nations.any?
+      nations_with_alt_urls = inapplicable_nations.select { |n| n["alternative_url"] && !n["alternative_url"].empty? }
```
boffbowsh: You can just use `n["alternative_url"].present?`, this is equivalent to `!thing.nil? && !thing.blank?`

tvararu: Cheers, I had a google for a few minutes but couldn't remember it. 👍

tvararu:
🎉

PR #142, opened 2016-05-09
Add further publication features

fofr:
Depends on https://github.com/alphagov/govuk-content-schemas/pull/304 for passing tests

* Use `document_type` not `format` for title context on all formats. Publications have varying document types which need to be used as the context in the title component. For all formats without varying document types, the value of document_type will match format. Removes the old approach used by Case Studies (“format_display_type”)
* Protect against duplicate links appearing in `part_of` and `from`
* Include ministers in `from`
* Include related statistical data sets and topical events in `part_of`
* Show national statistics logo when document_type is national_statistics

Example based on https://www.gov.uk/government/statistics/affordable-housing-supply-in-england-2011-to-2012

screen shot 2016-05-09 at 11 37 07

https://trello.com/c/v85kPiKN/366-9-publications-migration-mvp-content-schema-examples-and-front-end-work-medium

### DIFF
--- a/app/presenters/linkable.rb
+++ b/app/presenters/linkable.rb
@@ -17,6 +17,7 @@ module Linkable
       world_locations
       topics
       topical_events
+      related_statistical_data_sets
     })
   end
 
--- a/app/presenters/linkable.rb
+++ b/app/presenters/linkable.rb
@@ -1,17 +1,21 @@
 module Linkable
   def from
-    links("lead_organisations") +
-      links("organisations") +
-      links("supporting_organisations") +
-      links("worldwide_organisations")
+    links_group(%w{
+      lead_organisations
+      organisations
+      supporting_organisations
+      worldwide_organisations
+    })
   end
 
   def part_of
-    links("document_collections") +
-      links("related_policies") +
-      links("policies") +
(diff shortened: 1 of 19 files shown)


fofr:
Tests passing now

app/presenters/linkable.rb
```diff
@@ -22,4 +29,13 @@ def links(type)
       link_to(link["title"], link["base_path"])
     end
   end
+
+  def links_group(types)
+    collected_links = []
+    types.each do |type|
```
boffbowsh:
```ruby
def links_group(types)
  types.map { |type| links(type) }.uniq
end
```

app/presenters/linkable.rb
```diff
@@ -22,4 +29,13 @@ def links(type)
       link_to(link["title"], link["base_path"])
     end
   end
+
+  def links_group(types)
```
boffbowsh:
```ruby
def links_group(types)
  type.flat_map { |type| links(type) }.uniq
end
```

fofr: `flat_map`, nice. Rebased.

PR #144, opened 2016-05-09
Add national applicability to publications

fofr:
* Move national applicability to module
* Make nation applicability work with any format
* Add national applicability to publications

screen shot 2016-05-09 at 17 41 33

cc @tvararu


tvararu:
👍

PR #148, opened 2016-05-16
Run full suite of tests when contract testing on CI

tijmenb:
Currently this app only runs a specific set of tests when triggered by a new PR on govuk-content-schemas. This tripped us up today because some tests that have been excluded are using the content schema examples in the tests (#147).

The tests for this repo are really fast, so it’s not a problem to run the entire test suite for each change on govuk-content-schemas. This will prevents test runs to pass by accident and makes the build process easier to understand.

This PR also updates `jenkins.sh` to be more modern.

@fofr


fofr:
Looks good to me, but leaving for others with more jenkins knowledge to review.

PR #151, opened 2016-05-18
Use correct data for HTML Publications

boffbowsh:
- Use parent's `document_type` instead of `format_sub_type` key
- Organisation branding information is nested within the linked organisation's `details` hash


gpeng:
👍  when 🍏

boffbowsh:
Will pass when https://github.com/alphagov/govuk-content-schemas/pull/311 is merged

PR #152, opened 2016-05-19
Remove references to the service manual

thehenster:
The service manual has been moved to it's own frontend 'service-manual-frontend'. The content items have been republished with the new frontend as the rendering app so we can now safely remove the service manual bits from it's previous home.

Can anyone think of a place where more service manual stuff might be hiding?


boffbowsh:
Looks good to me, except one page is still using govt frontend: https://www.gov.uk/api/content/service-manual/service-assessments

```sql
SELECT * FROM content_items 
INNER JOIN states ON states.content_item_id = content_items.id 
WHERE rendering_app = 'government-frontend' 
AND schema_name in('service_manual_guide', 'service_manual_topic') 
AND states.name = 'published';
```

thehenster:
Oo, thanks. I won't be able to fix that until next week so I'll close this for now.

thehenster:
That page should be moved to the new frontend now @boffbowsh.

fofr:
😸

PR #153, opened 2016-05-23
Add fatality notice format

nickcolley:
Depends on https://github.com/alphagov/govuk-content-schemas/pull/316 being merged for tests.

Screenshot of example content:

fatality-notice-format-render-example

Paired with @fofr 🙉 

Live example:
https://www.gov.uk/government/fatalities/flight-lieutenants-alan-scott-and-geraint-roberts-have-died

Associated ticket: https://trello.com/c/0UAMj5gj/407-8-fatality-notice-migration-implement-publishing-of-format-to-publishing-api


app/presenters/fatality_notice_presenter.rb
```diff
@@ -0,0 +1,25 @@
+class FatalityNoticePresenter < ContentItemPresenter
+  include ActionView::Helpers::UrlHelper
```
boffbowsh: Is this needed?

boffbowsh: `extend` this in `Linkable` instead

boffbowsh:
Tiny comment but it can be addressed with the other presenters in a different PR.

test/test_helper.rb
```diff
@@ -44,9 +44,10 @@ def assert_component_parameter(component, label, value)
     end
   end
 
-  def assert_has_component_title(title)
+  def assert_has_component_title(title, context = nil)
```
dsingleton: 👍

PR #155, opened 2016-05-24
Use toplevel 'withdrawn notice' property...

steventux:
https://trello.com/c/78Q9mrDe/708-make-whitehall-use-unpublishing-endpoint-medium

Any content item can be withdrawn via the unpublish endpoint on the publishing API, this means the `withdrawn_notice` property is a top level metadata property, so change `Withdrawable` to read it.
The notice will continue to be supported until we've ensured no content is affected.

Depends on https://github.com/alphagov/publishing-api/pull/340 so do not merge yet please.

### DIFF
--- a/app/presenters/case_study_presenter.rb
+++ b/app/presenters/case_study_presenter.rb
@@ -2,7 +2,6 @@ class CaseStudyPresenter < ContentItemPresenter
   include ActionView::Helpers::UrlHelper
   include Linkable
   include Updatable
-  include Withdrawable
 
   attr_reader :body
 
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -1,4 +1,6 @@
 class ContentItemPresenter
+  include Withdrawable
+
   attr_reader :content_item, :title, :description, :format, :locale, :phase, :document_type
 
   def initialize(content_item)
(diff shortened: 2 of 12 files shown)


fofr:
If all formats could be withdrawn, should `Withdrawable` move to the base content item presenter?

boffbowsh:
I think the code layout is fine - having it in `Withdrawable` makes the separation of concerns clear.

fofr:
Spoke with @boffbowsh about this offline, probably worth keeping the `Withdrawable` module but including it on the content item presenter rather than the individual formats.

steventux:
@fofr I think [this needs to happen](https://github.com/alphagov/govuk-content-schemas/pull/318) for that to happen, PR :horse_racing:

steventux:
I was just about to remove the Do not merge, but this change would not currently work for content store items with a withdrawn_notice in the details hash (alone), I think we need to migrate anything in the content store with a withdrawn notice.

fofr:
@steventux I think `include Withdrawable` is still included in a few places it needn't be.

steventux:
my grep skills let me down again.

elliotcm:
@steventux This is sat in review in our backlog, is there anything we need to do with it?

steventux:
I've identified approx 400 withdrawn case studies and detailed guides in the publishing pipeline which would need republishing to make the notice appear as a top level property. This will need to happen first. Working on a migration for this currently.

elliotcm:
OK, cool.  I'll bump it back into in progress.

PR #160, opened 2016-07-04
Update available-languages to fix the spacing on smaller viewports

nickcolley:
- Adds a margin to the bottom
- Fixes the top margin so it is inline with the govuk-title (this shouldn't be part of the component long term, since it ties it to the overall view)
- Updates styles so they're not using magic numbers and replace with gutters.

Before

image

After

image

image

### How to test
- Check this branch (`avaliable-languages-component-fix-spacing`)
- Open http://government-frontend.dev.gov.uk/guidance/prepare-a-charity-annual-return
- Change the viewport size
- Spacing should be consistent with https://gov.uk/guidance/prepare-a-charity-annual-return
- Change the viewport size

### DIFF
--- a/app/assets/stylesheets/helpers/_available-languages.scss
+++ b/app/assets/stylesheets/helpers/_available-languages.scss
@@ -1,29 +1,28 @@
 .available-languages {
-  margin-top: $gutter-half;
+  margin-bottom: $gutter;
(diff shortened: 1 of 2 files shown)


tvararu:
👌

PR #163, opened 2016-07-21
Add breadcrumbs where needed

jackscotti:
Some items have parents, so we should display the breadcrumb components when their data is available. 

I am going to leave a couple of comments (in Whitehall) to the document types that yet have to be migrated to government-frontend but that should have breadcrumbs. 

https://trello.com/c/5HYjGVif/13-breadcrumb-for-government-frontend-content-2
worked on it with @carolinegreen.

# without:
screen shot 2016-07-22 at 16 10 59

# with:
screen shot 2016-07-22 at 16 08 52

|doc type| breadcrumbs required | migration status |
| --- | --- | --- |
|case study| yes | complete |
|detailed guide| yes | in progress - almost complete |
|document collection placeholder| yes |  not started|
|generic edition| yes | not started |
|html attachment| yes | ongoing |
|publication| yes |not started |
|topical event about page| yes | complete |
|coming soon| no | not relevant |
|contact| no | not relevant |
|ministerial role| no | not relevant |
|organisation| no | not relevant |
|person| no | not relevant |
|policy area| no | not relevant |
|publish intent| no | not relevant |
|statistic announcement| no | not relevant |
|take part| no | not relevant |
|topical event| no | not relevant |

### DIFF
--- a/app/presenters/breadcrumbs.rb
+++ b/app/presenters/breadcrumbs.rb
@@ -1,0 +1,18 @@
+module Breadcrumbs
+  def breadcrumbs
+    return [] unless parent
+
+    direct_parent = parent
+    ordered_parents = []
+
+    while direct_parent
+      ordered_parents.unshift(
+        title: direct_parent["title"],
+        url: direct_parent["base_path"],
+      )
+      direct_parent = direct_parent["parent"] && direct_parent["parent"].first
+    end
+
+    ordered_parents.unshift(title: "Home", url: "/")
+  end
+end
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -18,6 +18,12 @@ class ContentItemPresenter
     sorted_locales(@content_item["links"]["available_translations"])
   end
 
+  def parent
+    if content_item["links"].include?("parent")
+      content_item["links"]["parent"][0]
+    end
+  end
+
 private
 
   def display_time(timestamp)
--- a/test/presenters/breadcrumbs_test.rb
+++ b/test/presenters/breadcrumbs_test.rb
@@ -1,0 +1,69 @@
+require 'test_helper'
+
+class BreadcrumbsTest < ActionView::TestCase
+  test "returns empty breadcrumbs when parent is not specified" do
+    content_item = build_content_item
+    assert_equal [], content_item.breadcrumbs
+  end
+
+  test "returns empty breadcrumbs when parent is empty" do
+    links = { parent: [] }
+    content_item = build_content_item(links)
+    assert_equal [], content_item.breadcrumbs
+  end
+
+  test "returns breadcrumbs when there is a parent" do
+    links = {
+      "parent" => [
+        { "title" => "A-parent", "base_path" => "/a-parent" }
+      ]
+    }
+
+    content_item = build_content_item(links)
+
+    assert_equal [
+      { title: "Home", url: "/" },
+      { title: "A-parent", url: "/a-parent" },
+    ], content_item.breadcrumbs
+  end
+
+  test "returns breadcrumbs when there is more than one parent" do
+    links = {
+      "parent" => [
+        {
+          "title" => "A-parent",
+          "base_path" => "/a-parent",
+          "parent" => [{
+            "title" => "Another-parent",
+            "base_path" => "/another-parent",
+          }]
+        }
+      ]
+    }
+
+    content_item = build_content_item(links)
+
+    assert_equal [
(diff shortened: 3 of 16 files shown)


gpeng:
Looks ok to me. Could you add a screenshot (or two) to the PR? (Also the linter is complaining)

app/presenters/content_item_presenter.rb
```diff
@@ -1,5 +1,6 @@
 class ContentItemPresenter
   include Withdrawable
+  include BreadcrumbHelper
```
dsingleton: I don't _think_ this needs to be explicitly included, the other helpers aren't, but are available to the view.

dsingleton: Actually, shouldn't this be a module, like `Linkable` or `Widthdrawable`, rather than a helper?

dsingleton:
Generally, 👍 , it's nice to standardise the breadcrumb behaviour and logic.

It's not a blocker, but it feels like the Breadcrumb helper should be included in the presenter of each format that supports breadcrumbs - in the same way `Linkable` is.  IMO this is more explicit about what traits each format has, based on reading the presenter. What do you think?

jackscotti:
@dsingleton I see why you are saying what you are saying and I think you are right. Updating as suggested.

jackscotti:
@dsingleton give a look at the last commit that I've updated. Is `Breadcrumbs` an ok name for a module or should it be `Breadcrumbable` (😢) ?

jackscotti:
(I realised I did not squash the commits properly, will clean it up tomorrow morning)

gpeng:
I could go with either. There is a mix of naming in the presenter already. `Breadcrumbable` does make you pause over the spelling though. We've already spelled it in two different ways in this PR.

tijmenb:
@jackscotti I see we're adding this to specific presenters. Any reason why we can't add it to all?

jackscotti:
@tijmenb the reason why I did not add it to all document types is that we are not publishing parent date for some of them from Whitehall to Content-Store. 

Adding it to all will not create any problem but it will not provide any benefit until all document types will be published with a parent. I would leave it as is, and would add it to all documents when they will be all migrated to government-frontend, what do you recon?

test/presenters/breadcrumbs_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    assert_equal [
+      { title: "Home", url: "/" },
+      { title: "Another-parent", url: "/another-parent" },
+      { title: "A-parent", url: "/a-parent" },
+    ], content_item.breadcrumbs
+  end
+end
+
+class ItemWithBreadrumbs < ContentItemPresenter
```
Davidslv: can you move this to the top of this file, it's better to read that you created this class in this file before the tests

jackscotti:
Updated, I've also refactored the TopicalEventAboutPagePresenter to use the breadcrumb module in the last commit.

@dsingleton @gpeng @tijmenb

jackscotti:
this is ready to be merged now.

app/presenters/breadcrumbs.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    direct_parent = parent
+    ordered_parents = []
+
+    while direct_parent
+      ordered_parents.unshift(
+        title: direct_parent["title"],
+        url: direct_parent["base_path"],
+      )
+      direct_parent = direct_parent["parent"] && direct_parent["parent"].first
```
MatMoore: Does this mean there can only ever be one parent in this array, or are we explicitly ignoring everything but the first?

app/presenters/content_item_presenter.rb
```diff
@@ -18,6 +18,12 @@ def available_translations
     sorted_locales(@content_item["links"]["available_translations"])
   end
 
+  def parent
+    if content_item["links"].include?("parent")
+      content_item["links"]["parent"][0]
```
MatMoore: Should this method be provided by `Breadcrumb`? Both of these work together and have the same logic for extracting the first parent.

MatMoore:
Added a few minor comments but looks good to me, nice work. 👍  merge when ready.

app/presenters/content_item_presenter.rb
jackscotti: #parent is used for other things not related to breadcrumbs. So would leave it separate.

app/presenters/breadcrumbs.rb
jackscotti: Parent is a nested hash of parents, this method traverses it until it reaches an item that does not have a parent.

PR #165, opened 2016-07-28
Updates slimmer to 9.3.1 to enable full GOVUK Component i18n support

nickcolley:


nickcolley:
We could update slimmer to fix the tests rather than individual apps? Since we include slimmer test helpers.

boffbowsh:
Fixing the test helpers within slimmer sounds like the best idea 👍

PR #169, opened 2016-08-10
Update Rails to 5.0

nickcolley:
Updated to 4.2.7 then to 5.0.0

Adopts approach documented at http://guides.rubyonrails.org/upgrading_ruby_on_rails.html
which introduces new files and changes of old files in accordance with
the Rails application generator.

When comparing the diffs http://meldmerge.org/ was useful in this case.

When updating to 5.0.0 the phase label tests started failing, after digging
it was determined that the test relies on buggy behaviour that rails output the
partial name as a local which has been fixed here rails/rails@bef9484

Shoutout to https://github.com/andrewgarner for helping debug this 🌠


bin/update
```diff
@@ -0,0 +1,29 @@
...
+chdir APP_ROOT do
+  # This script is a way to update your development environment automatically.
+  # Add necessary update steps to this file.
+
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
+  system('bundle check') || system!('bundle install')
+
+  puts "\n== Updating database =="
+  system! 'bin/rails db:migrate'
```
andrewgarner: We should comment these two lines out. It will currently fail as we're not requiring active record.

config/application.rb
```diff
@@ -41,5 +42,10 @@ class Application < Rails::Application
 
     # Path within public/ where assets are compiled to
     config.assets.prefix = '/government-frontend'
+
+    # Do not swallow errors in after_commit/after_rollback callbacks.
+    # config.active_record.raise_in_transactional_callbacks = true
+
+    GdsApi.config.always_raise_for_not_found = true
```
andrewgarner: I think this should belong in `initializers/gds_api.rb` since it's related to configuration of the library and not the application.

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
andrewgarner: We haven't checked in a sample `database.yml` config so I would follow that precedence and do the same with the example action cable config.

config/initializers/new_framework_defaults.rb
```diff
@@ -0,0 +1,23 @@
+# Be sure to restart your server when you modify this file.
+#
+# This file contains migration options to ease your Rails 5.0 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
+#
+# Read the Rails 5.0 release notes for more info on each option.
+
+# Enable per-form CSRF tokens. Previous versions had false.
+Rails.application.config.action_controller.per_form_csrf_tokens = false
```
andrewgarner: The intention with this file is to suppress the new defaults in Rails 5 until we have confirmed that we can safely apply them. By committing this file it implies we are unable to apply the new defaults?

config/puma.rb
```diff
@@ -0,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
```
andrewgarner: I would exclude this file as we already depend on `unicorn` as a HTTP server.

config/application.rb
andrewgarner: On second thoughts, the deprecation warning is telling us we should no longer use `get_json`, `post_json` etc, instead we should use `get_json!`, `post_json!` etc.

Can we update our code so that we don't need to configure this option? If we can easily do it now we won't need to revisit before the 1st October 2016 when the feature is due to be removed.

config/environments/production.rb
```diff
@@ -30,41 +25,43 @@
...
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
   config.log_level = :info
 
   # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
-
-  # Use a different logger for distributed setups.
-  # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr))
```
andrewgarner: We need to add this back on line 78

config/environments/production.rb
```diff
@@ -73,15 +70,19 @@
...
 
-  # Disable automatic flushing of the log to improve performance.
-  # config.autoflush_log = false
-
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
+  # Use a different logger for distributed setups.
+  # require 'syslog/logger'
+  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
```
andrewgarner:
```rb
config.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr))
```

config/environments/production.rb
nickcolley: good spot

PR #170, opened 2016-08-10
Add migration status of formats to readme

tijmenb:
This adds a table to the readme with the migration status for the formats this app can render (I've used all presenters classes for this).

For an outsider it's quite difficult to find out which things are in progress and which are "real".

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,24 @@ parts of the [Whitehall](https://github.com/alphagov/whitehall) application.
 
 ![A  Case Study](https://raw.githubusercontent.com/alphagov/government-frontend/master/docs/assets/case-study-screenshot.png)
 
-## Live examples
+## Formats
 
-- [gov.uk/government/case-studies/2013-elections-in-swaziland](https://www.gov.uk/government/case-studies/2013-elections-in-swaziland)
-- [gov.uk/government/statistics/announcements/diagnostic-imaging-dataset-for-september-2015](https://www.gov.uk/government/statistics/announcements/diagnostic-imaging-dataset-for-september-2015)
+Not all formats that this app can handle are rendered by it in production.
+
+| Format | Example | Production status |
+|---|---|---|
+| Case study | [View example](https://www.gov.uk/government/case-studies/2013-elections-in-swaziland) | Migrated |
+| Coming soon | | Rendered by Whitehall |
+| Detailed guide | | Rendered by Whitehall |
+| Document collection | | Rendered by Whitehall |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,10 +8,24 @@ parts of the [Whitehall](https://github.com/alphagov/whitehall) application.
...
+| Coming soon | | Rendered by Whitehall |
+| Detailed guide | | Rendered by Whitehall |
+| Document collection | | Rendered by Whitehall |
+| Fatality notice | | Rendered by Whitehall |
+| HTML Publication | | Rendered by Whitehall |
+| Publication | | Rendered by Whitehall |
+| Statistics announcement | [gov.uk/government/statistics/announcements/diagnostic-imaging-dataset-for-september-2015](https://www.gov.uk/government/statistics/announcements/diagnostic-imaging-dataset-for-september-2015) | Migrated |
+| Take part | [www.gov.uk/government/get-involved/take-part/become-a-councillor](https://www.gov.uk/government/get-involved/take-part/become-a-councillor) | Migrated |
+| Topical event about page | [gov.uk/government/topical-events/2014-overseas-territories-joint-ministerial-council/about](https://www.gov.uk/government/topical-events/2014-overseas-territories-joint-ministerial-council/about) | Migrated |
+| Unpublishing | | Rendered by Whitehall |
```
boffbowsh: This isn't a format

tijmenb: Are you sure? Unpublishings have a [schema](https://github.com/alphagov/govuk-content-schemas/tree/master/dist/formats/unpublishing) and [presenter & views in this app](https://github.com/alphagov/government-frontend/blob/master/app/presenters/unpublishing_presenter.rb). Should those be removed?

boffbowsh: It's a state of a document rather than a format itself. But I guess it can live here for completeness.

tijmenb: I don't understand. What is a format then? There's content in the content-store with format `unpublishing` ([see](https://www.gov.uk/api/content/government/case-studies/stuart-scott)), there's a schema and this app could render the documents (but doesn't). Seems like a format.

boffbowsh: It's a really murky area at present, there's work going on in Whitehall, Publishing API and Content Store at present to sort it out a bit. I don't _think_ it'll be a format by itself in the end, and I also don't think it'll end up being rendered by `government-frontend`. It can stay here for the moment though while it gets sorted out.

tijmenb: Ah allright. Thanks! Amended the commit to say "might not be migrated".

PR #174, opened 2016-09-05
Updates Wraith to include up to date paths to test

nickcolley:
- Adds specific versions of tools to install that have been tested and work
- Updates config and usage to new Wraith version
- Updates the paths used by wraith to include current migrated formats, with `govuk_content_schemas` examples
- Updates the paths to include frequently hit pages for each format

Note that some commits related to this were pushed to `master` mistakenly, I've squashed them in this PR, if we want we could fix that on master or just roll forwards here. 😫


.gitignore
```diff
@@ -14,3 +14,7 @@
 # Ignore all logfiles and tempfiles.
 /log/*.log
 /tmp
+
+.DS_Store
```
andrewgarner: This isn't application specific so should probably be left for a global configuration such as [https://github.com/github/gitignore/blob/master/Global/macOS.gitignore]

README.md
```diff
@@ -69,20 +69,24 @@ Or to specify the location explicitly:
...
 comparison tool") to generate a visual diff to compare rendering changes in this
-application. Wraith has some dependencies you'll [need to install](http://bbc-news.github.io/wraith/os-install.html).
+application.
 
-First, on `master` branch, run:
+On the [Development VM](https://github.gds/gds/development) navigate to
+the project directory and run
 ```
-wraith history test/wraith/config.yaml
+sudo npm install --global phantomjs-prebuilt@2.1.7
```
andrewgarner: How about we add a `Gemfile` and `package.json` to `test/wraith` that allows us to explicitly declare these?

nickcolley: Could do, I normally prefer dependencies to be local to the project 👍

PR #176, opened 2016-09-14
Use shared helpers for breadcrumbs

MatMoore:
gds-api-adapters returns us a content item object that is just
a hash of data. We've created a shared helper to take this and extract the data
needed to render the govuk breadcrumb component.

This also adds a default "Home" breadcrumb to any pages where the template
for the format includes breadcrumbs, but the content item doesn't
have a parent link.

An example of this is https://www.gov.uk/government/case-studies/competition-law-case-studies
screen shot 2016-09-14 at 17 07 13

https://trello.com/c/mOeDK914/107-epic-related-links-and-panopticon


tijmenb:
@MatMoore I've updated this PR to use the latest version of the gem. I guess it's ready for review/merge?

MatMoore:
@tijmenb yep, ready for review.

Would be good if someone with context on topical event about pages could take a look too. The way it's adding "archived" into the topical event part of the breadcrumb is a pattern that isn't used for any other format (see comments in app/presenters/topical_event_about_page_presenter.rb).

I was wondering if there is a real need for this - if so we might need to think about building this into the component, but it doesn't make a lot of sense to me. I'm not sure if archiving is the same as withdrawing for this format, and why we don't just "archive" the about page if the topical event is archived.

PR #180, opened 2016-09-16
Fix breadcrumb presenter

tijmenb:
This fixes a bug in the breadcrumbs presenter.

The presenter is tested with an example from govuk-content-schemas. But the example contains an error: it uses a structure like `links -> parent -> parent`. This is not how the publishing-api expands the links, it nests it like `links -> parent -> links -> parent`.

However, because the liberal way the frontend `links` hash is validated, the example was considered valid. This is tackled in https://github.com/alphagov/govuk-content-schemas/pull/389.

### DIFF
--- a/app/presenters/breadcrumbs.rb
+++ b/app/presenters/breadcrumbs.rb
@@ -10,7 +10,7 @@ module Breadcrumbs
         title: direct_parent["title"],
         url: direct_parent["base_path"],
       )
-      direct_parent = direct_parent["parent"] && direct_parent["parent"].first
+      direct_parent = direct_parent = direct_parent.dig("links", "parent", 0)
     end
 
     ordered_parents.unshift(title: "Home", url: "/")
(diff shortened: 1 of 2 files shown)


app/presenters/breadcrumbs.rb
```diff
@@ -10,7 +10,7 @@ def breadcrumbs
         title: direct_parent["title"],
         url: direct_parent["base_path"],
       )
-      direct_parent = direct_parent["parent"] && direct_parent["parent"].first
+      direct_parent = direct_parent = direct_parent.dig("links", "parent", 0)
```
boffbowsh: TIL [`Hash#dig`](http://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig) 👏

tijmenb:
This has a circular dependency on the example changes in https://github.com/alphagov/govuk-content-schemas/pull/389. Mind merging either this or that PR while still red?

PR #190, opened 2016-10-13
Display logo image for Detailed guides if provided

boffbowsh:
One Detailed guide has a European Structural and Investment Fund logo
which must be displayed for regulatory reasons. This was missed during
migration.

How it looked in Whitehall: http://webarchive.nationalarchives.gov.uk/20160806171559/https://www.gov.uk/guidance/england-2014-to-2020-european-structural-and-investment-funds

Screenshots of Government Frontend at various break points:

screen shot 2016-10-13 at 12 02 20
screen shot 2016-10-13 at 12 02 34
screen shot 2016-10-13 at 12 02 47

### DIFF
--- a/app/assets/stylesheets/views/_detailed-guide.scss
+++ b/app/assets/stylesheets/views/_detailed-guide.scss
@@ -30,4 +30,13 @@
   .back-to-content {
     margin-left: $gutter-half;
   }
+
+  .logo-image {
+    max-width: 70%;
+
+    @include media(tablet) {
+      max-width: 100%;
+      margin-top: 45px;
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


tvararu:
Note to self: also should add an `alt` attribute later, but needs more investigation to see if we're going to mark it as decorative or if we can glean some information about the image being used.

app/assets/stylesheets/views/_detailed-guide.scss
```diff
@@ -30,4 +30,13 @@
   .back-to-content {
     margin-left: $gutter-half;
   }
+
+  .logo-image {
+    max-width: 70%;
+
+    @include media(tablet) {
+      max-width: 100%;
+      margin-top: 45px;
```
fofr: I don't like the magic number here, I’m surprised linting didn't catch it.

fofr:
Images provided by content item can have alt text and captions:
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json#L60-L80

Whitehall used to render with the alt tag "European structural investment funds".

PR #195, opened 2016-11-16
Fix DocumentCollection contents styling

gpeng:
This protects the first 'third' column in the event that there are no contents entries.

Also refactors the relevant class out into the `sidebar-with-body` helper which fixes a bug in working groups and removes some duplication.


fofr:
Screenshot showing the collapsed empty column vs existing content. The column should not collapse.

screen shot 2016-11-16 at 14 13 39

fofr:
LGTM :+1:

PR #196, opened 2016-11-17
Expose the content_id for each content item

mgrassotti:
Trello: https://trello.com/c/nYKudXNT/512-fix-whitehall-admin-bookmarklet-for-migrated-formats-medium

In order to preserve the whitehall admin bookmarklet functionality, we
are going to expose the content_id for each content item. This id will
be used to redirect to the related admin page.


boffbowsh:
Can you add a test please?

PR #197, opened 2016-11-17
Handle missing `links#documents` for Document Collections

gpeng:
When all of the documents in a collection have been unpublished the `links#documents` element is not present in the json. This commit handles that case.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/438

### DIFF
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -48,6 +48,6 @@ private
   end
 
   def documents_hash
-    @documents_hash ||= content_item["links"]["documents"].map { |d| [d["content_id"], d] }.to_h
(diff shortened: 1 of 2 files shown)


app/presenters/document_collection_presenter.rb
```diff
@@ -48,6 +48,6 @@ def group_title_id(title)
   end
 
   def documents_hash
-    @documents_hash ||= content_item["links"]["documents"].map { |d| [d["content_id"], d] }.to_h
+    @documents_hash ||= (content_item["links"]["documents"] || []).map { |d| [d["content_id"], d] }.to_h
```
andrewgarner: maybe `Array(content_item["links"]["documents"])`?

PR #199, opened 2016-11-21
Add Statistical Data Sets

bevanloon:
Add support for rendering Statistical Data Sets

[Trello](https://trello.com/c/3OZ36DPk)

Paired with @fofr and @tvararu 

Related pull request in [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/pull/440)

### Before 
statistical-data-set-before

### After
statistical-data-set-after

## Example with block attachment
An example with block attachments has also been created - screenshots below show the block attached element:

### Before
before_block_attachment

### After
afetr_block_attachment

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -34,3 +34,4 @@
 @import "views/publication";
 @import "views/document-collection";
 @import "views/fatality-notice";
+@import "views/statistical-data-set";
--- a/app/assets/stylesheets/views/_statistical-data-set.scss
+++ b/app/assets/stylesheets/views/_statistical-data-set.scss
@@ -1,3 +1,5 @@
 .statistical-data-set {
-
+  @include description;
+  @include sidebar-with-body;
+  @include history-notice;
 }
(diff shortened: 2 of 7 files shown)


fofr:
Looks good to me :+1:, though leaving final review/merge to someone who hasn't worked on it.

binaryberry:
Would it be useful adding an example with attachments in your opinion @fofr ? 
https://www.gov.uk/government/statistical-data-sets/fe-data-library-apprenticeship-vacancies?cachebust=1480075134&preview=678807

CI tests are intermittently failing too - seems there's an issue finding the corresponding schema.

Other than that, looks good and I'm happy to approve.

fofr:
@binaryberry You’re right, the examples should include attachments – otherwise it'd be a statistical data set without any data sets.

bevanloon:
@binaryberry @fofr added an example with block attachments in [govuk-content-schemas]( https://github.com/alphagov/govuk-content-schemas/commit/ad7a3a9)

bevanloon:
@binaryberry CI builds - @Rosa-Fox @cbaines and I investigated this and changed the `Choosing strategy` option in Jenkins from `Inverse` to `Default` which we think has fixed the problem. 

It looks like the `Inverse` strategy builds all branches **except** the one specified (https://raw.githubusercontent.com/ooyala/jenkins-git-plugin/master/src/main/resources/hudson/plugins/git/GitSCM/help-choosingStrategy.html).

PR #201, opened 2016-11-29
Render the consultation format

fofr:
Tests will fail until the corresponding schema changes have been merged and deployed: https://github.com/alphagov/govuk-content-schemas/pull/445

Main story:
https://trello.com/c/QseqtCaJ/509-consultations-migration-basic-content-schema-examples-and-front-end-work-large

Ports the following features from Whitehall:
* Display of consultation description (the body in content item)
* Display of consultation summary (the description in content item)
* Open and closing date and date ranges
  * Includes fix for showing an unopened consultation as "opening on" rather than already closed
  * Uses a legible date format, "10 June 2015 10:00am" is now "10am on 10 June 2015"
  * When the time is 12:00am, it shows midday
  * When the time is midnight it either hides the time for opening date or rolls back 1 minute for closing – this remove ambiguity
* A list of supporting documents using the same pattern as publications
* A ways to respond section
  * Only displays when open and when a respond online link, email or postal address have been provided. This matches existing logic. If an attachment is uploaded but no email or postal address are provided to return the completed copy it will not display. This is how it is in Whitehall but could be improved.
* An outcomes section
* An outcome documents section
* A public feedback section
* A public feedback documents section
* A notice displaying when a consultation is unopened, pending feedback, or closed with outcome
* National applicability, history mode, withdrawable
* Includes fix for publication titles at thin viewports, the same bug existed for consultations
* Renames `display_time` to `display_date` as the format provided does not include time
* Fixes display and presentation of unopened consultations: https://trello.com/c/GEyT6hzk/518-make-it-obvious-when-a-consultation-starts-small
* Consultation banner colour has been switched from purple to blue

## Ways to respond problems

Port of existing copy which isn’t great. Fixes the problem of "Respond online" looking like a heading, instead using the call to action pattern. Described here: https://trello.com/c/pEccXKHV/517-issue-with-ways-to-respond-section-in-consultations-medium

This highlights a problem with how we render content: this HTML is crafted from values provided by users, rather than govspeak – yet it is still content and must be rendered as content using the same styles. For now this means passing that HTML to the govspeak content for rendering.

An alternative would be some free text the user provides in markdown. This would make the format more flexible but less uniform.

## Headings structure

This has been ported from Whitehall, but has accessibility problems. We use `h1` for each section heading – this is the only way to guarantee that we provide a heading for dynamic content that appears in markdown – eg users could use `## level 2 headings`.

However, for a closed consultation we add an `original consultation` heading, which should wrap the whole of the original consultation. This is currently an h2. It could be an h1 but that would have the same problem. I think this should be looked at separately to this PR.

## Whitehall

If the migration of this format takes a while, which given its complexity I suspect it will, the fixes to this format should be back ported to Whitehall. ie readable date formats, blue box, correct handling of an unopened consultation.

## Omissions

* The share features have been split out into a new story: https://trello.com/c/MaZ7t0zi/753-add-share-buttons-to-consultations-on-government-frontend
* Specific testing for withdrawn state and history mode – some generic test helpers should be written to cover these scenarios without having to rely on examples

## Screenshots

| Feature | Old | New |
| --- | --- | --- |
| Open consultation | open_consultation_old | open_consultation_new
| Closed consultation | closed_consultation_old | closed_consultation_new
| Unopened consultation | unopened_consultation_old | unopened_consultation_new
| Ways to respond | ways_to_respond_consultation_old | ways_to_respond_consultation_new
| Pending feedback | pending_feedback_consultation_old | pending_feedback_consultation_new
| Mobile | consultation_mobile_old | consultation_mobile_new

### DIFF
--- a/app/assets/stylesheets/views/_consultation.scss
+++ b/app/assets/stylesheets/views/_consultation.scss
@@ -1,3 +1,67 @@
 .consultation {
+  @include sidebar-with-body;
+  @include history-notice;
+  @include withdrawal-notice;
 
+  .section-title {
+    @include bold-27;
+    margin-top: $gutter;
+    margin-bottom: $gutter-half;
+
+    @include media(tablet) {
+      margin-top: 0;
+      margin-bottom: $gutter;
+    }
+  }
+
+  .consultation-notice {
+    @include notice;
+  }
+
+  .consultation-banner {
+    @include responsive-bottom-margin;
+    @include core-19;
+    @include white-links;
+    background-color: $govuk-blue;
+    color: $white;
+    padding: $gutter-half;
+
+    @include media(tablet) {
+      padding: $gutter;
+    }
+
+    h2 {
+      @include bold-27;
+      margin-bottom: $gutter-one-third;
+    }
+
+    .consultation-dates {
+      @include media(tablet) {
+        padding-right: $gutter;
+      }
+    }
+
+    .consultation-summary {
+      margin-top: $gutter-half;
+
+      @include media(tablet) {
+        margin-top: 0;
+        margin-left: -$gutter-one-third;
+        padding-right: 0;
+        border-left: 1px solid $white;
+      }
+    }
+
+    .consultation-date {
+      @include core-24;
+    }
+  }
+
+  .original-consultation {
+    border-top: 1px solid $border-colour;
+
+    @include media(tablet) {
+      padding-top: $gutter;
+    }
+  }
 }
--- a/app/assets/stylesheets/views/_consultation.scss
+++ b/app/assets/stylesheets/views/_consultation.scss
@@ -1,0 +1,3 @@
+.consultation {
+
+}
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -28,16 +28,20 @@
 </div>
 <%= render 'shared/history_notice', content_item: @content_item %>
 
-<% if @content_item.pending_final_outcome? %>
-<div class="consultation-notice">
-  <h2>
-    We are analysing your feedback
-  </h2>
-  <p>Visit this page again soon to download the outcome to this public&nbsp;feedback.</p>
-</div>
-<% end %>
-
-<% if @content_item.final_outcome? %>
+<% if @content_item.unopened? %>
+  <div class="consultation-notice">
+    <h2>
+      This consultation opens at <time datetime="<%= @content_item.opening_date_time %>"><%= @content_item.opening_date %></time>
+    </h2>
+  </div>
+<% elsif @content_item.pending_final_outcome? %>
+  <div class="consultation-notice">
+    <h2>
+      We are analysing your feedback
+    </h2>
+    <p>Visit this page again soon to download the outcome to this public&nbsp;feedback.</p>
+  </div>
+<% elsif @content_item.final_outcome? %>
   <div class="consultation-notice">
     <h2>
       This consultation has concluded
@@ -71,9 +75,7 @@
           direction: page_text_direction %>
     </div>
   </div>
-<% end %>
 
-<% if @content_item.final_outcome? %>
   <section class="original-consultation">
     <h2 class="section-title">Original consultation</h2>
 <% end %>
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -1,0 +1,12 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, @content_item.title %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        context: t("content_item.format.#{@content_item.document_type}", count: 1),
+        title: @content_item.title %>
+  </div>
+</div>
+
+<%= render 'shared/description', description: @content_item.description %>
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -109,7 +109,9 @@
 
 <% if @content_item.final_outcome? %>
   <section class="original-consultation">
-    <h2 class="section-title">Original consultation</h2>
+    <header>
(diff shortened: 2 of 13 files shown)


nickcolley:
For the issues you've mentioned can you raise them as a separate issue so we can track it and resolve it later?

fofr:
@nickcolley I will do when this gets merged.

app/assets/stylesheets/views/_consultation.scss
```diff
@@ -0,0 +1,67 @@
...
+    margin-top: $gutter;
+    margin-bottom: $gutter-half;
+
+    @include media(tablet) {
+      margin-top: 0;
+      margin-bottom: $gutter;
+    }
+  }
+
+  .consultation-notice {
```
andrewgarner: This shouldn't need to be nested within `.consultation` given the `.consultation-` prefix?

app/views/content_items/consultation.html.erb
```diff
@@ -0,0 +1,232 @@
...
+    <h2>
+      This consultation opens <% if @content_item.opening_date_midnight? %>on<% else %>at<% end %> <time datetime="<%= @content_item.opening_date_time %>"><%= @content_item.opening_date %></time>
+    </h2>
+  </div>
+<% elsif @content_item.pending_final_outcome? %>
+  <div class="consultation-notice">
+    <h2>
+      We are analysing your feedback
+    </h2>
+    <p>Visit this page again soon to download the outcome to this public&nbsp;feedback.</p>
```
andrewgarner: There is subtle presentational behaviour here that is mixed in with the content. How about the following?

```html
<style>
  .nowrap { white-space: nowrap; }
</style>

<p>Visit this page again soon to download the outcome to this <span class="nowrap">public feedback.</span</p>
```

app/views/content_items/consultation.html.erb
```diff
@@ -0,0 +1,232 @@
...
+            content: @content_item.final_outcome_documents,
+            direction: page_text_direction %>
+      </div>
+    </div>
+  <% end %>
+
+  <div class="grid-row sidebar-with-body">
+    <div class="column-third">
+      <h1 class="section-title" id="final-outcome-detail-title">
+        <%= t("consultation.final_outcome_detail") %>
```
andrewgarner: We're not consistently translating strings in this file. Should for example "Download the full outcome", "This consultation has concluded" also be translated?

app/views/content_items/consultation.html.erb
```diff
@@ -0,0 +1,232 @@
...
+<% if @content_item.final_outcome? %>
+  <section class="original-consultation">
+    <h2 class="section-title">Original consultation</h2>
+<% end %>
+
+<div class="consultation-banner">
+  <div class="grid-row">
+    <div class="column-third consultation-dates">
+      <% if @content_item.closed? %>
+        <p>
```
andrewgarner: You don't need to duplicate the `<p></p>` tags as they remain consistent for each case.

test/integration/consultation_test.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  test "open consultation" do
+    setup_and_visit_content_item('open_consultation')
+
+    assert page.has_text?("Open consultation")
+    assert page.has_text?("closes at 4pm on 16 December 2216")
+  end
+
+  test "un-opened consultation" do
```
andrewgarner: "future consultation"?

app/assets/stylesheets/views/_consultation.scss
fofr: The current pattern is to have all styles for a format nested within their format name.

app/views/content_items/consultation.html.erb
fofr: Or it could use `nbsp_between_last_two_words`:
https://github.com/alphagov/government-frontend/blob/master/app/helpers/typography_helper.rb#L2

app/views/content_items/consultation.html.erb
fofr: This format isn't translated. I think the simplest thing to do would be to not translate any. Some of the date formatting isn't easily translatable in its current form. What do you think?

app/views/content_items/consultation.html.erb
andrewgarner: 👍 

If that's the approach adopted elsewhere then yes we should use that here too. We're still maybe mixing content and presentation so maybe that's something another time.

app/assets/stylesheets/views/_consultation.scss
andrewgarner: Fair enough. As someone new to the code base it wasn't clear if we've adopted a [BEM](getbem.com) type methodology.

app/views/content_items/consultation.html.erb
fofr: Fixed in 185a52a

app/views/content_items/consultation.html.erb
fofr: Fixed in c19c2cd

test/integration/consultation_test.rb
fofr: Fixed and rebased. I was about to suggest the same :smile:

andrewgarner:
Upstream the "state" of a consultation is all held within a single schema. We've chosen that path due to the way we integrate with our search index but we shouldn't to let that leak down into our frontend.

We could lift out the state of a consultation so that the rules and things that change between each state are more clear. For example we could have an `OpenConsultationPresenter` and a `ClosedConsultationPresenter` and a factory method to instantiate the correct one. We could also [nest our view code](http://guides.rubyonrails.org/layouts_and_rendering.html#using-nested-layouts) giving us less conditionals.

These are enhancements however so we shouldn't block this PR to think about them.

andrewgarner:
@fofr in the example screenshots the "Unopened consultation" example has documents listed in the before screenshot but not in the after screenshot. Is that a deliberate change?

fofr:
@andrewgarner That's an accidental omission from the example.

PR #202, opened 2016-11-29
Use `schema_name` consistently

tijmenb:
The content item attribute `format` is deprecated.

alphagov/govuk-content-schemas#444 attempts to remove `format` from the frontend, but the tests for this application fail without `format` in the examples. This is because the tests are testing kind of the wrong thing.

For example:

```
assert_equal schema_item['format'], presented_item.format
```

Checks the string in `presented_item.format` which actually points to `schema_item['schema_name']`. Because they're `format` and `schema_name` are identical in the examples this worked. Now that we've removed `format` it doesn't anymore.

https://trello.com/c/C8A9bnuO/344-work-on-making-the-document-type-concept-clearer

### DIFF
--- a/test/presenters/case_study_presenter_test.rb
+++ b/test/presenters/case_study_presenter_test.rb
@@ -9,7 +9,7 @@ class CaseStudyPresenterTest < PresenterTestCase
 
   test 'presents the basic details of a content item' do
     assert_equal schema_item['description'], presented_item.description
-    assert_equal schema_item['format'], presented_item.format
+    assert_equal schema_item['schema_name'], presented_item.format
     assert_equal schema_item['locale'], presented_item.locale
     assert_equal schema_item['title'], presented_item.title
     assert_equal schema_item['details']['body'], presented_item.body
--- a/test/presenters/coming_soon_presenter_test.rb
+++ b/test/presenters/coming_soon_presenter_test.rb
@@ -7,7 +7,7 @@ class ComingSoonPresenterTest < PresenterTestCase
 
   test 'presents the basic details required to display a coming soon item' do
     assert_equal schema_item['title'], presented_item.title
-    assert_equal schema_item['format'], presented_item.format
+    assert_equal schema_item['schema_name'], presented_item.format
     assert_equal schema_item['locale'], presented_item.locale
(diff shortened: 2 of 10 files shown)


app/presenters/national_applicability.rb
```diff
@@ -23,8 +23,8 @@ def applies_to
 
 private
 
-  def translated_format_name(count)
-    I18n.t("content_item.format.#{format}", count: count).downcase
+  def translated_schema_name(count)
+    I18n.t("content_item.format.#{schema_name}", count: count).downcase
```
andrewgarner: Should we update the locale key to `content_item.schemas.x`?

cbaines:
It's only useful for the frontend to have a concept of the schema_name if it needs to do validation, as that is what it should use as the filename of the schema. I think its best to use schema_name for this (its intended purpose) and no other purpose, as this will help avoid confusion. Is using the schemas something that is planned for the government-frontend, as I can't see any use of them within the app currently?

app/presenters/national_applicability.rb
tijmenb: Yes, probably. It does feel a little more risky though, and afaik they were imported from Whitehall originally so I'm a bit hesitant to touch them as part of this PR.

tijmenb:
@cbaines this app determines the layout of the page depending on `schema_name`. I think it's worth having a discussion about that (should it be `document_type`, or even a new `template_name`?). Probably out of scope for this PR though.

app/presenters/national_applicability.rb
andrewgarner: Fair enough. It might be worth leaving a comment in the code so that future intend is not forgotten.

tijmenb:
I've updated the commit to only change the bare minimum to get the tests on `govuk-content-schemas` passing. Removed the changes to make it use the `schema_name` consistently, because there's still translation keys called `content_item.format` (https://github.com/alphagov/government-frontend/pull/202#discussion_r90010580) and I don't have enough confidence in the tests to also fix those.

The original branch is here:

https://github.com/alphagov/government-frontend/compare/use-schema-name-consistently

andrewgarner:
👍

PR #203, opened 2016-12-02
Show closing date on unopened consultations

fofr:
The date that the consultation closes is useful information for users.

screen shot 2016-12-02 at 11 42 56

@andrewgarner


andrewgarner:
Should / can you get the dividing line to fill the height of the container (minus the padding)?

fofr:
Not easily, I don't think the complexity it'd add justifies it. This matches what's live. I toyed with not including the border at all when I was building the template, which would look like this:

screen shot 2016-12-02 at 11 58 32

andrewgarner:
👍

PR #204, opened 2016-12-05
Add lang attribute to page translation links

rubenarakelyan:
This commit adds the HTML lang attribute to links to translated content. This ensures that the language name is understood correctly by screen readers and browsers, and therefore read correctly. Fixes https://github.com/alphagov/whitehall/issues/2873.

### DIFF
--- a/app/views/shared/_available_languages.html.erb
+++ b/app/views/shared/_available_languages.html.erb
@@ -6,7 +6,7 @@
           <% if translation["locale"] == I18n.locale.to_s %>
             <%= native_language_name_for(translation["locale"]) %>
           <% else %>
-            <%= link_to native_language_name_for(translation["locale"]), translation["base_path"] %>
+            <%= link_to native_language_name_for(translation["locale"]), translation["base_path"], lang: translation["locale"] %>
           <% end %>
         </li>
       <% end %>


nickcolley:
Thanks a lot for this @rubenarakelyan 👍 

Do we have an example page where this is rendered?

rubenarakelyan:
Yup, https://www.gov.uk/government/case-studies/bahrain-country-case-study-update is a good example.

nickcolley:
I've tested this in https://government-frontend-pr-204.herokuapp.com/government/case-studies/bahrain-country-case-study-update and it looks good.

However when I try it with VoiceOver nothing is read out, however when I try the http://www.bbc.com/arabic which I know is marked up correctly so it works that's not working either.

So It'd be good to get this tested with a screenreader but looks good to me.

rubenarakelyan:
I think this needs to be tried out with a standalone screenreader, since my research makes me believe that VoiceOver can't automatically switch languages.

nickcolley:
As you can see here https://drive.google.com/file/d/0B-GzTJBVVSNzcnpQa2d6Nk5fVjA/view the iPad Pro that's using VoiceOver can switch between them so I guess they work differently!

rubenarakelyan:
Yeah, I read that iOS can do this but OS X can't.

rubenarakelyan:
@nickcolley Are you happy for this to be merged or do you want some more testing?

cfq:
Can confirm it works as intended with iOS VoiceOver - where it supports automatic language switching.

nickcolley:
Thanks @rubenarakelyan this is great 👍

PR #205, opened 2016-12-13
Don't display Document Collection withdrawn documents

tvararu:
This brings government-frontend in line with Whitehall, which doesn't show withdrawn documents.

Trello:
https://trello.com/c/IFfWVlb6/533-filter-out-withdrawn-documents-in-document-collections-medium

Follow-up to:

- https://github.com/alphagov/govuk-content-schemas/pull/454
- https://github.com/alphagov/publishing-api/pull/627

Depends on:

- https://github.com/alphagov/govuk-content-schemas/pull/459

# Before

Below screenshots use the fictional `/government/collections/with-withdrawn-links-documents` example I created that has `Financial sanctions, Somalia` marked as withdrawn.

screen shot 2016-12-13 at 16 00 32

# After

screen shot 2016-12-13 at 16 00 38

### DIFF
--- a/test/presenters/document_collection_presenter_test.rb
+++ b/test/presenters/document_collection_presenter_test.rb
@@ -124,4 +124,21 @@ class DocumentCollectionPresenterTest
       assert_empty group_with_missing_documents
     end
   end
+
+  class GroupWithDocumentsWhenThereAreWithdrawnDocuments < TestCase
+    test "does not present withdrawn documents" do
+      presenter = presented_item(
+        "document_collection_with_withdrawn_links_documents"
+      )
+
+      presented_documents = presenter.groups.first["documents"]
+
+      expected_number_of_presented_documents = 2
+
+      assert_equal(
+        presented_documents.size,
+        expected_number_of_presented_documents
+      )
+    end
+  end
 end
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -16,9 +16,14 @@ class DocumentCollectionPresenter < ContentItemPresenter
   end
 
   def groups
-    content_item["details"]["collection_groups"].reject do |group|
+    groups = content_item["details"]["collection_groups"].reject { |group|
       group_document_links(group).empty?
-    end
+    }
+
+    groups.map { |group|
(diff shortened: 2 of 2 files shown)


test/presenters/document_collection_presenter_test.rb
```diff
@@ -124,4 +124,21 @@ class GroupWithDocumentsWhenThereIsNoLinksDocuments < TestCase
...
+
+  class GroupWithDocumentsWhenThereAreWithdrawnDocuments < TestCase
+    test "does not present withdrawn documents" do
+      presenter = presented_item(
+        "document_collection_with_withdrawn_links_documents"
+      )
+
+      groups_without_withdrawn_documents = presenter.groups.first["documents"]
+
+      number_of_documents_that_have_withdrawn_false = 2
```
tvararu: Not happy with this but because both `group_documents` and `documents_hash` are private I couldn't figure out a neater way to check that the withdrawn ones have been excluded. The only downside with doing this is if somebody goes and modifies the example by say, adding another document, 99% chance is they will break these tests. However, that happens rare enough that this should be fine.

test/presenters/document_collection_presenter_test.rb
```diff
@@ -124,4 +124,21 @@ class GroupWithDocumentsWhenThereIsNoLinksDocuments < TestCase
...
     end
   end
+
+  class GroupWithDocumentsWhenThereAreWithdrawnDocuments < TestCase
+    test "does not present withdrawn documents" do
+      presenter = presented_item(
+        "document_collection_with_withdrawn_links_documents"
+      )
+
+      groups_without_withdrawn_documents = presenter.groups.first["documents"]
```
gpeng: I don't think this name correctly describes what is returned here. This returns (or should return for the test to pass) a collection of non-withdrawn documents. Maybe `non_withdrawn_documents` might be a better name?

gpeng: Or possibly just `presented_documents`? The withdrawn bit is redundant as that is in the name of the test.

test/presenters/document_collection_presenter_test.rb
gpeng: And then `expected_number_of_presented_documents`

PR #206, opened 2016-12-15
Add jenkinsfile

steventux:
https://trello.com/c/OMAng5PU/783-move-government-frontend-to-new-ci-1
Make government-frontend use Jenkins 2

_Jenkins 2, just when you thought it was safe to go back to CI_

This seems to work...
https://ci.integration.publishing.service.gov.uk/job/government-frontend/


Jenkinsfile
```diff
@@ -0,0 +1,80 @@
...
+
+    stage("Precompile assets") {
+      govuk.precompileAssets()
+    }
+
+    stage("Run tests") {
+      govuk.runRakeTask("default")
+    }
+
+    stage("Push release tag") {
```
cbaines: For the gds-api-adapters, I tried to make this stage conditional, which I think works, and is slightly less surprising, as for branch builds, this just does nothing, but says it succeeded.

https://github.com/alphagov/gds-api-adapters/pull/638/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R121

PR #207, opened 2016-12-16
Add share buttons to consultations

fofr:
* Port social media icons from Whitehall, remove unused icons and PNGCrush
  * Crush all the PNGs to make @nickcolley happy
* Simplify markup and CSS for social icons
* Look and behaviour matches Whitehall
  * Except: Use link colours for share buttons
* Add javascript tracking of share buttons
* Fix bug with `offset-one-third` bug on thin devices

(This is the first JS added to government-frontend. I'd like to include `jasmine-rails` for JS tests but I had problems running versions 12+ against PhantomJS 1.9.7 on our VM. I will try and fix in a separate PR) Edit: https://github.com/alphagov/government-frontend/pull/208

https://trello.com/c/MaZ7t0zi/521-add-share-buttons-to-consultations-on-government-frontend

## Whitehall
screen shot 2016-12-16 at 16 29 08

## This PR
screen shot 2016-12-16 at 16 28 29

screen shot 2016-12-16 at 16 29 49

PR #208, opened 2016-12-19
Add jasmine-rails for javascript testing

fofr:
Adding javascript in #207, we need to be able to test it.
Using jasmine-rails to stay consistent with static, finder-frontend and manuals-frontend.

* Add jasmine-rails
* Add jasmine rails route for running tests in browser
* Include dummy spec to prove tests run
* Run javascript tests by default
* Use phantomjs provided by path rather than gem

Similar to https://github.com/alphagov/finder-frontend/pull/7


tvararu:
Nice one. 👏

tvararu:
This PR is a follow-up: https://github.com/alphagov/government-frontend/pull/213

PR #210, opened 2016-12-20
Create Shareable presenter

bevanloon:
Share links will be required by multiple formats.
Refactoring the code into a Shareable presenter will allow other formats
to include the functionality easily.


bevanloon:
Would be good to get some hints on how to test the Shareable module we've created - what we have feels a little clumsy and relies on knowledge of what the private methods in Shareable are doing.

fofr:
What formats will need this other than consultations?

bevanloon:
World location news articles

app/views/content_items/consultation.html.erb
```diff
@@ -231,18 +231,7 @@
...
-          target: "_blank",
-          class: "share-button-link js-share-facebook" do %>
-        <span class="share-button share-button-facebook"></span><span class="visually-hidden">Share on </span>Facebook<% end %>
-
-      <%= link_to @content_item.twitter_share_url,
-          target: "_blank",
-          class: "share-button-link js-share-twitter" do %>
-        <span class="share-button share-button-twitter"></span><span class="visually-hidden">Share on </span>Twitter<% end %>
-    </div>
+    <%= render 'shared/share_icons', share_urls: @content_item.share_urls %>
```
fofr: Could we call the partial `share_buttons`? The partial provides more than icons.

danielroseman:
Should this perhaps be a shared component?

bevanloon:
@danielroseman - we did think about that - we weren't sure if that was necessary at this point given only two formats are using it so far so we put this together in the meantime....

happy to make it a shared component if that's a better way to go?

danielroseman:
It was just a suggestion, happy for it to be here for now and extracted later if necessary.

app/views/content_items/consultation.html.erb
bevanloon: Changed in 1f7d1c5f998ebc95904523c6b53cbe49d58c1c32

fofr:
Nothing outside of this app is using the share button pattern yet, my preference is for it to stay here until it needs to be abstracted a level higher.

PR #213, opened 2016-12-21
Monkey patch JasmineRails only if defined

tvararu:
This is necessary to prevent it being referenced in routes.rb in production, where the module doesn't exist.

### DIFF
--- a/config/initializers/jasmine_rails.rb
+++ b/config/initializers/jasmine_rails.rb
@@ -1,9 +1,13 @@
-module JasmineRails
-  class ApplicationController < ActionController::Base
-    before_action :skip_slimmer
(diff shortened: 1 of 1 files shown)


boffbowsh:
Another option:

```ruby
module JasmineRailsSkipSlimmer
  extend ActiveSupport::Concern

  included do
    before_action :skip_slimmer
  end
  
  def skip_slimmer
    response.headers[Slimmer::Headers::SKIP_HEADER] = "true"
  end
end

JasmineRails::ApplicationController.include(JasmineRailsSkipSlimmer) if defined?(JasmineRails)
```

tvararu:
@boffbowsh I like that better, I'll update it. 👍

PR #215, opened 2016-12-21
Render the speeches format

fofr:
Tests will fail until the corresponding schema changes have been merged and deployed: https://github.com/alphagov/govuk-content-schemas/pull/468

Main story:
https://trello.com/c/TWsLTRzc/547-10-speech-migration-mvp-content-schema-examples-and-front-end-work

* Ports the speeches format from Whitehalls
* Removes all placeholder images, favouring an empty left column, as used by most formats
* Handles speakers provided in links hash, and speakers provided as string in details (when they don't have a GOV.UK profile)

## Speech sub types

| Speech sub-type | Schema document type | Displays as |
| --- | --- | --- |
| Transcript | speech | Speech |
| Draft text | speech | Speech |
| Speaking notes | speech | Speech |
| Written statement to Parliament | written_statement | Written statement to Parliament |
| Oral statement to Parliament | oral_statement | Oral statement to Parliament |
| Authored article | authored_article | Authored article |

## Speaker images

The images are pulled from the profile of the assigned speaker. For now the frontend expects this to come from the `image` property in the details. It may in future make more sense for it to come via an image property in the links hash of the speaker – as that's the dependency.

## Omissions

* The share features are pending https://github.com/alphagov/government-frontend/pull/210

## Screenshots

| Feature | Old | New |
| --- | --- | --- |
| Speaker without a profile |speaker-without-profile-old|speaker-without-profile-new|
| Authored article |authored-article-old|authored-article-new|
| Draft text | draft-text-old|draft-text-new|
| Tablet view |draft-text-tablet-old|draft-text-tablet-new|


tvararu:
Solid work.

The speaker image doesn't look like it's full width on tablet in the new version, is that intentional?

app/views/content_items/speech.html.erb
```diff
@@ -0,0 +1,41 @@
...
+<%= render 'shared/description', description: @content_item.description %>
+<%= render 'shared/sidebar_with_body', content_item: @content_item %>
+
+<%= render 'govuk_component/document_footer',
+    from: @content_item.from,
+    published: @content_item.published,
+    updated: @content_item.updated,
+    history: @content_item.history,
+    part_of: @content_item.part_of
+%>
```
tvararu: There's 3 different variations of ending tag `%>` positioning in this file but I don't mind.

fofr:
@tvararu The image width is intentional yes, when it goes full size it swamps the view a little. This is the current image behaviour on other formats.

app/views/content_items/speech.html.erb
```diff
@@ -0,0 +1,41 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, @content_item.title %>
+
```
bevanloon: Should we also include:
`<%= content_for :meta_description, @content_item.description %>`
?

app/views/content_items/speech.html.erb
bevanloon: We do seem to have several different styles of closing `%>` through the project. I'm not overly bothered, but it might be worth a tidy up at some point.

fofr: Will be fixed by #216

app/views/content_items/speech.html.erb
fofr: Added in 99f3f8f

PR #219, opened 2017-01-04
Handle nil `field_of_operation` for fatality notice

gpeng:
When a draft is initially created there is no field of operation link
available due to the known lack of draft links issue.

This is a quick fix to allow a notice to be published. Tests will be
added in a subsequent commit.

### DIFF
--- a/app/presenters/fatality_notice_presenter.rb
+++ b/app/presenters/fatality_notice_presenter.rb
@@ -2,8 +2,11 @@ class FatalityNoticePresenter < ContentItemPresenter
   include Metadata
 
   def field_of_operation
-    attributes = content_item["links"]["field_of_operation"].first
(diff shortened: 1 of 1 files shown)


app/presenters/fatality_notice_presenter.rb
```diff
@@ -2,19 +2,24 @@ class FatalityNoticePresenter < ContentItemPresenter
...
-      m[:other] = {
-        "Field of operation" => link_to(field_of_operation.title, field_of_operation.path)
-      }
+    if field_of_operation
+      super.tap do |m|
+        m[:other] = {
+          "Field of operation" => link_to(field_of_operation.title, field_of_operation.path)
+        }
+      end
     end
```
boffbowsh:
```ruby
else
  super
```

PR #221, opened 2017-01-04
Add noopener and noreferrer to new window links

### LINKED ISSUE #217: Share links need [rel=noopener]
We link externally to Facebook and Twitter on pages like https://www.gov.uk/government/consultations/allowing-learner-drivers-to-have-driving-lessons-on-motorways

We're using `[target=_blank]` which has a vulnerability (https://mathiasbynens.github.io/rel-noopener/) we can add `[rel=noopener]` in these cases.

fofr:
Fixes #217

We're using `[target=_blank]` which has a vulnerability
https://mathiasbynens.github.io/rel-noopener/

“To prevent pages from abusing window.opener, use rel=noopener. This
ensures window.opener is null in Chrome 49 and Opera 36.”

“For older browsers, you could use rel=noreferrer which also disables
the Referer HTTP header”

@nickcolley

### DIFF
--- a/app/views/shared/_share_buttons.html.erb
+++ b/app/views/shared/_share_buttons.html.erb
@@ -2,11 +2,13 @@
 <div class="share-buttons" data-module="track-share-button-clicks">
   <%= link_to share_urls[:facebook],
     target: "_blank",
+    rel: "noopener noreferrer",
     class: "share-button-link js-share-facebook" do %>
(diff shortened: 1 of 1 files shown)


nickcolley:
Tested this by clicking through and checking 'window.opener', looks good! 👍 

I'll also raise issues on the other frontend applications that have this problem.

PR #224, opened 2017-01-04
Don't show empty contents lists on HTML publications

fofr:
Not all HTML publications have a list of headings. When there are none, don’t include the empty list or its associated heading.

@gpeng Can you confirm how we publish headings for HTML publications without any headings? Is it passed through as an empty ordered list as in the example? This exists as a bug on live, and I've presumed the logic hasn't been fixed: https://www.gov.uk/government/publications/canada-united-kingdom-joint-declaration/canada-united-kingdom-joint-declaration

## Before
screen shot 2017-01-04 at 17 00 51

## After
screen shot 2017-01-04 at 17 00 16

### DIFF
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -78,8 +78,11 @@
     }
   }
 
+  .headings {
+    margin-bottom: $gutter;
+  }
+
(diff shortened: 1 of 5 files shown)


fofr:
We shouldn't send an empty list for headings when there are none, but for now this will fix it for users.

PR #228, opened 2017-01-06
World Location News Article front-end

bevanloon:
Tests will fail until the corresponding schema changes have been merged and deployed: 
https://github.com/alphagov/govuk-content-schemas/pull/476

[Trello card](https://trello.com/c/UExU1tXe)

Add presenter, view and related artefacts needed to support rendering World Location News Articles in government-frontend.

| Feature | Old | New |
|---|---|---|
| Simple example | whitehall_basic | government-frontent_basic |
| With attachments | whitehall_attachments | government-frontend_attachments |
| History mode | whitehall_history_mode | government-frontend_history_mode |
| Images in body | whitehall_images |government-frontend_images |
| Multiple translations | whitehall_multiple_translations | dev_multilpe_translations |
| Translated | whitehall_world_location_news_article_translated |government-frontend_world_location_news_article_translated |

### DIFF
--- a/app/views/content_items/world_location_news_article.html.erb
+++ b/app/views/content_items/world_location_news_article.html.erb
@@ -1,0 +1,46 @@
+<%= content_for :page_class, @content_item.format.dasherize %>
+<%= content_for :title, @content_item.page_title %>
+<%= content_for :meta_description, @content_item.description %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        context: t("content_item.format.#{@content_item.document_type}", count: 1),
+        title: @content_item.title,
+        average_title_length: "long" %>
+  </div>
+</div>
+
+<%= render 'shared/withdrawal_notice', content_item: @content_item %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/metadata',
+        from: @content_item.from,
+        part_of: @content_item.part_of,
+        first_published: @content_item.published,
+        last_updated: @content_item.updated,
+        see_updates_link: true
+    %>
+  </div>
+</div>
+<%= render 'shared/history_notice', content_item: @content_item %>
+
+<%= render 'shared/description', description: @content_item.description %>
+
+<%= render 'shared/sidebar_with_body', content_item: @content_item %>
+
+<div class="grid-row sidebar-with-body">
+  <div class="column-two-thirds offset-one-third">
+    <%= render 'shared/share_buttons', share_urls: @content_item.share_urls %>
+  </div>
+</div>
+
+<%= render 'govuk_component/document_footer',
+    from: @content_item.from,
+    updated: @content_item.updated,
+    history: @content_item.history,
+    published: @content_item.published,
+    part_of: @content_item.part_of,
+    direction: page_text_direction
+%>
--- a/app/views/content_items/world_location_news_article.html.erb
+++ b/app/views/content_items/world_location_news_article.html.erb
@@ -2,38 +2,11 @@
 <%= content_for :title, @content_item.page_title %>
 <%= content_for :meta_description, @content_item.description %>
 
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= render 'govuk_component/title',
-        context: t("content_item.format.#{@content_item.document_type}", count: 1),
-        title: @content_item.title,
-        average_title_length: "long" %>
-  </div>
-
-  <div class="column-third">
-    <%= render 'shared/available_languages',
-      translations: @content_item.available_translations
-    %>
-  </div>
-</div>
-
+<%= render 'shared/title_and_translations', content_item: @content_item %>
 <%= render 'shared/withdrawal_notice', content_item: @content_item %>
-
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= render 'govuk_component/metadata',
-        from: @content_item.from,
-        part_of: @content_item.part_of,
-        first_published: @content_item.published,
-        last_updated: @content_item.updated,
-        see_updates_link: true
-    %>
-  </div>
-</div>
+<%= render 'shared/metadata', content_item: @content_item %>
 <%= render 'shared/history_notice', content_item: @content_item %>
-
 <%= render 'shared/description', description: @content_item.description %>
-
 <%= render 'shared/sidebar_with_body', content_item: @content_item %>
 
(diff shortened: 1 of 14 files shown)


app/views/content_items/world_location_news_article.html.erb
```diff
@@ -0,0 +1,52 @@
...
+        title: @content_item.title,
+        average_title_length: "long" %>
+  </div>
+
+  <div class="column-third">
+    <%= render 'shared/available_languages',
+      translations: @content_item.available_translations
+    %>
+  </div>
+</div>
```
fofr: This block should use the `title_and_translations` partial along with the TitleAndContext module in the presenter.

app/views/content_items/world_location_news_article.html.erb
```diff
@@ -0,0 +1,52 @@
...
+
+<%= render 'shared/description', description: @content_item.description %>
+
+<%= render 'shared/sidebar_with_body', content_item: @content_item %>
+
+<div class="grid-row sidebar-with-body">
+  <div class="column-two-thirds offset-one-third">
+    <%= render 'shared/share_buttons', share_urls: @content_item.share_urls %>
+  </div>
+</div>
```
fofr: The grid-row wrappers for share buttons could be moved into the partial. All uses of the partial have the same wrapper.

app/views/content_items/world_location_news_article.html.erb
```diff
@@ -0,0 +1,52 @@
...
+
+<%= render 'govuk_component/document_footer',
+    from: @content_item.from,
+    updated: @content_item.updated,
+    history: @content_item.history,
+    published: @content_item.published,
+    part_of: @content_item.part_of,
+    direction: page_text_direction
+%>
```
fofr: This block should use the Metadata module and `@content_item.document_footer` to pass in parameters

config/locales/en.yml
```diff
@@ -225,6 +225,9 @@ en:
       transparency:
         one: Transparency data
         other: Transparency data
+      world_location_news_article:
```
fofr: `document_type` is always "news_article" in the schema examples. Is this string needed?
If we do need to keep it then we need to generate the translations for the other locales too, as this format can have translations.

bevanloon: so this is currently used when rendering the history notice (https://github.com/alphagov/government-frontend/blob/master/app/views/shared/_history_notice.html.erb#L3).

I notice that 'news_article' is already translated - so do we want to fall back to that for our translation? Otherwise this probably needs to be translated

app/views/content_items/world_location_news_article.html.erb
bevanloon: refactored in 880736c5faa27fdbd9542ebec04c62a37fd6968e

app/views/content_items/world_location_news_article.html.erb
```diff
@@ -0,0 +1,52 @@
...
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/metadata',
+        from: @content_item.from,
+        part_of: @content_item.part_of,
+        first_published: @content_item.published,
+        last_updated: @content_item.updated,
+        see_updates_link: true
+    %>
+  </div>
+</div>
```
bevanloon: refactored in 880736c5faa27fdbd9542ebec04c62a37fd6968e

app/views/content_items/world_location_news_article.html.erb
bevanloon: refactored in 880736c5faa27fdbd9542ebec04c62a37fd6968e

app/views/content_items/world_location_news_article.html.erb
bevanloon: 👍 60f19b0ac3f521050197ab972b4a520173904833 should do that

config/locales/en.yml
bevanloon: Had a quick chat in the core channel about this and one suggestion from @boffbowsh is that eventually WLNA might be consolidated into being a "news_article". This sounds reasonable but I'm not sure what's involved in doing this though. It possibly needs some investigation (created trello card for this: https://trello.com/c/kAJM8OfE/815-investigate-consolidating-wlna-and-news-article) and if we do want to do that then perhaps it warrants a story of its own.

For now, I've kept the entry in `en.yml` so that history notices show for WLNA. In addition, I've added empty entries to the other locales for world_location_news_article with commit 8ec75a8fb59f22e8ac9e36a596b18e5b1cc71393 so it's consistent with the rest of the application - there are other empty entries in the locales (thanks @gpeng) as well so following prior art in the meantime.

fofr: I think the history notice reads better as "This news article...". I don't think the user needs to understand the difference between news article and world location news article.

Rosa-Fox: Hi,

@fofr we can chat about this in the morning, but I have added a commit which translates 'world location news article' to 'news article' in English. 

Reason for this was that I was going to pull in the `document_type` as opposed to the `format` in `_history_notice` which would mean it would pull in `"document_type": "news_article"` for WLNAs. Sadly this messed up the text for a few other formats, for example in Detailed Guides the text “This guidance” changed to “This detailed guide".

Could have added a condition to the view that pulled in the `document_type` for `WLNA`s and the `format` for the other doc_types but thought it might be better to keep it consistent and keep the blank lines in the translation files until we get a translation at some point.

bevanloon: @Rosa-Fox - what did content think about simply leaving out the `format` from the history notice? That would mean everything would simply say "This was published under the..."?

Instead of a condition in the view, we could also use Rails' [`default`](http://guides.rubyonrails.org/i18n.html#defaults) option in the `I18n.t` - so we could have something like

```ruby
I18n.t(
       "content_item.format.#{content_item.format}",
       default: ["content_item.format.#{content_item.document_type}".to_sym],
       count: 1
      ).downcase
```

Then if there were no translations for "World location news article" it would fall back to using `document_type` which would be "news article". My only real concern here is that it isn't really a default, it's a fallback so the naming is annoying.

config/locales/en.yml
```diff
@@ -226,8 +226,8 @@ en:
         one: Transparency data
         other: Transparency data
       world_location_news_article:
```
fofr: We can remove this completely now can't we? And from all the translation files.
I believe the history partial was the only thing using it.

Rosa-Fox:
@fofr have removed the `world location news article` translations.

PR #231, opened 2017-01-16
Add config and instructions for comparing examples with wraith

fofr:
Not all formats are live, and live content changes. Sometimes it’s preferable to compare examples on a branch against master.

* Create wraith config for testing diffs of examples
* Include instructions on how to generate diffs in readme

Example output:
https://fofr.github.io/government-frontend-wraith-tests/shots/gallery.html

Examples generated using this gist:
https://gist.github.com/fofr/765604f8ae27164736ac4a8c8e96ddeb

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,27 +68,46 @@ Or to specify the location explicitly:
 
 `GOVUK_CONTENT_SCHEMAS_PATH=/some/dir/govuk-content-schemas bundle exec rake`
 
-#### Visual regression tests
+### Visual regression tests
 
 Use [Wraith](http://bbc-news.github.io/wraith/) ("A responsive screenshot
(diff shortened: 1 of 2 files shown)


nickcolley:
Nit: Should `config-examples` be `content-examples`?

fofr:
@nickcolley All other configs begin with `config-`

nickcolley:
Ah, sorry, stupid question then.

fofr:
The example linked to here
https://fofr.github.io/government-frontend-wraith-tests/shots/gallery.html

Gives a couple of false positives. Running it a few times the false positives remain, but appear on different formats. I think it's something to do with the way phantom is handling broken images and how it spaces them, as there's no code differences between samples.

It's also odd that the webfont isn't rendering, but I recall this being an issue with all wraith tests.

PR #234, opened 2017-01-20
Remove IE specific stylesheets

fofr:
There are no IE specific styles in this application.

The govuk_frontend_toolkit defines some IE helpers here: https://github.com/alphagov/govuk_frontend_toolkit/blob/2011e77436fccd31f32a0a975a2bf4cc3b961495/stylesheets/_conditionals.scss#L67-L81

These rely on the IE6-8 specific stylesheets. We've ported almost all the formats now and we haven't had to override styles for these older browsers.


fofr:
These stylesheets are needed to correctly interpret media queries.

PR #237, opened 2017-01-24
Navigation AB Test

alecgibson:
Add functionality to check for A/B Testing, and display content according to which "bucket" the user has been added to.

This WIP PR adds a helper method to the controller to check for A/B bucket (with associated tests).

As a proof-of-concept, it also updates two of the `content_item` views to show either an A or B view. Currently the only difference between the two is the display of "normal" breadcrumbs (view A) or "taxon" breadcrumbs (view B).

If we continue in this style, we should add A/B views for _all_ Content Items. Is this the best approach? If so, should we be grouping them in subfolders under `content_items` to reduce clutter? Is there any scope for commonising, or is that a waste of time, given that - if testing is successful - we'd just want to delete all the "A" views anyway in the long run.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,23 +1,18 @@
 require 'gds_api/content_store'
 
 class ContentItemsController < ApplicationController
-  include ABTestable
-
   rescue_from GdsApi::HTTPForbidden, with: :error_403
   rescue_from GdsApi::HTTPNotFound, with: :error_notfound
 
+  attr_accessor :content_item
+
   def show
     load_content_item
+    set_up_education_navigation_ab_testing
     set_expiry
-    set_page_variant
     with_locale do
       render content_item_template
     end
-  end
-
-  # Visible for testing
-  def should_present_new_navigation_view?
-    (education_navigation_ab_testing_group == "B") && new_navigation_enabled? && content_is_tagged_to_a_taxon?
   end
 
 private
@@ -45,6 +40,21 @@ private
     expires_in(max_age, public: !cache_private)
   end
 
+  def set_up_education_navigation_ab_testing
+    @education_navigation_ab_test = EducationNavigationAbTestRequest.new(request)
+    return unless @education_navigation_ab_test.content_schema_has_new_navigation?(@content_item.content_item)
+
+    @education_navigation_ab_test.set_response_vary_header response
+
+    # Setting a variant on a request is a type of Rails Dark Magic that will use a convention to automagically load
+    # an alternative partial/view/layout.
+    # For example, if I set a variant of :new_navigation and we render a partial called _breadcrumbs.html.erb then Rails
+    # will attempt to load _breadcrumbs.html+new_navigation.erb instead. If such a file does not exist, then it falls
+    # back to _breadcrumbs.html.erb.
+    # See: http://edgeguides.rubyonrails.org/4_1_release_notes.html#action-pack-variants
+    request.variant = :new_navigation if @education_navigation_ab_test.should_present_new_navigation_view?(@content_item.content_item)
+  end
+
   def with_locale
     I18n.with_locale(@content_item.locale || I18n.default_locale) { yield }
   end
@@ -63,23 +73,5 @@ private
 
   def error_notfound
     render plain: 'Not found', status: :not_found
-  end
-
-  # Setting a variant on a request is a type of Rails Dark Magic that will use a convention to automagically load
-  # an alternative partial/view/layout.
-  # For example, if I set a variant of :new_navigation and we render a partial called _breadcrumbs.html.erb then Rails
-  # will attempt to load _breadcrumbs.html+new_navigation.erb instead. If such a file does not exist, then it falls
-  # back to _breadcrumbs.html.erb.
-  # See: http://edgeguides.rubyonrails.org/4_1_release_notes.html#action-pack-variants
-  def set_page_variant
-    request.variant = :new_navigation if should_present_new_navigation_view?
-  end
-
-  def new_navigation_enabled?
-    ENV['ENABLE_NEW_NAVIGATION'] == 'yes'
-  end
-
-  def content_is_tagged_to_a_taxon?
-    @content_item.taxons.any?
   end
 end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,15 +1,23 @@
 require 'gds_api/content_store'
 
 class ContentItemsController < ApplicationController
+  include ABTestable
+
   rescue_from GdsApi::HTTPForbidden, with: :error_403
   rescue_from GdsApi::HTTPNotFound, with: :error_notfound
 
   def show
     load_content_item
     set_expiry
+    set_page_variant
     with_locale do
       render content_item_template
     end
+  end
+
+  # Visible for testing
+  def should_present_new_navigation_view?
+    (education_navigation_ab_testing_group == "B") && new_navigation_enabled? && content_is_tagged_to_a_taxon?
   end
 
 private
@@ -55,5 +63,23 @@ private
 
   def error_notfound
     render plain: 'Not found', status: :not_found
+  end
+
+  # Setting a variant on a request is a type of Rails Dark Magic that will use a convention to automagically load
+  # an alternative partial/view/layout.
+  # For example, if I set a variant of :new_navigation and we render a partial called _breadcrumbs.html.erb then Rails
+  # will attempt to load _breadcrumbs.html+new_navigation.erb instead. If such a file does not exist, then it falls
+  # back to _breadcrumbs.html.erb.
+  # See: http://edgeguides.rubyonrails.org/4_1_release_notes.html#action-pack-variants
+  def set_page_variant
+    request.variant = :new_navigation if should_present_new_navigation_view?
+  end
+
+  def new_navigation_enabled?
+    ENV['ENABLE_NEW_NAVIGATION'] == 'yes'
+  end
+
+  def content_is_tagged_to_a_taxon?
+    @content_item.taxons.any?
   end
 end
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -3,8 +3,6 @@ class ContentItemPresenter
 
   attr_reader :content_item, :title, :description, :format, :locale, :phase, :document_type
 
-  delegate :breadcrumbs, to: :@nav_helper
-
   def initialize(content_item)
     @content_item = content_item
     @title = content_item["title"]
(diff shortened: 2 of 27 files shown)


nickcolley:
If we're only testing the breadcrumbs wouldnt only want to A/B that component itself?

alecgibson:
@nickcolley we'll eventually be shuffling actual HTML around. For example, in Detailed Guide, we'll need to make space on the right-hand-side for a Related Links box.

carvil:
We will also be removing "Part of" as well, so this will be a bit more comprehensive than just breadcrumbs.

app/controllers/content_items_controller.rb
```diff
@@ -7,11 +7,18 @@ class ContentItemsController < ApplicationController
...
     load_content_item
     set_expiry
+    @ab_testing_group = ab_testing_group
     with_locale do
       render content_item_template
     end
   end
 
+  def should_present_b_view?
+    (ab_testing_group == "B") && new_navigation_enabled? && content_is_tagged_to_a_taxon?
```
carvil: Do we have the new chrome extension ready so we can add ourselves to the B group in the A/B test on integration?

alecgibson:
Yes, this is just a small change to start with to keep the PR manageable and get early feedback.

app/controllers/content_items_controller.rb
```diff
@@ -7,11 +7,18 @@ class ContentItemsController < ApplicationController
...
+  end
+
+  helper_method :should_present_b_view?
```
carvil: What do you think of changing this to `should_present_new_content_view`? It could be hard to understand what the B version is without more context.

app/controllers/content_items_controller.rb
alecgibson: Not sure. @suzannehamilton ? At the moment you can manually manipulate your own headers with a Chrome extension.

app/views/content_items/_case_study_a.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= content_for :title, "#{@content_item.page_title} - #{t("content_item.format.#{@content_item.document_type}", count: 1)}" %>
```
carvil: Like the comment above, I think we could try to rename these partials to something other than a/b, so it's clearer what they render. What do you think?

alecgibson: What do you propose? Is "New"/"Old" any better than "A"/"B"?

carvil: What do you think of:

```
_case_study.html.erb
```

for the A version and:

```
_case_study_for_new_navigation.html.erb
```

for the B version?

app/controllers/content_items_controller.rb
tijmenb: That's currently on master, we'll release soon.

app/controllers/content_items_controller.rb
```diff
@@ -56,4 +63,16 @@ def error_403(exception)
   def error_notfound
     render plain: 'Not found', status: :not_found
   end
+
+  def ab_testing_group
```
fofr: Could the logic for generic A/B testing be collected in a separate file?
Can we have a further separate place for the logic specific to each test?

app/controllers/content_items_controller.rb
```diff
@@ -56,4 +63,16 @@ def error_403(exception)
...
+    request.headers["HTTP_GOVUK_ABTEST_EDUCATIONNAVIGATION"] == "B" ? "B" : "A"
+  end
+
+  def new_navigation_enabled?
+    ENV['ENABLE_NEW_NAVIGATION'] == 'yes'
+  end
+
+  def content_is_tagged_to_a_taxon?
```
fofr: What's a taxon?

app/presenters/content_item_presenter.rb
```diff
@@ -31,6 +29,22 @@ def content_id
     content_item.parsed_content["content_id"]
   end
 
+  def taxons
```
fofr: Breadcrumbs aren't included on every format.
We use a mixin pattern to include features when they're needed.
Should these methods be in a mixin and included on just the formats that need them?

This will prevent the single content_item_presenter from getting too large.

app/controllers/content_items_controller.rb
alecgibson: Could do, although we'd still need a method here to check if the content is tagged to a taxon. Where's the best place for such a method to live?

app/controllers/content_items_controller.rb
alecgibson: It's a new document type that defines the new navigation structure. A taxonomy is a tree of taxons. Content is tagged to these taxon nodes essentially as leaf nodes.

app/views/content_items/_detailed_guide_b.html.erb
```diff
@@ -0,0 +1,54 @@
+<%= content_for :title, @content_item.page_title %>
```
fofr: It's not clear what the differences between `a.html.erb` and `b.html.erb` are. Rather than two templates, could the variations be provided by picking different partials? This reduces duplication and makes the template variations clearer.

app/presenters/content_item_presenter.rb
alecgibson: Sure sounds good to me.

app/views/content_items/_detailed_guide_b.html.erb
fofr: eg
```
if a:
<%= render "shared/taxon_breadcrumbs" %>
else:
<%= render "shared/breadcrumbs" %>
```

alecgibson: I've not dived into it yet, but I think the HTML is going to be moving around. For example, on Detailed Guides, we need to shift the content to the left two thirds to make room for a Related Items box on the right. This would include shifting the contents and some metadata out of the way (presumably to the top of the document?). I'm not sure that just picking different partials will really cut it in this case.

fofr: Or the logic could be in the partial itself, if it affects all breadcrumbs on all formats with breadcrumbs.

alecgibson: The reason for not putting the switch in the partial itself is that I didn't want two levels of switching - it's clear coming to Detailed Guide that there's a big switch on the template. It'd be less clear if you came to a single Detailed Guide that the partials have switches buried in them, making it harder still to track down differences between views.

app/controllers/content_items_controller.rb
fofr: It could be included as a module.

app/views/content_items/_detailed_guide_b.html.erb
fofr: To someone editing a detailed guide or case study, it is not clear that they need to make the change in two places.

Why does it need to be clear that there is a big switch?

alecgibson: I think this way you're making it explicit by saying "This is what the _old_ content looks like", and "This is what the _new_ content looks like."

fofr: > It's not clear what the differences between a.html.erb and b.html.erb are.
I think this needs to be solved, whichever approach is taken.

app/presenters/content_item_presenter.rb
tijmenb: Related: https://github.com/alphagov/government-frontend/pull/238

test/controllers/content_items_controller_test.rb
```diff
@@ -75,6 +75,43 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_response :forbidden
   end
 
+  test "honours AB Testing cookie" do
+    assert_equal "yes", ENV['ENABLE_NEW_NAVIGATION']
```
fofr: Should the testing of the AB feature be more generic, ie not be tied to this specific test?

alecgibson: What would you propose?

app/views/content_items/_detailed_guide_b.html.erb
alecgibson: @fofr I'm not sure I understand what you're proposing. Surely the only way to be able to have arbitrarily different HTML views is to have two different templates (or variants)?

test/controllers/content_items_controller_test.rb
fofr: Mock the environment variable and mock the headers?

Then add tests specific to _this AB test_.

alecgibson: I'm not sure what good generic tests would be? The checks for AB testing aren't (and don't need to be?) generic.

nickcolley:
Variants looking good! 👍  Tiny typo case_stude -> case_study

test/controllers/content_items_controller_test.rb
fofr: Maybe just your test that needs re-phrasing then, "honours AB Testing cookie" sounds generic.

alecgibson:
@carvil @nickcolley @fofr @tijmenb This is now ready for an actual review

fofr:
Can the commit history on this PR be cleaned up?

alecgibson:
@carvil any more changes needed on this?

carvil:
@alecgibson could you please deploy this branch to integration so we can see how it looks?

app/views/shared/_footer.html+new_navigation.erb
```diff
@@ -0,0 +1 @@
+<%= render 'govuk_component/document_footer', @content_item.document_footer_for_new_navigation %>
```
fofr: It looks like the only difference between templates are the parameters being passed to a component. Rather than switching templates, would it make more sense to use the presenter to alter the parameters passed to the component? This will prevent templates from getting out of sync.

Here is an example of a format overriding the default method:
https://github.com/alphagov/government-frontend/blob/f3700d94a92c318bae5032174d1e5ac2b2f4361a/app/presenters/speech_presenter.rb#L28-L41

Default defined here:
https://github.com/alphagov/government-frontend/blob/f3700d94a92c318bae5032174d1e5ac2b2f4361a/app/presenters/metadata.rb

alecgibson: Okay, so I've had a look at this, but I think if we wind up doing this, we have to start passing flags from the controllers to the presenters, and you start scattering your A/B logic between controllers, views/partials _and_ presenters (whereas right now all A/B logic exists in just controllers and views/partials). I think this third place for logic obscures the code a bit, which outweighs the risk of things drifting, especially since this code should be merged into actual prod code (or not) over the next couple of months anyway.

fofr: We've been gradually moving as much logic into presenters as we can, so that templates are simplified for consolidation.

alecgibson:
@carvil this is finally ready for re-review (cc @mobaig )

mobaig:
LG 👍

alecgibson:
@carvil I've now bumped the versions of both `govuk_navigation_helpers` and `govuk_ab_testing`

carvil:
@alecgibson could you please `git cherry-pick` b7f62d3 to this PR? That way we can ship everything in one go.

alecgibson:
@carvil done

tijmenb:
🎈

nickcolley:
🎉

alecgibson:
Belatedly adding this Trello card for future context: https://trello.com/c/yLGE64pa/1013-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-government-frontend

PR #238, opened 2017-01-25
Show breadcrumbs consistently

tijmenb:
Currently we don't show the breadcrumbs if the format traditionally doesn't have them.

On other pages on GOV.UK we've introduced the breadcrumbs, whether or not they're set. For example, https://www.gov.uk/aaib-reports just has a "Home" breadcrumb, because
https://www.gov.uk/aaib-reports/aaib-investigation-to-airbus-a319-111-g-ezfj has a breadcrumb to that page.

This adds consistent breadcrumbs to all pages. This means that we'll show empty "Home" breadcrumbs for fatality notices, statistics announcements, statistical data sets, "take part" pages and groups. Note that it's super easy to set the breadcrumb for these pages, by adding a `parent` link to the presenter in Whitehall. 

This should make https://github.com/alphagov/government-frontend/pull/237 a little easier.

@nickcolley @fofr thoughts?

## Examples

| before | after |
|--|--|
| screen shot 2017-01-25 at 12 41 11 | screen shot 2017-01-25 at 12 41 23 |
| screen shot 2017-01-25 at 12 41 27 | screen shot 2017-01-25 at 12 41 40 |

### DIFF
--- a/app/views/content_items/case_study.html.erb
+++ b/app/views/content_items/case_study.html.erb
@@ -1,6 +1,5 @@
 <%= content_for :title, "#{@content_item.page_title} - #{t("content_item.format.#{@content_item.document_type}", count: 1)}" %>
 
-<%= render "shared/breadcrumbs" %>
 <%= render 'shared/title_and_translations', content_item: @content_item %>
 <%= render 'shared/withdrawal_notice', content_item: @content_item %>
 <%= render 'shared/metadata', content_item: @content_item %>
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -1,7 +1,5 @@
 <%= content_for :title, @content_item.page_title %>
 
-<%= render "shared/breadcrumbs" %>
-
 <div class="grid-row">
   <div class="column-two-thirds">
     <%= render 'govuk_component/title', @content_item.title_and_context %>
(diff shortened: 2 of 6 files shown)


fofr:
:+1: to the simplification

However, is this useful to users? It looks like we're pushing the content down further and adding a duplicate link.

nickcolley:
I think @alextea mentioned they had user research suggesting that people do indeed use the home breadcrumb on it's own.

alextea:
@nickcolley we've not tested showing pages with just "Home", so not sure about that, but not all users understand that the logo is a link so I think adding this is a step in the right direction.

bevanloon:
Just to weigh in - I'm currently fixing a bug in whitehall's statistical data set presenter to include `parent` in the `links` hash. So I think statistical data sets, at least, will need breadcrumbs anyway.

In general, I'll echo @fofr on the 👍 for simplification, but also raise a question about the user need for a single breadcrumb that only goes home.

carolinegreen:
@nickcolley In a lot of the user research users would go to click the "Home" breadcrumb link first before anything else, but like Alex said we haven't tested pages with home on it's own.

tijmenb:
We've also done a quick research in analytics just now:

If we take https://www.gov.uk/aaib-reports as an example of a page with only a "Home" breadcrumb. It has 7,691 unique pageviews over the last month,  of which 160 people clicked "Home". That's quite a lot.

Overall in the last 30 days 741,789 people clicked the "Home" in the breadcrumb, mostly from browse & done pages.

screen shot 2017-01-25 at 16 04 19

fofr:
We should validate that "Home" takes people to the place they expect.

PR #243, opened 2017-02-01
Guard against missing change history

andrewgarner:
Fixes https://errbit.publishing.service.gov.uk/apps/546493f90da1150e4e001cfc/problems/58905b8465786333bfd80000


fofr:
Want to tweak a test so we avoid regressions?

PR #245, opened 2017-02-06
Fix nil organisations for HtmlPublicationPresenter

mgrassotti:
Trello: https://trello.com/c/dMbLwfHJ/585-fix-bug-with-html-pub-having-no-organisation-2

Avoid these lines raise an exception if `content_item["links”][“organisations”]` is `nil`


fofr:
Can you include a test so we avoid regressions?

mgrassotti:
@fofr thanks but we are fixing it with a data migration instead. https://github.com/alphagov/whitehall/pull/3016

PR #248, opened 2017-02-09
Add translations to publications

gpeng:
Publications can be translated but we have neglected to add the necessary code to render the locale choice UI in the publication view.

This PR remedies that along with a small refactor to avoid duplication.

There could be an issue if National Statistics type publications are translated as this might break the grid but a) there are none in the Whitehall database, b) it seems unlikely that that sub-type would be translated and c) the quick test one that I knocked up appears to be OK. So... I think we can proceed without worrying about that for now to unblock deploying [Whitehall](https://github.com/alphagov/whitehall/pull/2986)

Before:
screen shot 2017-02-09 at 18 22 29

After:
screen shot 2017-02-09 at 18 22 48

Test with national statistics logo and translations:
screen shot 2017-02-09 at 18 23 05

### DIFF
--- a/app/views/content_items/publication.html.erb
+++ b/app/views/content_items/publication.html.erb
@@ -7,6 +7,7 @@
         title: @content_item.title,
         average_title_length: "long" %>
   </div>
+  <%= render 'shared/translations' %>
   <% if @content_item.national_statistics? %>
     <%= render 'shared/national_statistics_logo' %>
   <% end %>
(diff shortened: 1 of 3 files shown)


fofr:
Can we put the translation links above the logo?

gpeng:
Logo moved in d80c7cc (although subjectively I prefer it the other way)

screen shot 2017-02-10 at 10 02 18

gpeng:
Discussed offline. As there aren't any instances of translated national statistics we're going to deploy this to unblock Whitehall and fix the layout in a later PR [Trello](https://trello.com/c/9k9udb9g/843-fix-translated-national-statistics-layout).

PR #251, opened 2017-02-17
Update jasmine-rails to 0.14.1

nickcolley:
Removes deprecation warnings when running the jasmine tests


Gemfile.lock
```diff
@@ -272,5 +272,8 @@ DEPENDENCIES
   unicorn (= 4.8)
   webmock (~> 1.18.0)
 
+RUBY VERSION
```
fofr: Should this be here?

nickcolley: No, we should get that from the .ruby-version file, I'll remove it.

PR #255, opened 2017-02-22
Render travel advice format

fofr:
Part of https://trello.com/c/L5Nd0AdH/137-3-render-travel-advice-in-government-frontend-l

__Ports travel advice from multipage-frontend__

Notable differences:
* The two-line "Summary/Current travel advice" navigation has been switched to "Current travel advice"
* The navigation links use grids for columns, meaning they are aligned slightly differently, but more consistently with our other content
* The map and download link are now in a `<figure>` tag
* Parts have unique page titles (before the `<title>` tag was the same)
* Logic for splitting navigation list includes summary when counting, making the two lists more equal
* Adds support for alt tags on map images if they every come through in the content item

Notable omissions:
* No atom feeds yet
* No print views yet
* No next/prev/atom meta tags in the header yet
* No publishing request id meta tags, they aren't being used

Notable features:
* Adds a custom route, controller and presenter initialiser for travel advice. This is to get around the problem where no content item exists for child pages of advice:
  * eg Content for https://www.gov.uk/foreign-travel-advice/albania/travel-advice-help-and-support can't be reached at https://www.gov.uk/api/content/foreign-travel-advice/albania/travel-advice-help-and-support
  * Being tracked here: https://trello.com/c/w8HN3M4A/
  * Passes the currently showing part as an argument to presenter so presenter can know which part to display
* Non-existent parts continue to redirect back to the guide's base path as per multipage-frontend
* Feed link and atom link methods in presenter may be replaced with Rails routes when atom/print get added
* Includes "alert status" feature which isn't being actively used (but still _could_ be)

## Summary

|Old|New|
|--|--|
|summary-old|summary-new-2|

## Summary small viewport

|Old|New|
|--|--|
|summary-mobile-old|summary-mobile-new-2|

## Travel advice part

|Old|New|
|--|--|
|part_old|part-new-2|


nickcolley:
Nice work!

I'll do a full review shortly but just something I've noticed:

I think we need to address the vertical rhythm between components on our new version, previous the whitespace really helped break up the page and I think we're losing that somewhat at the moment.

fofr:
> I think we need to address the vertical rhythm between components on our new version, previous the whitespace really helped break up the page and I think we're losing that somewhat at the moment.

Can you be specific about which margins?

Do you mean around the part title, metadata and the map?

nickcolley:
From a discussion, we think we should make the navigation consistent with the second heading, so make Summary equal to Current Travel Advice etc.

fofr:
Margins and first part title updated.

fofr:
Screenshots updated.

app/views/content_items/travel_advice.html.erb
```diff
@@ -0,0 +1,69 @@
...
+  content_for :title, @content_item.page_title
+  content_for :simple_header, true
+%>
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title', @content_item.title_and_context %>
+    <aside class="part-navigation-container" role="complementary">
+      <nav role="navigation" class="grid-row part-navigation" aria-label="Travel advice pages">
+        <% @content_item.parts_navigation.each_with_index do |part_group, i| %>
+          <ol class="column-half" <% if i > 0 %>start="<%= @content_item.parts_navigation_second_list_start %>"<% end %>>
```
nickcolley: Can we break this up across a few lines so it's easy to read?

fofr: Updated and rebased.

PR #256, opened 2017-02-23
Education navigation ab testing

alecgibson:
This extends the styling changes made in https://github.com/alphagov/government-frontend/pull/237 to also encompass the following content item schemas:

- `document_collection`
- `publication`

The changes are to:

- Add a `meta` tag to these documents to indicate that they are in an A/B test (when their content is tagged to a taxon)
- Set the `Vary` header on responses from these documents
- Move the content for these documents into the left two thirds of the page (when the user is in the "B" group)

### Trello

https://trello.com/c/jd56fcgd/338-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-government-frontend-part-2

### DIFF
--- a/app/views/shared/_national_statistics_logo.html+new_navigation.erb
+++ b/app/views/shared/_national_statistics_logo.html+new_navigation.erb
@@ -1,0 +1,3 @@
+<div class="national-statistics-logo">
+  <%= image_tag 'national-statistics.png', alt: 'National Statistics' %>
+</div>
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -137,15 +137,15 @@ class ContentItemsControllerTest < ActionController::TestCase
 
     content_store_has_item(content_item['base_path'], content_item)
 
-    with_variant EducationNavigation: "A" do
-      get :show, params: { path: path_for(content_item) }
-      assert_equal [], @request.variant
-    end
+    setup_ab_variant('EducationNavigation', 'A')
+    get :show, params: { path: path_for(content_item) }
+    assert_equal [], @request.variant
+    assert_unaffected_by_ab_test
 
-    with_variant EducationNavigation: "B" do
-      get :show, params: { path: path_for(content_item) }
-      assert_equal [], @request.variant
-    end
+    setup_ab_variant('EducationNavigation', 'B')
+    get :show, params: { path: path_for(content_item) }
+    assert_equal [], @request.variant
+    assert_unaffected_by_ab_test
   end
 
   test "Case Studies are not included in the AB Test" do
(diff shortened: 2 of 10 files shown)


test/controllers/content_items_controller_test.rb
```diff
@@ -137,15 +137,13 @@ class ContentItemsControllerTest < ActionController::TestCase
 
     content_store_has_item(content_item['base_path'], content_item)
 
-    with_variant EducationNavigation: "A" do
```
suzannehamilton: What's the issue with using `with_variant` here? Is it something we should change in the A/B gem?

test/controllers/content_items_controller_test.rb
```diff
@@ -137,15 +137,13 @@ class ContentItemsControllerTest < ActionController::TestCase
...
-      get :show, params: { path: path_for(content_item) }
-      assert_equal [], @request.variant
-    end
+    setup_ab_variant('educationnavigation', 'A')
```
suzannehamilton: This should be `EducationNavigation` to match the casing elsewhere.

test/controllers/content_items_controller_test.rb
```diff
@@ -137,15 +137,13 @@ class ContentItemsControllerTest < ActionController::TestCase
...
+    get :show, params: { path: path_for(content_item) }
+    assert_equal [], @request.variant
 
-    with_variant EducationNavigation: "B" do
-      get :show, params: { path: path_for(content_item) }
-      assert_equal [], @request.variant
-    end
+    setup_ab_variant('educationnavigation', 'B')
```
suzannehamilton: Ditto.

test/controllers/content_items_controller_test.rb
alecgibson: No, it's just that I incorrectly added the meta tag and response headers before, where they shouldn't be applied if we've not got content tagged to a taxon.

alecgibson:
@suzannehamilton the templates are probably worth reviewing in case I missed anything. Sorry the diffing isn't easier.

test/controllers/content_items_controller_test.rb
suzannehamilton: In that case, you could include the `assert_response_not_modified_for_ab_test` assertion, which will check that the meta tag and response headers haven't been added.

app/views/shared/_national_statistics_logo.html+new_navigation.erb
```diff
@@ -0,0 +1,3 @@
+<div class="national-statistics-logo">
```
suzannehamilton: Why does this logo need a `new_navigation` file? Is it different in the A and B versions?

alecgibson: Yes - it was in a `one-third-column` div

PR #257, opened 2017-02-24
Add atom feeds to travel advice

fofr:
https://trello.com/c/mNRwOtHL/139-2-render-travel-advice-atom-feeds-in-government-frontend-m

Render atom feeds for travel advice countries. Output of atom feed is identical to multipage-frontend’s (and live). Template and tests ported from multipage-frontend.

__When reviewing please double-check the routing logic and tests__

Add support for rendering atom routes:
* Update routing to allow atom format
* Use a regex to constrain matches on locale and format
* Test that all supported locales work with regex
* Append the locale when matched to the content_item_path
* Return 406 when unacceptable format requested

Paired with @andrewgarner

Example feed:
Matches: https://www.gov.uk/foreign-travel-advice/thailand.atom

```
<?xml version="1.0" encoding="UTF-8"?>
<feed xml:lang="en-US" xmlns="http://www.w3.org/2005/Atom">
  <id>https://www.gov.uk/foreign-travel-advice/thailand</id>
  <link rel="alternate" type="text/html" href="https://www.gov.uk/foreign-travel-advice/thailand"/>
  <link rel="self" type="application/atom+xml" href="http://government-frontend.dev.gov.uk/foreign-travel-advice/thailand.atom"/>
  <title>Travel Advice Summary</title>
  <updated>2017-02-08T10:23:34+00:00</updated>
  <author>
    <name>GOV.UK</name>
  </author>
  <entry>
    <id>https://www.gov.uk/foreign-travel-advice/thailand#2017-02-08T10:23:34+00:00</id>
    <updated>2017-02-08T10:23:34+00:00</updated>
    <link rel="alternate" type="text/html" href="https://www.gov.uk/foreign-travel-advice/thailand"/>
    <title>Thailand travel advice</title>
    <link rel="self" type="application/atom+xml" href="https://www.gov.uk/foreign-travel-advice/thailand.atom"/>
    <summary type="xhtml">
      <div xmlns="http://www.w3.org/1999/xhtml">
<p>Latest update: Summary – updated guidance for visitors following the passing of His Majesty King Bhumibol Adulyadej
</p>      </div>
    </summary>
  </entry>
</feed>
```

### DIFF
--- a/app/presenters/travel_advice_presenter.rb
+++ b/app/presenters/travel_advice_presenter.rb
@@ -37,6 +37,10 @@ class TravelAdvicePresenter < ContentItemPresenter
     }
   end
 
+  def country_name
+    content_item["details"]["country"]["name"]
+  end
+
   def related_items
     items = ordered_related_items.map do |link|
       {
@@ -139,10 +143,6 @@ private
     else
       parts.find { |part| part["slug"] == @part_slug }
     end
-  end
-
-  def country_name
-    content_item["details"]["country"]["name"]
   end
 
   def parts
--- a/app/presenters/travel_advice_presenter.rb
+++ b/app/presenters/travel_advice_presenter.rb
@@ -112,6 +112,18 @@ class TravelAdvicePresenter < ContentItemPresenter
     alert_statuses.join('').html_safe
   end
 
+  def atom_change_description
+    simple_format(HTMLEntities.new.encode(change_description, :basic, :decimal))
+  end
+
+  def atom_public_updated_at
+    DateTime.parse(content_item["public_updated_at"])
+  end
+
+  def web_url
+    Plek.current.website_root + content_item["base_path"]
+  end
+
 private
 
   def summary_part
@@ -163,6 +175,10 @@ private
     content_item["links"]["ordered_related_items"] || []
   end
 
+  def change_description
+    content_item['details']['change_description']
+  end
+
   # FIXME: Update publishing app UI and remove from content
   # Change description is used as "Latest update" but isn't labelled that way
   # in the publisher. The frontend didn't add this label before.
@@ -171,7 +187,6 @@ private
   # has a latest update label, so we can strip this out.
   # Avoids: "Latest update: Latest update - …"
   def latest_update
-    change_description = content_item['details']['change_description']
     change_description.sub(/^Latest update:?\s-?\s?/i, '').tap do |latest|
       latest[0] = latest[0].capitalize
     end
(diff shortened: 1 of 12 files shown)


config/routes.rb
```diff
@@ -2,10 +2,14 @@
...
-  end
+  get 'healthcheck', to: proc { [200, {}, ['']] }
+
+  # FIXME: Update when https://trello.com/c/w8HN3M4A/ is ready
+  get 'foreign-travel-advice/:country/:part' => 'travel_advice#show'
+
+  get '*path(.:locale)(.:format)' => 'content_items#show',
+      constraints: {
+        format: /atom/,
+        locale: /\w{2}(-[\d\w]{2,3})?/,
```
nickcolley: This regex is gnarly for my simple monkey brain, should we have a few tests around this?

fofr: There is a test: https://github.com/alphagov/government-frontend/pull/257/files#diff-7dd9638113bd162736166ccac21fae7cR25

It tests every supported locale against the regex. If we add a new locale that's not supported, the test will fail.

config/routes.rb
```diff
@@ -2,10 +2,14 @@
...
-    # FIXME: Update when https://trello.com/c/w8HN3M4A/ is ready
-    r.get 'foreign-travel-advice/:country/:part' => 'travel_advice#show'
-    r.get '*path' => 'content_items#show', constraints: { path: %r[.*] }
-  end
+  get 'healthcheck', to: proc { [200, {}, ['']] }
+
+  # FIXME: Update when https://trello.com/c/w8HN3M4A/ is ready
+  get 'foreign-travel-advice/:country/:part' => 'travel_advice#show'
+
+  get '*path(.:locale)(.:format)' => 'content_items#show',
```
nickcolley: It looks like the only thing that constrains this route is that it's got an /atom/ and a valid locale.

What happens when you try to render something that is not a travel advice format? Does this work?

fofr: It returns a 406: ba4a046 (ie not an accepted format)
If the templates were added, they would work.

app/presenters/travel_advice_presenter.rb
```diff
@@ -1,3 +1,5 @@
+require 'htmlentities'
```
fofr: Hmm. I was following prior art: https://github.com/alphagov/multipage-frontend/blob/4bcfa99526be326888c04eaf12598785a6920e00/app/presenters/travel_advice_presenter.rb#L1

fofr: Fixed and rebased.

PR #261, opened 2017-02-27
Fix #181: Highlight nav correctly for news articles

### LINKED ISSUE #181: Current navigation item not highlighted
image

image

dsingleton:
Hiya 👋 

I spotted the broken active nav when someone on Twitter linked to a news article. There was an issue already. This fixes it.

---

Current nav is based on part of the path, which works for common cases like 'publications', but some don't follow the convention, like news articles (with 'government/news/slug') that sit under 'announcements'.

Add a hash to map paths that don't follow the convention.

Add tests to for news articles getting the right active nav, and another to a format that does follow the convention, to ensure that still works.

I'm not sure if there are formats other than news articles that need a mapping, but they'd be easy to add now.

---

**Cyber Screenshot:**

screen shot 2017-02-27 at 22 44 46

🎉

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -10,6 +10,18 @@ module ApplicationHelper
   def active_proposition
     # Which of the government sections is the page part of?
     # Derive this from the request path, eg: /government/consultations => consultations
-    request.original_fullpath.split('/')[2]
+    active = request.original_fullpath.split('/')[2]
+    active_proposition_mapping.fetch(active, active)
+  end
+
+private
+
+  def active_proposition_mapping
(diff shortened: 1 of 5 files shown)


fofr:
PR looks good, I don't think it fixes #181 though as it doesn't fix the example in the screenshots.

dsingleton:
You're right - it needs more mappings and tests to cover them.

> I'm not sure if there are formats other than news articles that need a mapping, but they'd be easy to add now.

Happy to add announcements and update this PR later on.

dsingleton:
@fofr PR updated to fix highlighting nav for 'fatalities' too.

PR #264, opened 2017-03-01
Upgrade govuk_ab_testing to 1.0.1

carvil:
Main changes are:
- passing in the request headers to the variant checker;
- setting the acceptance testing framework in the configuration of the
  gem.

Gem changes: https://github.com/alphagov/govuk_ab_testing/pull/20

Trello: https://trello.com/c/t4n8kPpx/434-make-sure-collections-can-be-toggled-with-the-a-b-test-and-react-to-it

### DIFF
--- a/app/models/education_navigation_ab_test_request.rb
+++ b/app/models/education_navigation_ab_test_request.rb
@@ -8,7 +8,7 @@ class EducationNavigationAbTestRequest
   def initialize(request, content_item)
     @content_item = content_item
     @ab_test = GovukAbTesting::AbTest.new("EducationNavigation", dimension: 41)
-    @requested_variant = @ab_test.requested_variant request
+    @requested_variant = @ab_test.requested_variant(request.headers)
   end
 
   def ab_test_applies?
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -121,12 +121,11 @@ class ContentItemsControllerTest < ActionController::TestCase
 
     with_variant EducationNavigation: "A" do
       get :show, params: { path: path_for(content_item) }
-      assert_equal [], @request.variant
+      refute_match(/A Taxon/, taxonomy_sidebar)
     end
 
     with_variant EducationNavigation: "B" do
       get :show, params: { path: path_for(content_item) }
-      assert_equal [:new_navigation], @request.variant
       assert_match(/A Taxon/, taxonomy_sidebar)
     end
   end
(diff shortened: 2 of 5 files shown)


app/models/education_navigation_ab_test_request.rb
```diff
@@ -8,7 +8,7 @@ class EducationNavigationAbTestRequest
   def initialize(request, content_item)
     @content_item = content_item
     @ab_test = GovukAbTesting::AbTest.new("EducationNavigation", dimension: 41)
-    @requested_variant = @ab_test.requested_variant request
+    @requested_variant = @ab_test.requested_variant(request.headers)
```
mcgoooo: this does not the govuk_ab_testing, i.e. why pass the headers as opposed to the request as [shown in the docs](https://github.com/alphagov/govuk_ab_testing#usage)

carvil: Apologies @mcgoooo , those docs are now outdated. I will fix that in a different PR.

The gem `govuk_ab_testing` only needs information in the request headers, so only passing that in makes it clearer what the API needs. Furthermore, it makes testing the A/B tests easier as different acceptance testing frameworks (capybara, etc) treat a request object differently, so accessing the headers is widely different).

Does that make sense?

test/controllers/content_items_controller_test.rb
```diff
@@ -119,14 +119,13 @@ class ContentItemsControllerTest < ActionController::TestCase
 
     content_store_has_item(content_item['base_path'], content_item)
 
-    with_variant EducationNavigation: "A" do
+    with_variant EducationNavigation: "A", assert_meta_tag: false do
```
suzannehamilton: Why does the A variant have `assert_meta_tag: false`? Shouldn't the meta tag be added to both variants?

app/models/education_navigation_ab_test_request.rb
mcgoooo: yep, totally :+1:

test/controllers/content_items_controller_test.rb
mcgoooo: i believe that is the default case?

carvil: The gem is memoising `analytics_meta_tags`, which means that when we use `with_variant` more than once in the same test case, we keep the first meta tag. This is a bug, which I am fixing here: https://github.com/alphagov/govuk_ab_testing/pull/24

When that's merged, I will bump the version of the gem here and remove that `assert_meta_tag: false`.

suzannehamilton: 👍

PR #265, opened 2017-03-01
Render specialist documents

fofr:
https://trello.com/c/h9hITGPy/158-render-specialist-content-in-government-frontend
Tests depend on merge and deploy of https://github.com/alphagov/govuk-content-schemas/pull/548

This is the first part of switching specialist content to government-frontend

* Renders with an incomplete metadata and footer
  * Metadata is pending https://trello.com/c/6CBttvZL/ and is covered by a later story: https://trello.com/c/al2fPH5p/
* Renders with existing contents list rather than the nested version currently on specialist frontend
* Use the oldest entry in the change history to determine first published date – this matches specialist frontend but should be updated when https://trello.com/c/xCJ3RN6W/ is ready
* Display change history in reverse chronological order, matching all other formats

Note: This doesn't include recently added button support: https://github.com/alphagov/specialist-frontend/pull/138

## Screenshots

| Old | New |
|--|--|
|cma-cases-old|cma-cases-new|
|aaib-old|aaib-new|

### DIFF
--- a/app/presenters/updatable.rb
+++ b/app/presenters/updatable.rb
@@ -1,6 +1,6 @@
 module Updatable
   def published
-    display_date(content_item["details"]["first_public_at"])
+    display_date(first_public_at)
   end
 
   def updated
@@ -9,9 +9,14 @@ module Updatable
 
   def history
     return [] unless any_updates?
-    return [] unless content_item["details"]["change_history"].present?
+    reverse_chronological_change_history
+  end
 
-    content_item["details"]["change_history"].map do |item|
+private
+
+  def change_history
+    changes = content_item["details"]["change_history"] || []
+    changes.map do |item|
       {
         display_time: display_date(item["public_timestamp"]),
         note: item["note"],
@@ -20,13 +25,21 @@ module Updatable
     end
   end
 
-private
+  # The direction of change history isn’t guaranteed
+  # https://github.com/alphagov/govuk-content-schemas/issues/545
+  def reverse_chronological_change_history
+    change_history.sort_by { |item| DateTime.parse(item[:timestamp]) }.reverse
+  end
 
   def any_updates?
-    if (first_public_at = content_item["details"]["first_public_at"]).present?
+    if first_public_at.present?
       DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
     else
       false
     end
+  end
+
+  def first_public_at
+    content_item["details"]["first_public_at"]
   end
 end
--- a/app/views/content_items/specialist_document.html.erb
+++ b/app/views/content_items/specialist_document.html.erb
@@ -1,0 +1,23 @@
+<%
+  content_for :simple_header, true
+  content_for :title, @content_item.page_title
+%>
+
+<%= render 'shared/title_and_translations', content_item: @content_item %>
+<%= render 'shared/metadata', content_item: @content_item %>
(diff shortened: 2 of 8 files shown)


app/presenters/updatable.rb
```diff
@@ -20,13 +25,21 @@ def history
...
+    if first_public_at.present?
       DateTime.parse(content_item["public_updated_at"]) != DateTime.parse(first_public_at)
     else
       false
     end
   end
+
+  def first_public_at
+    content_item["details"]["first_public_at"]
+  end
```
mcgoooo: just wondering if we should just use the specialist document presenter way of doing this everywhere so we don't hit that snag again until we have a reliable fix? question more than comment

app/views/content_items/specialist_document.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    </div>
+  <% end %>
+  <div class="column-two-thirds <% unless @content_item.contents.any? %>offset-one-third<% end %>">
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction %>
+  </div>
+</div>
+
+<%= render 'shared/footer', @content_item.document_footer %>
```
mcgoooo: i presume we will be handling the normalisation of this and travel advice later?

app/presenters/updatable.rb
fofr: All other formats that display published/updated uses `first_public_at` at the moment. I think it's ok to say "specialist is different" and leave as-is for now. Might need to re-consider when we bring in other formats.

When the data upstream is fixed we'll only need to change it in two places.

app/views/content_items/specialist_document.html.erb
fofr: Travel advice doesn't have a footer.

But yes, normalisation between the similar contacts and travel advice can happen when all has been merged.

fofr: I prefer to do those refactors as separate PRs

app/presenters/updatable.rb
mcgoooo: :+1:

app/views/content_items/specialist_document.html.erb
mcgoooo: yep no worries 👍

app/presenters/updatable.rb
```diff
@@ -20,13 +25,21 @@ def history
...
-private
+  # The direction of change history isn’t guaranteed
+  # https://github.com/alphagov/govuk-content-schemas/issues/545
+  def reverse_chronological_change_history
+    change_history.sort_by { |item| DateTime.parse(item[:timestamp]) }.reverse
+  end
 
   def any_updates?
-    if (first_public_at = content_item["details"]["first_public_at"]).present?
+    if first_public_at.present?
```
bevanloon: This implies that we `first_public_at` might not exist. I don't think we have an explicit test for this case? 
If we can get a situation of not having `first_public_at` we should perhaps test for it. If we never get a situation where that can happen then we can do away with the conditional here.

h-lame: It can be blank for first drafts (I don't know if government-frontend is used to preview content in the draft stack though).

fofr: It doesn't exist for specialist_documents, the subject of this PR

PR #269, opened 2017-03-02
Track document collections as `finding` pages in Google Analytics

suzannehamilton:
Add meta tag which marks document collection pages as being in the `finding` (navigation) stage of a user journey, as opposed to the `thing` (content) stage.

This will enable us to analyse how users navigate the site so that we can work out whether the new navigation pages are helping users find what they need.

Trello: https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -42,6 +42,10 @@ class ContentItemPresenter
     @nav_helper.taxonomy_sidebar
   end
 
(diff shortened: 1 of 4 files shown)


suzannehamilton:
Closing because we're going to approach this differently, using the content store to add the `finding` vs `thing` analytics classification.

PR #272, opened 2017-03-03
Remove Whitehall header navigation from new navigation pages

alecgibson:
The Whitehall navigation in the header is not compatible with the new Education navigation being A/B tested.

This change removes that navigation for all "B" variants of the education navigation pages.

### Trello

https://trello.com/c/PkN40aGb/479-don-t-show-whitehall-navigation-in-government-frontend-a-b-test

### DIFF
--- a/app/views/content_items/detailed_guide.html+new_navigation.erb
+++ b/app/views/content_items/detailed_guide.html+new_navigation.erb
@@ -1,4 +1,5 @@
 <%= content_for :title, @content_item.page_title %>
+<%= content_for :simple_header, true %>
 
 <div class="grid-row" data-module="sticky-element-container">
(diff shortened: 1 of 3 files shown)


tijmenb:
How is it not compatible?

alecgibson:
@carvil? I think does the navigation just not make sense? Or does it give us access to non-guidance stuff?

carvil:
@tijmenb education related content doesn't have anything to do with those whitehall categories, so having those links at the top will not help users trying to find related content.

screen shot 2017-03-03 at 12 09 19

It also adds a lot of visual noise to these pages, now that we have breadcrumbs + sidebar. This was discussed with @alextea and Holly. Does that make sense?

PR #274, opened 2017-03-03
Add print view for travel advice

fofr:
https://trello.com/c/MluRcP1W/138-2-render-travel-advice-print-page-in-government-frontend-m

* Add route for handling print requests
* Render a print view showing all parts
* Use variants to indicate which template to render
* Render summary and all parts using standard template rather than slimmer’s print version – this applies the print styles we use across GOV.UK rather than the broken bespoke slimmer ones
* Indicate that the link to the print page is an alternate version
* Don’t index the print version
* Add “Print” to title element for unique title

Note that we opted against general line-length improvements to print: https://github.com/alphagov/static/pull/913

When reviewing I recommend comparing the PDF versions. (ProTip: You can drag and drop PDFs to PRs)

| Old | New |
|--|--|
|[Old print view (PDF)](https://github.com/alphagov/government-frontend/files/816749/old-print.pdf)|[New print view (PDF)](https://github.com/alphagov/government-frontend/files/816748/new-print.pdf)|
|old-print-view|new-print-view|

### DIFF
--- a/app/views/content_items/travel_advice.html+print.erb
+++ b/app/views/content_items/travel_advice.html+print.erb
@@ -1,0 +1,34 @@
+<%
+  content_for :title, "Print #{@content_item.page_title}"
+  content_for :simple_header, true
+  content_for :extra_head_content do %>
+  <meta name="robots" content="noindex, nofollow">
+  <script>window.onload = function() { window.print(); }</script>
+<% end %>
+
+<div class="grid-row travel-advice-print">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title', @content_item.title_and_context %>
+
+    <section>
+      <h1 class="part-title">
+        <%= @content_item.current_part_title %>
+      </h1>
+      <%= render 'shared/travel_advice_summary', content_item: @content_item %>
+      <%= render 'govuk_component/govspeak',
+          content: @content_item.current_part_body,
+          direction: page_text_direction %>
+    </section>
+
+    <% @content_item.parts.each do |part| %>
+      <section>
+        <h1 class="part-title">
+          <%= part['title'] %>
+        </h1>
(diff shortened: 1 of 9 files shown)


app/views/content_items/travel_advice.html+print.erb
```diff
@@ -0,0 +1,34 @@
+<%
+  content_for :title, "Print #{@content_item.page_title}"
+  content_for :simple_header, true
+  content_for :extra_head_content do %>
+  <meta name="robots" content="noindex, nofollow">
+  <script>window.onload = function() { window.print(); }</script>
```
mcgoooo: should we maybe put this in an external ajavscript for when we do csp?

fofr: Seems like overkill. I'd expect to use a nonce, etc. Can fix when we use CSP. (This is a direct port)

mcgoooo: :+1:

nickcolley:
I'm not sure we should include the govuk_template layout in the new version?

fofr:
@nickcolley What's your specific objection to it?

1. That the print view doesn't look different to other pages, before being printed?
2. The risk that the header/footer will be accidentally printed?

Or something else?

nickcolley:
I don't have a good objection other than it feels odd to have a 'print' page that could be confused for a normal page.

mcgoooo:
Generally i think this is a good idea, it is slightly odd, but the user owudl only see it on accident anyway, so good to give them an option to navigate away

nickcolley:
I've made a card to look into making the print view look more like a print view if you cancel the print dialog etc.

fofr:
Story is here: https://trello.com/c/UgJijsaL/182-update-travel-advice-print-specific-page

PR #275, opened 2017-03-03
[discussion only] alternative to plek env

mcgoooo:
this pull is only for discussion, and definitely not saying this is the correct way, just illustrating how we used to do it at @mintdigial 

so everything is contained in an env yml file, all the app config variables, and then can be done by standard env variables in a production environment.

for me this provides the following advantages
1. one central way to handle app config
2. urls are obvious and explicit
3. dependencies are recorded in code in one central place as opposed to littered around all working in various different ways
4. makes sure keys are not accidentally checked in due to the process being defined and easy
5. multiple configs can easily live in the same place, just comment/out/in
6. works exactly the same as env variables, so easy for production environments and heroku

the weaknesses as far as i can see with this implementation that i can see are as follows, please add more :D

1. across app config
  this could be solved by [loading a common file](https://github.com/philnash/envyable#other-applications) from one of the govuk gems

2. could be said convention over configuration is preferrable
personally i think, in the case of config, it is good to be explicit

3. the gem needs work on truthy falsy values imo

The motivation for this comes out having problems setting up projects on your local machine, and it's somewhat inconsistent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@
 /tmp
 
 /public/government-frontend
+
+# Don't check in credentials 
+config/env.yml
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gem 'airbrake-ruby', '1.5'
 
 gem 'govuk_frontend_toolkit', '5.1.0'
 gem 'logstasher', '0.6.1'
-gem 'plek', '1.11'
 gem 'rails', '~> 5.0'
 gem 'sass-rails', '~> 5.0.4'
 if ENV['SLIMMER_DEV'] == "true"
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 ruby File.read(".ruby-version").strip
-
+gem 'envyable', require: 'envyable/rails-now'
 gem 'airbrake', '~> 5.5'
 gem 'airbrake-ruby', '1.5'
 
@@ -10,7 +10,7 @@ gem 'logstasher', '0.6.1'
 gem 'plek', '1.11'
 gem 'rails', '~> 5.0'
 gem 'sass-rails', '~> 5.0.4'
-if ENV['SLIMMER_DEV']
+if ENV['SLIMMER_DEV'] == "true"
   gem 'slimmer', path: '../slimmer'
 else
   gem 'slimmer', '9.6.0'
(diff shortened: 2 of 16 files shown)


ijmad:
There is some logic (albeit fairly light) in Plek for production-like environments that you lose by stripping it out wholesale. It is interesting to wonder, though, if that logic could be encoded elsewhere, like in deployment scripts, to set up env variables in production at that point. Perhaps that is more 12 factor?

MatMoore:
I think regardless of what we do with plek, we could explicitly set the environment variables in development, and it would make the app easier to use without inside knowledge of GOV.UK. We already added dotenv to whitehall, which does pretty much the same thing as envyable.

Plek can be entirely overridden by the environment already https://github.com/alphagov/plek/blob/master/lib/plek.rb#L166

tijmenb:
If we want to set env vars for all apps, this is the place:

https://github.com/alphagov/govuk-puppet/blob/8a2696f6c7572e477cfcc2989a33be7dc525b167/modules/govuk/manifests/deploy/config.pp#L92-L103

By adding all possible service URLs there, we could get rid of `Plek`.

tijmenb:
Another upside of using this system is that we can require certain environment variables to be set. We've been bitten a lot by typos and missing `ENV["ERRBIT_API_KEY"]` over the last year if I recall correctly. We didn't notice them missing, because they'd just be `nil`. If we could use `ENV.fetch("ERRBIT_API_KEY")` instead we'd notice much sooner.

mcgoooo:
@ijmad i agree these things should really be in deployment scripts etc. A lot of setups i see have taken the production first and then downported a lot of things development, i.e. favouring less deployment complexity over more complex/harder to understand development environments. 

I believe this is the wrong way round in some ways, and doing this way is leaking over boundaries.  it is more common for a developer to have to change things/understand how things work than understand the deployment scripts. 

this has made harder in the current setup by subtle differences in how all our apps handle things, and plek has made it easier to mix code and config, so you have to go searching. This combined with [a](https://github.com/alphagov/slimmer/blob/83287c1750342ee88a8a9b6cd177dc9e5d88c900/lib/slimmer/component_resolver.rb#L54)  [lot](https://github.com/alphagov/slimmer/blob/e4a951e64092f9672097fea0bdc39096853ff571/lib/slimmer/i18n_backend.rb#L34] of hidden dependencies you cannot see hidden makes for a slightly _magical_ app 

This may be a bit of a frontender centric view on this, but setting env variables in deployment scripts, and abstracting them out makes for a more robust system imo, [twelve factor](https://12factor.net/config) points as this is the way to go.   

it also makes the barrier to entry of just starting an app harder, which makes it less likely to get people like designers tinkering and testing things in a real world app as it's quite involved process to get our stuff running.

mcgoooo:
gonna close this for the moment as it would need a wider range discussion

PR #280, opened 2017-03-06
Add breadcrumbs to contacts page

mcgoooo:
Ideally this would be coming from the parents links in the content api, but
this is blocked until the publishing platform has completed
https://trello.com/c/0OvN0dtQ/

we wish to get the page out before the code rots, so we copied over
the way contacts-frontend was doing it initially

is dependent on 
https://github.com/alphagov/govuk-content-schemas/pull/556

### DIFF
--- a/app/presenters/contact_presenter.rb
+++ b/app/presenters/contact_presenter.rb
@@ -134,6 +134,31 @@ class ContactPresenter < ContentItemPresenter
     whitelisted_paths.include?(content_item["base_path"])
   end
 
+  def breadcrumbs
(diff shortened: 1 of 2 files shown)


app/presenters/contact_presenter.rb
```diff
@@ -134,6 +134,28 @@ def show_webchat?
     whitelisted_paths.include?(content_item["base_path"])
   end
 
+  def breadcrumbs
+    org         = content_item["links"]['organisations'].first
```
fofr: How does this test with no orgs? (empty array and when omitted)

PR #283, opened 2017-03-08
Improve layout of contacts content

fofr:
Make contacts more like contacts frontend:
* Only render phone headings when there are multiple numbers
* Render telephone numbers in bold so they can be found easily

General improvements:
* Move description of phone number above the number, this is consistent with postal addresses and works well with existing content. eg Description says when you should use number and what things you'll need when you make the call

On contacts, when many phone numbers are listed, these are presented using h3 titles (h2: Phone, h3: Type of number). Within this “Opening times” and “Best time to call” are presented as h4s.
In a list of numbers the distinction between phone numbers and these h4 headings was being lost, making the list hard to scan.

* Reduce the visual importance of  specific h4s by making them a normal
font weight. This corrects the visual hierarchy of the content while
preserving semantics.
* Add colons to headings to make them more clearly headings, this
matches contacts-frontend

## Screenshots

| Live | Old frontend | New frontend |
|--|--|--|
|contact-1-old|contact-1-old-gov|contact-1-new|
|contact-2-old|No screenshot – errors|contact-2-new
|contact-3-old|contact-3-old-gov|contact-3-new|

### DIFF
--- a/app/assets/stylesheets/views/_contact.scss
+++ b/app/assets/stylesheets/views/_contact.scss
@@ -2,4 +2,13 @@
   .add-title-margin {
     @include responsive-top-margin;
   }
+
+  // FIXME: Overrides specific govspeak heading styles while h3s & h4s look similar
+  // Govspeak component renders h3s and h4s at the same font size with different margins
+  // Reduce the visual importance of specific h4s by making them a normal font weight.
+  // This gives content correct visual hierarchy and semantics
+  // `normal-weight` is not a class used by the govspeak component
+  .govuk-govspeak .normal-weight {
+    font-weight: normal;
+  }
 }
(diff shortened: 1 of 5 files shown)


nickcolley:
With the second example, are we okay with having the 'Find out about call charges' so far at the bottom?

fofr:
@nickcolley Any suggestions? It does feel too far down, and I reckon the cost of 0300 numbers is relatively unknown.

nickcolley:
I think we could put the existing behaviour back into our version. Or consider changing them so they're repeated, but an anchor link?

mcgoooo:
@nickcolley @fofr i have split out the call charges movement into a seperate ticket which can be seen here
https://trello.com/c/Z2cJ7IMs/181-look-at-design-of-call-charges

PR #288, opened 2017-03-09
Adds metadata page to HTML Attachment print version

elenatanasoiu:
[Trello card](https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents)

### New metadata page 

An extra page has been added. It contains:
- the copyright logo
- document metadata: Web ISBN & contact address
- copyright text

The page will only be visible when printing HTML attachments. 
Here are pages 1, 2 and 3 of a document that is about to be printed. The second page is contains the metadata. 

#### Page 1
screen shot 2017-03-09 at 14 22 30

#### Page 2 - Metadata page
screen shot 2017-03-02 at 11 39 53

#### Page 3
screen shot 2017-03-09 at 14 22 35

#### Original PDF version on which the metadata page was based: 

screen_shot_2016-09-14_at_14 32 28

## Assumptions

1. The year of the copyright notice is considered the year when the document was created

2. There are two sentences that might not make sense when someone is printing a document from their own printer, and have been removed from the final page:

```
Printed on paper containing 75% recycled fibre content minimum.

Printed in the UK by the Williams Lea Group on behalf of the Controller of Her Majesty's Stationery Office
```

3. An assumption was made regarding classifying the ISBN field as a Web ISBN. This is because
in the PDF version there are two ISBNs present in the Metadata page: the Print ISBN and the
Web ISBN. Checking the database, we can find 10 distinct ISBNs assigned to 74 different
HtmlAttachment documents. Through a google lookup, we can identify at least 4 that are
marked as Web ISBNs on pdf documents belonging to gov.uk. Two others are just marked as ISBN. Here are the URLs below:

- Web ISBNs:

https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/446159/Annual_Report_201415.pdf
https://www.google.co.uk/url?sa=t&rct=j&q=&esrc=s&source=web&cd=3&cad=rja&uact=8&ved=0ahUKEwjZiPWejanSAhUhD8AKHRXOBKwQFggiMAI&url=http%3A%2F%2Fwww.milton-keynes.gov.uk%2Fassets%2Fattach%2F22362%2FCode_of_Practice_on_illegal_immigrants_and_private_rented_accommodation__web_.pdf&usg=AFQjCNHtjH9YEn4gT9AwQDAIC9gIfE4Y_A&sig2=jhIz4a1Njaw0hJ8b6SAjlw&bvm=bv.148073327,d.ZGg
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/505928/53949_Armed_Forces_Pay_Review_2016_Accessible.pdf
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/538537/Annual_Report_2016.pdf

- ISBN:

http://content.digital.nhs.uk/catalogue/PUB18117/nhs-out-fram-ind-sep-15-comm.pdf
http://www.housinglin.org.uk/_library/Resources/Housing/OtherOrganisation/An_introduction_to_the_Home_Building_Fund.pdf


app/presenters/html_publication_presenter.rb
```diff
@@ -13,6 +13,18 @@ def contents
     content_item["details"]["headings"].html_safe
   end
 
+  def isbn
+    content_item["details"]["isbn"] if content_item["details"]["isbn"].present?
```
fofr: I'm not sure `if content_item["details"]["isbn"].present?` is necessary, as `content_item["details"]["isbn"]` will return nil if not present.

nickcolley:
Hey :)

With the image, is there a reason it needs whitespace around it?

I've optimized the image https://www.npmjs.com/package/imagemin and cropped it.

This takes it from 11kb to 6kb

open-government-licence-opt

app/views/content_items/_print_meta_data.html.erb
```diff
@@ -0,0 +1,33 @@
+<br/>
```
fofr: This print metadata is specific to HTML publications, can the partial name be clearer?

app/views/content_items/_print_meta_data.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  © Crown copyright <%= @content_item.copyright_year %>
+</p>
+<p>
+  This publication is licensed under the terms of the Open Government Licence v3.0 except where otherwise stated. To view this licence, visit <a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3">nationalarchives.gov.uk/doc/open-government-licence/version/3</a> or write to the Information Policy Team, The National Archives, Kew, London TW9 4DU, or email: <a href="mailto:psi@nationalarchives.gsi.gov.uk">psi@nationalarchives.gsi.gov.uk</a>.
+</p>
+<p>
+  Where we have identified any third party copyright information you will need to obtain permission from the copyright holders concerned.
+</p>
+<p>
+  This publication is available at <a href="www.gov.uk/government/publications">www.gov.uk/government/publications</a>
```
fofr: Can this link to the full publication URL?

app/views/content_items/_print_meta_data.html.erb
```diff
@@ -0,0 +1,33 @@
...
+</p>
+<% unless @content_item.print_meta_data_contact_address.blank? %>
```
fofr: Would `if @content_item.print_meta_data_contact_address` be clearer?

app/views/content_items/_print_meta_data.html.erb
nickcolley: Should we be using CSS over page breaks to create space in the document?

app/views/layouts/application.html.erb
```diff
@@ -6,6 +6,7 @@
   <!--[if IE 6]><%= stylesheet_link_tag "application-ie6" %><script>var ieVersion = 6;</script><![endif]-->
   <!--[if IE 7]><%= stylesheet_link_tag "application-ie7" %><script>var ieVersion = 7;</script><![endif]-->
   <!--[if IE 8]><%= stylesheet_link_tag "application-ie8" %><script>var ieVersion = 8;</script><![endif]-->
+  <%= stylesheet_link_tag "print/print.css", media: "print" %>
```
fofr: As this is specific to HTML publications, it could be included just on the HTML publications template (for now)

app/assets/stylesheets/print/print.scss
```diff
@@ -0,0 +1,5 @@
+// PRINT STYLESHEET COMPILER
+
+.print-wrapper {
```
fofr: Can the compile live at `print.css` rather than `print/print.css`, I'd expect only the includes to be within the print directory.

app/presenters/html_publication_presenter.rb
```diff
@@ -13,6 +13,18 @@ def contents
...
+  end
+
+  def print_meta_data_contact_address
+    content_item["details"]["print_meta_data_contact_address"] if content_item["details"]["print_meta_data_contact_address"].present?
+  end
+
+  def copyright_year
+    content_item["details"]["public_timestamp"].to_date.year if content_item["details"]["public_timestamp"].present?
```
fofr: There are a couple of things calling `content_item["details"]["public_timestamp"]`, can that be put into it's own private method? eg `public_timestamp`

nickcolley:
Thinking generally about this change, if this is important information in a print view, should it also be in the normal view?

fofr:
I have a couple of questions about the feature:
* Does the copyright notice need to be for page 2?
* Does the feature need to be there for all HTML publications, or only certain types?

If it always needs its own page, and needs to be on every HTML pub, then that will lead to many users wasting paper when printing these documents.

elenatanasoiu:
@nickcolley The point of adding those two fields was to make HTML Attachment as similar to the PDF version as possible. 

See Trello card here: https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents

The stated request in the story is: 
```
There are two deliverables from this ticket. The first is to make sure that the print style sheets **include the copyright notice inline**. The second is that **the metadata page must also be included in the print stylesheets**
```

The document discussing what to display from a PDF is here: https://docs.google.com/document/d/1M_VqSkpkLHxED_RhtEkev_4fo4wDz_Vh7UumelDrexg/edit

The first page offers examples of PDF documents: 
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/490962/sos-section70-charities-act-2006-print-ready.pdf

https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/508196/Budget_2016_print_ready_final.pdf

https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/497366/53927-cm_9201-print-ready.pdf

elenatanasoiu:
@nickcolley Thanks for minimizing the image for me! I didn't know it could be minimized further. Will keep that in mind for next time.

app/views/content_items/_print_meta_data.html.erb
elenatanasoiu: It would be, yes. But in this case I'm also testing for "" in which case .blank? is less wordy. I can change it, but we don't want to reserve space for an address in the document if the address is "".

app/views/content_items/_print_meta_data.html.erb
elenatanasoiu: It was originally used in whitehall and I think it was interacting with some other CSS used in the printing version of HtmlAttachments. Which is why I resorted to spacing. 

I've removed them and used CSS here instead.

app/assets/stylesheets/print/print.scss
elenatanasoiu: I've moved it

app/views/content_items/_print_meta_data.html.erb
elenatanasoiu: I've moved it to `content_items/html_publication/_print_meta_data.html.erb`. Hope that makes it a bit clearer.

app/views/content_items/_print_meta_data.html.erb
elenatanasoiu: I've added the full link

app/views/content_items/_print_meta_data.html.erb
elenatanasoiu: Should I change it is to `if @content_item.print_meta_data_contact_address` ?

app/views/layouts/application.html.erb
elenatanasoiu: I've moved this to `html_publication.html.erb`

app/presenters/html_publication_presenter.rb
elenatanasoiu: Done

app/views/content_items/_print_meta_data.html.erb
h-lame: To re-write this using `if` not `unless` we should use `if foo.present?` not `if foo`.  It's subtle, but as @elenatanasoiu already indicated, we don't want to show this when the address is the empty string, and  `if foo` will show it, but `if foo.present?` won't.  The `blank?` and `present?` methods in rails do extra work on most objects to decide beyond simple `nil` vs. `<an instance>` for truthiness (e.g. an empty array, hash, or string are all considered to be `blank?` (e.g. not `present?`), but would be considered truthy by ruby as they are objects that exist).

elenatanasoiu:
I've added a test for the presenter and an integration test.

test/test_helper.rb
```diff
@@ -4,12 +4,18 @@
...
-require 'support/govuk_content_schema_examples'
 require 'capybara/rails'
 require 'slimmer/test_helpers/govuk_components'
 require 'mocha/mini_test'
 
+Dir[Rails.root.join('test/support/*.rb')].each { |f| require f }
+
+class Minitest::Test
+  def teardown
+    Capybara.current_session.driver.clear_memory_cache
```
tijmenb: 👍 for this solution to the caching issue. Might be worth putting in a comment to explain that this is because we re-use URLs and capypolterphantom caches pages.

elenatanasoiu:
After adding Poltergeist, memory caching in Capybara was turned on by default. This meant that some tests were ocassionally failing when they were visiting the same path. In our case, if `government_navigation_test.rb` ran before `phase_label_test.rb`, then the second test would fail. The first test would visit `"/government/case-studies/get-britain-building-carlisle-park"` and cache it. Consequently, the second test would fail with an error that wouldn't obviously indicate caching is the problem. 

```
# Running:

.F

Failure:
PhaseLabelTest#test_Alpha_phase_label_is_displayed_for_a_Case_Study_in_phase_'alpha' [/var/govuk/government-frontend/test/integration/phase_label_test.rb:12]:
Expected false to be truthy.

bin/rails test test/integration/phase_label_test.rb:4

Finished in 3.388734s, 0.5902 runs/s, 0.5902 assertions/s.
```

Due to it taking a while to figure the cause, caching in Capybara has been turned off by default. This is how it's being done: 

```
#test_helper.rb

class Minitest::Test
  def teardown
    Capybara.current_session.driver.clear_memory_cache
  end
end
```

Caching can be turned on for specific tests if necessary.

test/test_helper.rb
elenatanasoiu: Thanks

PR #290, opened 2017-03-09
[discussion] WIP of checking wraith response codes

mcgoooo:
this is a work in progress for checking the wraith response codes,we had
a failed deploy as wraith does not really highlight 500's at all well
and it seems as if another tool may be needed for this. this small hacked
up script will check the status codes across the environments and see
if they match. 

i think this tool may come in useful in general, just want to see if anyone ahs any adverse reactions to this idea. the gif below shows it in action, the gif shows it checking 130 urls on integration and production.

please note its a wip

wraith

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem 'logstasher', '0.6.1'
 gem 'plek', '1.11'
 gem 'rails', '~> 5.0'
 gem 'sass-rails', '~> 5.0.4'
-gem 'parallel'
 if ENV['SLIMMER_DEV']
   gem 'slimmer', path: '../slimmer'
 else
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'logstasher', '0.6.1'
 gem 'plek', '1.11'
 gem 'rails', '~> 5.0'
 gem 'sass-rails', '~> 5.0.4'
+gem 'parallel'
```
fofr: I'd include this in dev/test only

mcgoooo: :+1:

nickcolley:
Going to close this out for now

PR #294, opened 2017-03-10
Run integration tests against random but valid content items

fofr:
* Use GovukSchemas gem to generate random content
* Request random content for any given format using `/random/:schema`, eg /random/publication
* Run integration tests against 10 random publications and contacts, fix errors found

## Examples of random content

example_random_publication
example_random_contact

cc @tijmenb @nickcolley @mcgoooo


nickcolley:
LGTM 👍 

This is great!

PR #300, opened 2017-03-15
Make more formats resilient

fofr:
This tests 4 further formats against random content.

## How should we report bad content?

Frontend principles:
1. Attempt to give the users what they need by rendering anything that's provided by the content-store
2. Report back to us when crucial data is missing so it can be fixed (ie not masking errors)

This PR only does part 1, ie the frontend will not fall over and users will see content, but we won't know that it's broken. When this happens how should we be reporting it? Send to Errbit while still rendering content? This mostly applies to HTML publications and their relationship with other content (organisations, parent).

---

1. HTML Publications
* Protect against parent document type not existing
* Protect against no organisations
* Protect against organisations with no logo

2. Topical event about pages
* Protect against parent in breadcrumb not existing
* Protect against missing end date

Also adds tests against random content for working groups and case studies

### DIFF
--- a/app/presenters/organisation_branding.rb
+++ b/app/presenters/organisation_branding.rb
@@ -1,7 +1,8 @@
 module OrganisationBranding
   def organisation_logo(organisation = default_organisation)
-    logo = organisation["details"]["logo"]
+    return nil unless organisation && organisation.dig("details", "logo")
 
+    logo = organisation["details"]["logo"]
     logo_component_params = {
       organisation: {
         name: logo["formatted_title"],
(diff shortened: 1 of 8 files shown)


nickcolley:
We almost need a way to say we're expecting this data but do not fail if it does not exist, and then log that.

app/presenters/organisation_branding.rb
```diff
@@ -28,7 +29,8 @@ def organisation_brand_class(organisation = default_organisation)
 private
 
   def default_organisation
-    content_item["links"]["organisations"].first
+    orgs = content_item["links"]["organisations"] || []
```
tijmenb: You could also do `content_item.dig("links", "organisations", 0)` here

PR #307, opened 2017-03-22
Migrate answers from frontend

mcgoooo:
this migrates answers to government frontend

there is not much going on here, this pull does not actually do anything until the publisher is changed, which will be done in a further pull. This code is safe to be on live.

| old  | new |
| ------------- | ------------- |
| screencapture-gov-uk-contact-student-finance-england-1490184836899 | screencapture-localhost-3090-contact-student-finance-england-1490184850134 |

### DIFF
--- a/app/assets/stylesheets/views/_answers.scss
+++ b/app/assets/stylesheets/views/_answers.scss
@@ -2,8 +2,4 @@
   .add-title-margin {
     @include responsive-top-margin;
   }
-
-  strong {
-    font-weight: bold;
-  }
 }
--- a/app/assets/stylesheets/views/_answers.scss
+++ b/app/assets/stylesheets/views/_answers.scss
@@ -1,0 +1,9 @@
+.answers {
+  .add-title-margin {
+    @include responsive-top-margin;
+  }
+
+  strong {
+    font-weight: bold;
+  }
+}
--- a/app/views/content_items/answer.html.erb
+++ b/app/views/content_items/answer.html.erb
@@ -7,7 +7,7 @@
         title: @content_item.title %>
     <%= render 'govuk_component/govspeak',
         content: @content_item.body,
-        direction: page_text_direction %>
+        direction: page_text_direction, rich_govspeak: true %>
   </div>
   <%= render 'shared/related_items', content_item: @content_item  %>
 </div>
--- a/app/views/content_items/answer.html.erb
+++ b/app/views/content_items/answer.html.erb
@@ -1,13 +1,15 @@
 <% content_for :simple_header, true %>
-<%= content_for :title, @content_item.title %>
+<% content_for :title, @content_item.title %>
 
-<div class="grid-row sidebar-with-body answers">
+<div class="grid-row sidebar-with-body">
   <div class="column-two-thirds">
     <%= render 'govuk_component/title',
         title: @content_item.title %>
     <%= render 'govuk_component/govspeak',
         content: @content_item.body,
-        direction: page_text_direction, rich_govspeak: true %>
+        direction: page_text_direction,
+        rich_govspeak: true
+    %>
   </div>
   <%= render 'shared/related_items', content_item: @content_item  %>
 </div>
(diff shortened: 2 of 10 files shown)


app/assets/stylesheets/views/_answers.scss
```diff
@@ -0,0 +1,9 @@
+.answers {
+  .add-title-margin {
+    @include responsive-top-margin;
+  }
+
+  strong {
```
nickcolley: Can we put a comment why this is here since it's inconsistent with the other formats?

app/views/shared/_related_items.html+new_navigation.erb
```diff
@@ -0,0 +1,5 @@
+<div class="column-third">
```
nickcolley: Can you show this before and after in screenshots?

mcgoooo: this needs publishing which is not there yet

app/assets/stylesheets/views/_answers.scss
mcgoooo: i dont really know why tbh, as i don't have the history for that, some formats have it some don't

fofr: We should use `rich_govspeak: true` if it's needed for the body text. Or it should be removed if bold is bad.

fofr:
Based on the screenshots it looks like we've lost some of the address styling.

mcgoooo:
@fofr sorry meant to fix that, they are the wrong way round,
fixed now

app/views/shared/_related_items.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="column-third">
```
mcgoooo: i will do this in a separate pull, doesn't seem part of this

app/assets/stylesheets/views/_answers.scss
mcgoooo: @fofr what does that do, this fails the grep test
https://github.com/search?q=org%3Aalphagov+rich_govspeak&type=Code

mcgoooo: where should we document this? i.e. rich_govspeak is not documented

app/views/content_items/answer.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :simple_header, true %>
+<%= content_for :title, @content_item.title %>
```
fofr: Minor: This should be `<%` not `<%=`

app/views/content_items/answer.html.erb
```diff
@@ -0,0 +1,13 @@
...
+
+<div class="grid-row sidebar-with-body answers">
```
fofr: Why do we need the `answers` class here?

app/views/content_items/answer.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        title: @content_item.title %>
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction, rich_govspeak: true %>
```
fofr: Minor: Can the `rich_govspeak` param be on a new line too?

test/integration/answers_test.rb
```diff
@@ -0,0 +1,9 @@
+require 'test_helper'
+
+class AnswerTest < ActionDispatch::IntegrationTest
+  test "renders title, description and body" do
```
fofr: It doesn't render a description.

test/integration/answers_test.rb
```diff
@@ -0,0 +1,9 @@
...
+    setup_and_visit_content_item('answer')
+    assert page.has_text?(@content_item["title"])
+    assert_has_component_govspeak(@content_item["details"]["body"])
+  end
```
fofr: Can you add a test for related items?

fofr:
I've just noticed this is missing the "Last updated: …" at the bottom of the page.

See: https://www.gov.uk/national-minimum-wage-rates

screen shot 2017-04-21 at 11 33 27

PR #315, opened 2017-03-30
Add buttons rendering to specialist documents

nickcolley:
I've added the button in a way that it should be easy to move to be a [GOV.UK Publishing Component](http://govuk-component-guide.herokuapp.com/)

cc @rubenarakelyan I've ended up using the same approach you started with using GOV.UK Elements but I've tried to reduce the amount of global conflict by only adding what's needed with some added comments. Apologies for making you change it the first time!

### DIFF
--- a/app/assets/stylesheets/helpers/_button.scss
+++ b/app/assets/stylesheets/helpers/_button.scss
@@ -1,0 +1,15 @@
+// TODO: We can move this into GOV.UK Publishing Components if it makes sense to do so.
+
+@import "design-patterns/buttons"; // Elements Button import globally depends on mixins defined in here.
+@import "elements/helpers"; // Elements Button import globally depends on mixins defined in here.
+// DEBT/RISK: Includes a .example-highlight-grid global CSS helper
+// https://github.com/alphagov/govuk_elements/blob/874e287/public/sass/elements/_helpers.scss#L12
+// and a reimplementation of .visually-hidden
+// https://github.com/alphagov/govuk_elements/blob/874e287/public/sass/elements/_helpers.scss#L36
+@import "elements/buttons";
+
+.button-info {
+  display: block;
+  margin-top: .5em;
+  max-width: 14em;
+}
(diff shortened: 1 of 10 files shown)


app/views/shared/_button.html.erb
```diff
@@ -0,0 +1,19 @@
+<a
+  class="
+    button
+    <% if local_assigns.include?(:start) && start == true %>
```
fofr: You could put `<% start ||= false %>` at the beginning of the partial to simplify the logic, same with href.

app/assets/stylesheets/helpers/_button.scss
```diff
@@ -0,0 +1,15 @@
+// TODO: We can move this into GOV.UK Publishing Components if it makes sense to do so.
+
+@import "design-patterns/buttons"; // Elements Button import globally depends on mixins defined in here.
+@import "elements/helpers"; // Elements Button import globally depends on mixins defined in here.
+// DEBT/RISK: Includes a .example-highlight-grid global CSS helper
+// https://github.com/alphagov/govuk_elements/blob/874e287/public/sass/elements/_helpers.scss#L12
+// and a reimplementation of .visually-hidden
+// https://github.com/alphagov/govuk_elements/blob/874e287/public/sass/elements/_helpers.scss#L36
+@import "elements/buttons";
```
fofr: Have these issues be raised against elements?
Can elements be easily refactored to avoid these problems?

This _feels_ heavy handed. I'd expect the styling to be ported from `frontend` where it originally lived, and a more focused switch to elements/govuk-frontend to come later.

nickcolley: I originally did as you described but it felt really odd to bring across legacy styles when we have fairly self contained versions of the buttons in GOV.UK Elements.

That said I did this in mind of moving this into static when it's needed so I would be okay doing as you've proposed.

nickcolley:
I think the pointer.png assets need to be brought across too, they don't seem to be rendering on the heroku deployment... Hmm

PR #318, opened 2017-04-10
Remove the phantom gem, the only thing this does is download a binary

mcgoooo:
this gem is not official as far as i can tell,
and only downloads a binary, instead list as a
dependency in the readme, and leave it as an excercise
to the reader/vm to have phantomjs

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ This is a Ruby on Rails application that fetches documents from
 
 - [content-store](https://github.com/alphagov/content-store) - provides documents
 - [static](https://github.com/alphagov/static) - provides shared GOV.UK assets and templates.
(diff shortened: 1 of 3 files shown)


nickcolley:
Should we indicate a hard version that we test against, what version is used in the VM?

PhantomsJS has a 1.x and 2.x release for example.

README.md
```diff
@@ -44,6 +44,7 @@ This is a Ruby on Rails application that fetches documents from
 
 - [content-store](https://github.com/alphagov/content-store) - provides documents
 - [static](https://github.com/alphagov/static) - provides shared GOV.UK assets and templates.
+- [phantomjs](http://phantomjs.org/) used for testing
```
fofr: Could this be more specific: "Used by poltergeist for integration testing"

PR #320, opened 2017-04-10
Use London timezone

fofr:
Most dates in content items are UTC, occasionally some are BST. Any UTC dates during daylight savings time would render an hour out – when the time is around midnight, then the date will also be wrong.

## How bug appeared
via https://govuk.zendesk.com/agent/tickets/2044273
On: https://www.gov.uk/government/publications/your-new-state-pension-explained
```json
"public_updated_at": "2017-04-09T23:15:05.000+00:00",
"public_timestamp": "2017-04-10T00:15:05.000+01:00",
```
Those two dates are the same when taking into account time zone. But they rendered as:
screen shot 2017-04-10 at 13 52 44

## Fix
* Use London timezone
* Update all uses of Time.parse to Time.zone.parse
* Replace most uses of DateTime.parse with Time.zone.parse
* Fix tests now they correctly handle the daylight savings time (DST)
* Include comment about far future dates having no DST
* Update tests to use times in November in GMT, not BST (this has
highlighted an [odd usage of daylight savings timezones](https://github.com/alphagov/govuk-content-schemas/blob/master/formats/consultation/frontend/examples/open_consultation.json#L14) during a non DST
time period in the content schema examples – e.g. 4 Nov is in BST, but
that’s not in a DST period)

### DIFF
--- a/app/presenters/consultation_presenter.rb
+++ b/app/presenters/consultation_presenter.rb
@@ -22,7 +22,7 @@ class ConsultationPresenter < ContentItemPresenter
   end
 
   def opening_date_midnight?
-    Time.parse(opening_date_time).strftime("%l:%M%P") == "12:00am"
+    Time.zone.parse(opening_date_time).strftime("%l:%M%P") == "12:00am"
   end
 
   def closing_date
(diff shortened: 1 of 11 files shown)


nickcolley:
This looks great 👍

PR #321, opened 2017-04-10
Remove redundant stylesheet link

alecgibson:
A snippet of code was used from this website:
https://makandracards.com/makandra/11133-how-to-test-print-stylesheets-with-cucumber-and-capybara
However, this snippet referenced a stylesheet called screen.css that was
never implemented on our end.

Since we never created the stylesheet, but we referenced it, all HTML
publications have been trying and failing to fetch this stylesheet,
resulting in a 404.

We don’t actually need screen.css, so this Pull Request removes the
reference to it.

### DIFF
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -1,7 +1,6 @@
 <%- if params[:medium] == 'print' %>
   <%= stylesheet_link_tag "print.css", :media => "screen" %>
 <%- else %>
-  <%= stylesheet_link_tag "screen", :media => "screen" %>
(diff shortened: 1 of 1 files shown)


fofr:
What's the `if params[:medium] == 'print'` bit for?

fofr:
I think the whole block can be simplified to:
`<%= stylesheet_link_tag "print.css", :media => "print" %>`

alecgibson:
According to @elenatanasoiu Capybara/Cucumber will always load the page as `screen` media, so in order to test the print styles, she passed the query string `?medium=print` in the URL, which is picked up here to load the stylesheet.

PR #322, opened 2017-04-10
Stop the webchat events firing on every ajax check

mcgoooo:
previous functionality was to send analytics on every state change,
this now only sends state changes to analytics as opposed to every
time we poll.

it alos includes lolex so we can do fake timers to test this,
recommendation from @nickcolley


nickcolley:
Thanks Andrew 👍

PR #325, opened 2017-04-18
Splits ISBN into print ISBN and web ISBN

elenatanasoiu:
[Trello card](https://trello.com/c/OEZIdXTE/135-separate-file-attachment-form-and-html-attachment-form)

Depends on: 
https://github.com/alphagov/whitehall/pull/3170
https://github.com/alphagov/govuk-content-schemas/pull/607

## Background
It was pointed out that the "ISBN (web version)" field is showing up not only in the HTML Attachments form, but also in the File Attachments form. This is becoming confusing to people, as they were entering the Print ISBN in that field. The two forms will now show different content. 

## What this does
This PR will separate the "ISBN (web version)" field into "Print ISBN" and "Web ISBN". 
Both fields are optional and will appear (if present) when an HTML publication is printed.

## UI

screen shot 2017-04-18 at 15 36 30

### DIFF
--- a/app/presenters/html_publication_presenter.rb
+++ b/app/presenters/html_publication_presenter.rb
@@ -17,6 +17,10 @@ class HtmlPublicationPresenter < ContentItemPresenter
     content_item["details"]["isbn"]
   end
 
+  def web_isbn
+    content_item["details"]["web_isbn"]
+  end
+
   def print_meta_data_contact_address
     content_item["details"]["print_meta_data_contact_address"]
   end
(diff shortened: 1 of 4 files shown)


elenatanasoiu:
Test will pass once this is merged: https://github.com/alphagov/govuk-content-schemas/pull/607

nickcolley:
Looks good to me from this side 👍

nickcolley:
👍  When green

PR #327, opened 2017-04-18
Move Access-Control-Allow-Origin header into parent controller

steventux:
An update to https://github.com/alphagov/government-frontend/pull/326
We should add this header for all atom pages served by government-frontend
so move the logic into the parent ContentItemsController.


nickcolley:
How does .format work? It is possible with this to spoof the content_type of a request and get any page with CORs headers?

steventux:
@nickcolley https://github.com/rails/rails/blob/5-1-stable/actionpack/lib/action_controller/metal/mime_responds.rb#L31

PR #328, opened 2017-04-19
Render nested contents for specialist documents

fofr:
https://trello.com/c/phs722pV/186-1-add-new-contents-list

* Use an em dash for all contents lists
* Include a two level contents list (based on h2s and h3s)
* Style the h2s without a list marker and in bold
* h3s appear like standard contents lists
* h4s and further, although appearing in the `headers` hash are not displayed – at this level the contents list stops being useful
* Markup mostly ported from specialist-frontend
* Move strip_trailing_colons into shared helper

| Nested | Normal |
|--|--|
|screen shot 2017-04-19 at 11 24 55|screen shot 2017-04-19 at 11 25 00|

On older browsers that don't support `content: ` or `@supports`:
screen shot 2017-04-19 at 12 51 12

### DIFF
--- a/app/assets/stylesheets/helpers/_dash-list.scss
+++ b/app/assets/stylesheets/helpers/_dash-list.scss
@@ -2,20 +2,20 @@
   margin-bottom: $gutter-two-thirds;
 
   li {
-    list-style-type: disc;
-    margin-left: $gutter-half;
-    padding-right: $gutter-half;
+    margin-left: $gutter-half + 5;
+    padding-right: $gutter-half + 5;
     padding-top: $gutter / 6;
+    list-style-type: disc;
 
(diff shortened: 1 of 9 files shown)


nickcolley:
Did you get Mia's eyes on the lack of emdash at the top level?

fofr:
@nickcolley Yep.

nickcolley:
I seem to remember (might be wrong) that some of our other documents also have nested contents list, does this change only update specialist documents to have nested list behaviour?

fofr:
@nickcolley No other format in government-frontend has nested contents. The existing contents list is always one level deep and remains the same. Nested contents only applies to specialist documents.

miaallers:
Looks good to me!

PR #329, opened 2017-04-21
Render help pages

mcgoooo:
I think maybe these two content types could be merged into one
to avoid duplicate code in the front end. With the css i have just followed
the current pattern of repetition, until we tackle it more wholesale.

i changed the gsub's for squish in the govspeak test helper, as it seems more
robust, and a content item was causing leading space problems


app/presenters/govspeak_presenter.rb
```diff
@@ -0,0 +1,7 @@
+class GovspeakPresenter < ContentItemPresenter
+  include Linkable
```
fofr: What's `Linkable` used for?

app/views/content_items/answer.html.erb
```diff
@@ -1,15 +1 @@
...
-        title: @content_item.title %>
-    <%= render 'govuk_component/govspeak',
-        content: @content_item.body,
-        direction: page_text_direction,
-        rich_govspeak: true
-    %>
-  </div>
-  <%= render 'shared/related_items', content_item: @content_item  %>
-</div>
+<%= render 'content_items/short_text/just_govspeak' %>
```
fofr: Could we avoid the `short_text` nesting? Technically everything in government-frontend is short text except HTML pubs.

fofr: I think the label is confusing.

app/views/content_items/short_text/_just_govspeak.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :simple_header, true %>
```
fofr: Technically this isn't _just govspeak_, it's a title, body and related links.
Maybe body_with_related_links, or something similar? This would be consistent with the names of other partials.

app/presenters/help_page_presenter.rb
```diff
@@ -0,0 +1,2 @@
+class HelpPagePresenter < GovspeakPresenter
```
fofr: I think GovspeakPresenter is misleading. Many formats present govspeak.

Perhaps a `body` module that has one method:
```ruby
def body
  content_item["details"]["body"]
end
```

Then the two presenters include that module. Presenters currently use modules rather than inheritance.

fofr: (I think inheritance could go a long way to cleaning up some of presenters, but I'd like to consider more formats when investigation that)

test/test_helper.rb
```diff
@@ -72,7 +72,7 @@ def assert_has_component_title(title)
 
   def assert_has_component_govspeak(content)
     within_component_govspeak do
-      assert_equal content.gsub(/\s+/, ' '), JSON.parse(page.text).fetch("content").gsub(/\s+/, ' ')
+      assert_equal content.squish, JSON.parse(page.text).fetch("content").squish
```
fofr: :+1:

app/presenters/help_page_presenter.rb
mcgoooo: The naming, i agree with, but overall, if we want to clean this up, perhaps we should be putting the tickets in now?

app/presenters/govspeak_presenter.rb
mcgoooo: oversite, i think, will double check

app/views/content_items/answer.html.erb
mcgoooo: 👍

app/presenters/help_page_presenter.rb
mcgoooo: overall, the way they have done modularity in here has not worked out well imo

fofr: >  perhaps we should be putting the tickets in now?
Never a reason not to put a ticket in the backlog for discussion and prioritising.

app/controllers/content_items_controller.rb
```diff
@@ -26,8 +26,6 @@ def present(content_item)
     presenter_name = content_item['schema_name'].classify + 'Presenter'
     presenter_class = Object.const_get(presenter_name)
     presenter_class.new(content_item)
-  rescue NameError
```
fofr: Why have you remove this?

app/presenters/updated.rb
```diff
@@ -0,0 +1,5 @@
+module Updated
```
fofr: Could we call this `LastUpdated`?

test/integration/answer_test.rb
```diff
@@ -5,6 +5,7 @@ class AnswerTest < ActionDispatch::IntegrationTest
     setup_and_visit_content_item('answer')
     assert page.has_text?(@content_item["title"])
     assert_has_component_govspeak(@content_item["details"]["body"])
+    assert page.has_text?("Last Updated: 28 Mai 2015")
```
fofr: Why is this `Mai` and not `May`?

app/controllers/content_items_controller.rb
mcgoooo: because it does nothing and masks real errors, everytime you have an error anywhere in your presenter, it gives the reraised error, which is not helpful

test/integration/answer_test.rb
mcgoooo: it's in welsh

app/controllers/content_items_controller.rb
fofr: This sort of change should be in a separate commit with a clear reason.
I don't think it belongs in this PR.

app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        title: @content_item.title %>
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction %>
+    <div class="meta-data group">
+      <% if @content_item.last_updated %>
+        <p class="modified-date">
+          <%= t 'common.last_updated' %>: <%= @content_item.last_updated %>
```
fofr: The strings for `'common.last_updated'` haven't been added to the locale files.

app/controllers/content_items_controller.rb
mcgoooo: i can put it in a seperate commit, but you think it needs a whole seperate PR, and deploy for something like that, that only effects development?

test/integration/answer_test.rb
fofr: Should `Last Updated:` be welsh too?
https://github.com/alphagov/frontend/blob/master/config/locales/cy.yml#L3

test/presenters/help_page_presenter_test.rb
```diff
@@ -0,0 +1,17 @@
+require 'presenter_test_helper'
+
+class HelpPagePresenterTest < PresenterTestCase
+  include ActionView::Helpers::UrlHelper
```
fofr: Whats the `UrlHelper` needed for?

app/views/content_items/_body_with_related_links.html.erb
mcgoooo: i thought the locales came in from some automagic through the static/slimmer nastiness?

fofr: I think that will switch the text from "Last Updated" to "Last updated" (lowercase U):
https://github.com/alphagov/frontend/blob/master/config/locales/en.yml#L3

app/controllers/content_items_controller.rb
fofr: Looking at Errbit now, I can see this error is being raised on production:
eg `No support for schema "special_route"`

Worth some separate investigation.

app/assets/stylesheets/views/_help-page.scss
```diff
@@ -0,0 +1,9 @@
+.help-page {
+  .add-title-margin {
+    @include responsive-top-margin;
+  }
+
+  .meta-data {
```
fofr: Could this be `last-updated` too? `meta-data` risks being confused with the metadata component.

app/assets/stylesheets/views/_help-page.scss
```diff
@@ -0,0 +1,9 @@
...
+    padding-bottom: 15px;
```
fofr: `15px` should be `$gutter-half`

app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<%= content_for :title, @content_item.title %>
+
+<div class="grid-row sidebar-with-body">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+        title: @content_item.title %>
+    <%= render 'govuk_component/govspeak',
+        content: @content_item.body,
+        direction: page_text_direction %>
+    <div class="meta-data group">
```
fofr: Why do we need the `group` class?

The `<p class="modified-date">` container should be enough. We can remove the `div` container completely.

fofr:
Can you include some screenshots of your latest changes?
From what I can tell there are no styles to render `Last updated` in a lighter colour.

app/views/content_items/_body_with_related_links.html.erb
mcgoooo: tbh, this was copied directly from frontend, that is what is there

fofr:
The title is missing the 'Help' context:
screen shot 2017-04-26 at 17 19 16

mcgoooo:
@fofr already discussed with mia, not really needed. the context

fofr:
The value of the last updated date appears to be different, despite both pointing at the same content item (this is from https://www.gov.uk/help/cookies):
screen shot 2017-04-26 at 17 20 20

mcgoooo:
@fofr i also noticed that, and i believe the public_updated_at would be the correct one to use, if not, the schema is incorrect

fofr:
@mcgoooo It depends on how Publisher has been wired up, I'd ask some of the folks who migrated it to confirm or fix behaviour.

mcgoooo:
@fofr can you tag them in this pull please, i have no idea who they are

mcgoooo:
we are going to await the public updated at to be fixed, for the moment we shall remove it

PR #330, opened 2017-04-21
[Do not merge] Add schema name analytics meta tag

nickcolley:
At the moment, we track the format but the way government-frontend chooses
to render a page is based on the schema_name, this means using the format will
result in duplication of testing, and when testing is slow with visual
regression testing we want to test quickly as possible.

This relies on some upstream changes in Static so that we can add this meta tag against a dimension.

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -1,7 +1,7 @@
 class ContentItemPresenter
   include Withdrawable
 
-  attr_reader :content_item, :base_path, :title, :description, :format, :locale, :phase, :document_type
+  attr_reader :content_item, :base_path, :title, :description, :format, :locale, :phase, :document_type, :schema_name
 
   def initialize(content_item)
     @content_item = content_item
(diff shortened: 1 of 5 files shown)


tijmenb:
Is it an idea to do this in the analytics component so that all apps have the meta tag?

nickcolley:
@tijmenb yeah, we want to make sure this is useful on government-frontend first, then we can put it into the Component.

nickcolley:
We're going to close this after all in favour of it being updated in Static.

https://github.com/alphagov/static/pull/1006

app/presenters/content_item_presenter.rb
```diff
@@ -1,14 +1,15 @@
...
-  attr_reader :content_item, :base_path, :title, :description, :format, :locale, :phase, :document_type
+  attr_reader :content_item, :base_path, :title, :description, :format, :locale, :phase, :document_type, :schema_name
 
   def initialize(content_item)
     @content_item = content_item
     @base_path = content_item["base_path"]
     @title = content_item["title"]
     @description = content_item["description"]
     @format = content_item["schema_name"]
+    @schema_name = content_item["schema_name"]
```
fofr: Could you instead refactor out the deprecated use of format, ie the line above this?
Note: `@format = content_item["schema_name"]`

nickcolley: https://trello.com/c/58vR0SZZ/254-review-use-of-format-in-government-frontend

PR #331, opened 2017-04-24
Adds SRI

elenatanasoiu:
[Trello card](https://trello.com/c/oFhtO0Gm/146-enable-subresource-integrity-sri-on-government-frontend-l)

This will add security checks to verify that the content being served on GOV.UK hasn't been modified by third parties:

- adds sprockets-rails so we can use the SRI check feature
- adds `integrity: true` to javascript_include_tag and stylesheet_link_tag

Depends on:

- static: https://github.com/alphagov/static/pull/1008
- govuk_template: https://github.com/alphagov/govuk_template/pull/294

In order to use SRI, the browser expects CORS to be supported. This is why we need to add `crossorigin: true` alongside `integrity: true`. See more [here](https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/)

### DIFF
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -1,7 +1,7 @@
 <%- if params[:medium] == 'print' %>
-  <%= stylesheet_link_tag "print.css", :media => "screen" %>
+  <%= stylesheet_link_tag "print.css", :media => "screen", integrity: true, crossorigin: "anonymous" %>
 <%- else %>
-  <%= stylesheet_link_tag "print.css", :media => "print" %>
+  <%= stylesheet_link_tag "print.css", :media => "print", integrity: true, crossorigin: "anonymous" %>
 <%- end %>
 
 <%
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -2,11 +2,11 @@
 <html>
 <head>
   <title><%= yield :title %> - GOV.UK</title>
-  <!--[if gt IE 8]><!--><%= stylesheet_link_tag "application" %><!--<![endif]-->
+  <!--[if gt IE 8]><!--><%= stylesheet_link_tag "application", integrity: true, crossorigin: "anonymous" %><!--<![endif]-->
   <!--[if IE 6]><%= stylesheet_link_tag "application-ie6" %><script>var ieVersion = 6;</script><![endif]-->
   <!--[if IE 7]><%= stylesheet_link_tag "application-ie7" %><script>var ieVersion = 7;</script><![endif]-->
   <!--[if IE 8]><%= stylesheet_link_tag "application-ie8" %><script>var ieVersion = 8;</script><![endif]-->
-  <%= javascript_include_tag "application" %>
+  <%= javascript_include_tag "application", integrity: true, crossorigin: "anonymous" %>
   <%= csrf_meta_tags %>
   <%= render partial: 'govuk_component/analytics_meta_tags', locals: { content_item: @content_item.content_item } %>
   <% if @content_item.description %>
--- a/app/views/shared/_webchat.html.erb
+++ b/app/views/shared/_webchat.html.erb
@@ -16,4 +16,4 @@
   </span>
 </span>
 <% # This is inline in the source however slimmer will optimize this. %>
-<%= javascript_include_tag "webchat" %>
+<%= javascript_include_tag "webchat", integrity: true, crossorigin: "anonymous" %>


mcgoooo:
@elenatanasoiu can you fix up the tests on this please

Gemfile.lock
```diff
@@ -297,4 +297,4 @@ RUBY VERSION
    ruby 2.3.1p112
 
 BUNDLED WITH
-   1.14.6
+   1.14.5
```
mcgoooo: maybe upgrade your bundler?

app/presenters/contents_list.rb
```diff
@@ -20,7 +20,8 @@ def contents_link(text, id)
         track_category: 'contentsClicked',
         track_action: 'leftColumnH2',
         track_label: id
-      })
+      }
+)
```
fofr: What's with the unrelated formatting changes in this PR?

Gemfile
```diff
@@ -38,6 +39,7 @@ group :development, :test do
   gem 'govuk-lint'
   gem 'pry-byebug'
   gem 'jasmine-rails', '~> 0.14.0'
+  gem 'jasmine-core', '2.6.0'
```
fofr: What's this dependency for?

app/presenters/contents_list.rb
elenatanasoiu: govuk-lint is complaining about this, so I've fixed it up here

Gemfile
elenatanasoiu: Removed

app/presenters/contents_list.rb
fofr: When running the linter in Jenkins it runs in [diff mode](https://github.com/alphagov/govuk-puppet/blob/7ded141088152004a54441ff36781ad7d3581737/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L388), meaning it should only check lines that have changed.

This change looks wrong:
```ruby
  def contents_link(text, id)
    link_to(
      text,
      "##{id}",
      data: {
        track_category: 'contentsClicked',
        track_action: 'leftColumnH2',
        track_label: id
      }
)
  end
```

fofr: Unrelated linting changes are usually put into separate commits/PRs.

mcgoooo:
@fofr the linter is broken and picks up random changes now. The linter shoudln't really run in diff mode really.

elenatanasoiu:
I'm testing this together with the other dependent commits on integration.

app/presenters/contents_list.rb
h-lame: I agree with @fofr - most of the changes in this PR are unrelated, lets see if we can drop them and still keep the linter happy.

Gemfile
```diff
@@ -24,6 +24,7 @@ gem 'rails-controller-testing', '~> 0.1'
 gem 'govuk_ab_testing', '~> 2.0'
 gem 'htmlentities', '4.3.4'
 gem 'statsd-ruby', '1.3.0', require: 'statsd'
+gem 'sprockets-rails'
```
h-lame: I think it's worth adding a version constraint like "> 3.2" here to make it clear that we get a version that supports the integrity attribute.  As it is, the app already uses sprockets-rails so this entry might be redundant without the version constraint.  I'd probably be happy with dropping this from the Gemfile too TBH.

elenatanasoiu: government-frontend is using sprockets-rails 3.1.1 at the moment, so I'll have to add an explicit version

elenatanasoiu: Although SRI was introduced in 3.1.0 so there's no need to bump it up to 3.2. I'll remove it instead.

h-lame: oh, my bad, I thought it was 3.2

fofr:
What's the `crossorigin: "anonymous"` part for? Looked for details in PR or commit but couldn't see.

elenatanasoiu:
In the description for [What is SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) you can see it also adds `crossorigin="anonymous"` alongside the `integrity` attribute. 

In order to use SRI, the browser [expects CORS to be supported](https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/):

>An important side note is that for Subresource Integrity to work, the CDN must support Cross-Origin Resource Sharing (CORS). The crossorigin attribute in the above code snippet enforces a CORS-enabled load. The anonymous value means that the browser should omit any cookies or authentication that the user may have associated with the domain. This prevents cross-origin data leaks, and also makes the request smaller.

PR #335, opened 2017-04-27
Linting answers presenter

mcgoooo:
did not get picked up on build on the pull for some reason


issyl0:
Looks good. How strange that it wasn't picked up in the first place. *shakes fist at the linter*

PR #338, opened 2017-05-02
Add support for rendering guides

fofr:
* Refactors part logic from travel advice to make it re-usable with guides (and other formats)
* Add guides styles, template and tests
* Include a guides print view for rendering all pages
* Accept guides with no parts, but send airbrake notifications when detected
* Include support for guides with only one part (ie no part navigation, print link, etc)

## Omissions

PR excludes important features which will block switching the rendering app. These will be in later PRs:

* Disable YouTube video expansion using a flag on govspeak component
* Tweaks to govspeak component typography for multi page guides
* ~~A collapsing mobile navigation~~, being omitted from design
* "Last updated" at the bottom of the page (see: https://trello.com/c/83XgOEGO/102-fix-dates-presented-to-publishing-api-from-publisher)

## Requesting content item from part rather than parent

Details in https://github.com/alphagov/government-frontend/commit/1767b1ce53ffb215ef0a600c9065f6e9772f8afe

The content store used to return a 404 when requesting the content item for a part (eg `/foreign-travel-advice/albania/money`). We needed custom routing for travel advice to request the content via the parent content item (`/foreign-travel-advice/albania/`). Content store has been updated to return a 301 when requesting a document's part, so a request to `/foreign-travel-advice/albania/money` will return the content we need. See: https://trello.com/c/w8HN3M4A/

The returned content item is the same for all parts and performs a 301 for _any part slug_, the frontend must determine which part was requested and whether it was valid.

* Guides do not have a common prefix so cannot be determined to be
content with parts before making a request to the content store
* Guides and travel advice use _mostly_ the same pattern for parts – however travel advice bundles up the summary (ie the default part) differently, whereas the default part for a guide is always the first one in `details.parts`
* Remove special case travel advice controller, route and tests – any
format could have parts
* Pass the requested content item path to each presenter to determine
what part is requested, if any
* Check whether a part is being requested in the
content_item_controller and redirect to base_path if part is invalid
* Use part_slug to determine which part to show, if part is a valid

## Screenshots

| Old | New |
|--|--|
|guides-old|guides-new|


fofr:
@andrewgarner @gpeng I've requested your review on this PR as this refactors some of the routing logic I discussed with you when building travel advice.

PR #339, opened 2017-05-02
Add Wraith task for checking per format off the search api

mcgoooo:
It is now possible to run a rake task and see the top 10 per format from the search api,
using the rake command.

i also moved wraith gemfile out and moved it to the base directory. It was a total pain
on a day to day basis to have to change in and out of the directories, and this also makes
this script workable, without doing weird things like changing the directory on the user.


lib/tasks/wraith.rake
```diff
@@ -0,0 +1,17 @@
...
+  format = args[:format]
+  wraith = YAML::load(File.open('test/wraith/config.yaml'))
+  search = RestClient.get "https://www.gov.uk/api/search.json?filter_content_store_document_type=#{format}&count=10"
+  links  = JSON.parse(search.body)["results"].map { |result| result["link"];  }
+  file_name = "test/wraith/wip-config-#{format}.yaml"
+  wraith["paths"] = {}
+
+  links.each_with_index { |link,index| wraith["paths"]["#{format}#{index}"] = link }
+  File.open(file_name, 'w') {|f| f.write wraith.to_yaml }
+  exec("bundle exec wraith capture #{file_name}")
```
fofr: Should it clean up the WIP file after?

mcgoooo: i decided not to in the end, as it can be useful to look at that afterwards, there will only ever be one per format (already in the gitignore)

theleebriggs:
Just a thought but should this just be a separate tool? It's not really tied to this repo, and would be useful when testing a format change regardless of rendering app. Obvs can be done at a later date.

Gemfile
```diff
@@ -43,6 +43,8 @@ end
 group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
+  gem 'wraith', '~> 3.0'
+  gem 'phantomjs', '~> 2.0'
```
fofr: Not sure we need the phantomjs dependency here as it's preloaded on the VM.

:+1: For combining the Gemfiles.

lib/tasks/wraith.rake
```diff
@@ -0,0 +1,17 @@
+require 'pry'
+require 'rest-client'
+
+desc "check format in wraith."
+task :wraith_format, [:format] do |t, args|
+  format = args[:format]
+  wraith = YAML::load(File.open('test/wraith/config.yaml'))
+  search = RestClient.get "https://www.gov.uk/api/search.json?filter_content_store_document_type=#{format}&count=10"
```
fofr: `filter_content_store_document_type` returns `document_type` not `schema_name`.
Can the task be called "check document type", and format replaced with `document_type` throughout?

eg. `specialist_document` is a format – it's a schema name, but it returns no results: https://www.gov.uk/api/search.json?filter_content_store_document_type=specialist_document&count=10

`cma_case` is a document_type, one of the allowed values for specialist_document schemas, which returns results:
https://www.gov.uk/api/search.json?filter_content_store_document_type=cma_case&count=10

Format is deprecated in schema.

mcgoooo: done, all seems a bit confusing, do we have a data dictionary anywhere on these things?

lib/tasks/wraith.rake
```diff
@@ -0,0 +1,17 @@
+require 'pry'
+require 'rest-client'
+
+desc "check document_type in wraith."
```
fofr: Perhaps: "check top 10 content items for document_type using wraith"

lib/tasks/wraith.rake
fofr: I agree, it's confusing. This PR and those linked should be able to give more context: https://github.com/alphagov/govuk-content-schemas/pull/444

Gemfile
mcgoooo: i don't use the vm, and not everyone does, we shoudln't be coding for the vm imo

fofr: As discussed over Slack:
* We previously avoided phantomjs because of security concerns
* It  previously downloaded the binary over http and didn't do a checksum on the downloaded thing (now works over https)
* It uses a different version to our managed environment, it's hard to keep the two in sync

mcgoooo: removed

PR #341, opened 2017-05-04
Remove pry, accidental include

mcgoooo:


fofr:
It's interesting that our tests didn't fail on this.

PR #344, opened 2017-05-04
Fix CorporateInformationPage Preview bug

emmabeynon:
When previewing a Corporate Information page in Whitehall, a bug was found
where the content didn't display. This is because Corporate Information pages
don't currently use edition links, and so the CorporateInformationPagePresenter
couldn't find `organisations` in the `links` hash.  This fix allows
CorporateInformationPagePresenter to handle this more gracefully.

### DIFF
--- a/app/presenters/corporate_information_page_presenter.rb
+++ b/app/presenters/corporate_information_page_presenter.rb
@@ -9,7 +9,9 @@ class CorporateInformationPagePresenter < ContentItemPresenter
   end
 
   def page_title
-    "#{title} - #{default_organisation['title']}"
+    page_title = title.to_s
+    page_title += " - #{default_organisation['title']}" if default_organisation
+    page_title
   end
 
   def title_and_context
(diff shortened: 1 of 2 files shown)


davidbasalla:
Would it be possible to add a test for this? Maybe with a comment that we're planning to fix this at some point...?

fofr:
My feeling about this part of the schema:
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/corporate_information_page/publisher/details.json#L13-L16

Is that it should behave more like finders do for specialist documents – ie 1 specific edition link that is required, maybe called "organisations":
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/specialist_document/publisher/edition_links.json#L9-L13

Is that the sort of fix you were talking about @davidbasalla?

davidbasalla:
@fofr yes I think so... based on a conversation with @gpeng it sounded like there's some work to be done regarding adding the organisation links to corporate information pages. This PR is intended as more of a fix until we can get to that, so that these pages don't error.

test/presenters/corporate_information_page_presenter_test.rb
```diff
@@ -14,6 +14,11 @@ def format_name
       assert_equal "About us - Department of Health", presented_item.page_title
     end
 
+    test 'does not present an organisation in the title when it is not present in links' do
+      presented_item = presented_item(format_name, "links" => { "organisations" => [] })
```
davidbasalla: I think to fully test the current bug we're seeing, the `links` hash shouldn't even have the organisations key/value present... I think your code implementation would still work, but would good to update this test to be more explicit about the exact conditions under which the bug happens.

app/presenters/corporate_information_page_presenter.rb
```diff
@@ -33,12 +35,18 @@ def extra_headings
 
   def default_organisation
     organisation_content_id = content_item["details"]["organisation"]
-    organisation = content_item["links"]["organisations"].detect do |org|
-      org["content_id"] == organisation_content_id
+    organisation = nil
+    if organisations.present?
+      organisation = organisations.detect do |org|
```
davidbasalla: very minor, this could be one line

davidbasalla:
Looks good, will wait for @gpeng to approve the requested changes!

PR #348, opened 2017-05-08
Allow overriding the defaults for `--live`

boffbowsh:
If you use `./startup.sh --live`, you should also be able to override
the locations of static or the content store using the normal
environment variables

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -5,8 +5,8 @@ bundle install
 if [[ $1 == "--live" ]] ; then
   GOVUK_APP_DOMAIN=www.gov.uk \
   GOVUK_WEBSITE_ROOT=https://www.gov.uk \
-  PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
-  PLEK_SERVICE_STATIC_URI=assets.publishing.service.gov.uk \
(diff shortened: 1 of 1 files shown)


startup.sh
```diff
@@ -5,8 +5,8 @@ bundle install
 if [[ $1 == "--live" ]] ; then
   GOVUK_APP_DOMAIN=www.gov.uk \
   GOVUK_WEBSITE_ROOT=https://www.gov.uk \
-  PLEK_SERVICE_CONTENT_STORE_URI=https://www.gov.uk/api \
-  PLEK_SERVICE_STATIC_URI=assets.publishing.service.gov.uk \
+  PLEK_SERVICE_CONTENT_STORE_URI=${PLEK_SERVICE_CONTENT_STORE_URI-https://www.gov.uk/api} \
+  PLEK_SERVICE_STATIC_URI=${PLEK_SERVICE_STATIC_URI-assets.publishing.service.gov.uk} \
```
cbaines: I've always used `:-`, and had to look up what omitting the colon meant [1], just commenting for reference.

1: https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

PR #351, opened 2017-05-09
Show old related links for selected content items

carvil:
We are in the process of releasing the new navigation for Education
content. One item that needs to be solved from a product perspective is
how to deal with curated related links. Until that's solved, some
content items still need to show the old related links instead of the
new navigation for users in the B group of our A/B test.

These overrides existed in the `frontend` application. Since `answers`
were ported to `government-frontend`, we also need to port the override.

Trello: https://trello.com/c/U6JKZN6s/115-make-sure-we-override-the-related-links-on-some-mainstream-content-that-was-recently-moved-to-government-frontend

### DIFF
--- a/app/views/shared/_related_items.html+new_navigation.erb
+++ b/app/views/shared/_related_items.html+new_navigation.erb
@@ -1,3 +1,15 @@
 <div class="column-third add-title-margin">
-  <%= render partial: 'govuk_component/taxonomy_sidebar', locals: content_item.taxonomy_sidebar %>
+  <% if MainstreamContentFetcher.with_curated_sidebar.include?(content_item.base_path) %>
+    <%=
+      render(
+        partial: 'govuk_component/related_items',
+        locals: content_item.related_items
+      ) %>
+  <% else %>
+    <%=
+      render(
+        partial: 'govuk_component/taxonomy_sidebar',
+        locals: content_item.taxonomy_sidebar
+      ) %>
+  <% end %>
 </div>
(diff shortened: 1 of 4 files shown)


app/views/shared/_related_items.html+new_navigation.erb
```diff
@@ -1,3 +1,15 @@
 <div class="column-third add-title-margin">
-  <%= render partial: 'govuk_component/taxonomy_sidebar', locals: content_item.taxonomy_sidebar %>
+  <% if MainstreamContentFetcher.with_curated_sidebar.include?(content_item.base_path) %>
+    <%=
+      render(
+        partial: 'govuk_component/relted_items',
```
fofr: `'govuk_component/relted_items'`
Typo here.

fofr: Not sure how your tests were passing.

carvil: well spotted! I've fixed it.

carvil: there is no rendering going on locally, only a mock of the component with the data...

fofr: I think this is where using the correct component CSS selectors would have surfaced the bug in tests.

fofr: eg https://github.com/alphagov/government-frontend/blob/master/test/test_helper.rb#L98-L100

fofr: Can you try switching to that?

carvil: @fofr thanks for the references, I've added a new commit for this: f21f4945751c987a57b71f495b6a6af9b51af236

PR #357, opened 2017-05-10
Add pagination to travel advice

fofr:
* When linking to the first part of a guide, link to the base path. eg: www.gov.uk/attendance-allowance/overview is identical to www.gov.uk/attendance-allowance – avoids duplicate URLs with the same content.
* This also works with travel advice summaries. Summaries are not included as a part by the content item, but for all purposes, the frontend renders it as a part. Coerce summary into being the first part, which conveniently has no slug, so links to the base_path like the first part of a guide.
* Now parts logic is common between travel advice and guides the logic for next and previous pagination can be shared
* Add pagination to travel advice, ensuring that the link to the first part is without slug

## Screenshots 
screen shot 2017-05-10 at 11 40 56
screen shot 2017-05-10 at 11 41 11
screen shot 2017-05-10 at 11 41 24

### DIFF
--- a/app/presenters/guide_presenter.rb
+++ b/app/presenters/guide_presenter.rb
@@ -32,36 +32,4 @@ class GuidePresenter < ContentItemPresenter
   def related_items
     @nav_helper.related_items
   end
-
-  def previous_and_next_navigation
-    nav = {}
-
-    nav[:previous_page] = {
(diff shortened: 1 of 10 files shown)


michaelihejirika:
I built a local branch the navigations look fine by me

PR #362, opened 2017-05-12
Content update for UKVI AB Test

ikennaokpala:
Related to #361 
[Trello card](https://trello.com/c/OtqjOlKD/616-5-a-b-copy-test-ukvi)

This PR updates the content as follows:

CHANGE: ^If you’re visiting for 6 months or less, check if you need a Standard Visitor visa.^

TO: ^If you’re visiting for 6 months or less, check if you need a Standard Visitor visa.^

REASON: The action is 'check' and should be in the link.


chao-xian:
The description in the PR looks like the copy paste from the Trello card lost what the link was being applied to so.

ikennaokpala:
@chao-xian Thanks for your observation.. I agree with your comment.

PR #363, opened 2017-05-16
Append draft access token to parts links on guides

fofr:
Fact checkers are given an access token to let them view content without a sign on account. For multi page content this token doesn’t persist when fact checkers click part navigation.

* When tokens exist append them to part links in navigation and pagination
* Loosely based on https://github.com/alphagov/frontend/pull/1133/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR127 (default_url_params don’t work unless you’re
using Rails path helpers)

Ideally government-frontend would only do this on the draft stack, but it has no knowledge of what content store/stack it is using.

cc @davidbasalla @whoojemaflip @suzannehamilton

### DIFF
--- a/app/presenters/guide_presenter.rb
+++ b/app/presenters/guide_presenter.rb
@@ -2,6 +2,8 @@ class GuidePresenter < ContentItemPresenter
   include Parts
   include UkviABTest
 
+  attr_accessor :draft_access_token
+
   def page_title
     if @part_slug
       "#{title}: #{current_part_title}"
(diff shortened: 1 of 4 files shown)


fofr:
Need to deploy this to integration to test end to end.

app/presenters/guide_presenter.rb
```diff
@@ -33,4 +35,21 @@ def print_link
...
+      end
+    else
+      super
+    end
+  end
+
+private
+
+  def draft_access_token_param
+    "token=#{draft_access_token}" if draft_access_token
```
davidbasalla: very minor, this `if` clause seems redundant since we check the same on line 41

fofr:
Tested end to end on Integration. Works as expected.

PR #367, opened 2017-05-19
Back to top arrow style updated

andysellick:
* SVG arrow added to fit with new design
* Font size also adjusted to fit new design

Before: 

screen shot 2017-05-22 at 08 57 19

After:

screen shot 2017-05-19 at 15 31 28

https://trello.com/c/J6YkyF1T/356-deployment-back-to-top-button-2

Now relates to https://github.com/alphagov/static/pull/1067


app/views/shared/_back_to_content_link.html.erb
```diff
@@ -1,11 +1,9 @@
...
-  },
-) %>
+<a class="back-to-content"
+  data-track-category="contentsClicked"
+  data-track-action="backToContentsLinkClicked"
+  data-track-label="contents"
+  data-module="track-click"
+  href="#contents">
+    <svg class="icon" width="14" height="17" viewBox="0 0 14 17" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><g transform="translate(-885.000000, -1380.000000) translate(886.000000, 1380.000000) translate(0.000000, 1.000000)" stroke-width="2" stroke="currentColor" fill="none" fill-rule="evenodd"><polyline points="12 6.61354167 6.10565074 0.719192411 0.312351484 6.51249167"/><path d="M6,1.52415365 L6,15" stroke-linecap="square"/></g></svg>
+    Contents
```
fofr: This should still be translated. It used to be `t('content_item.contents')`.

fofr: You could keep the `link_to` method by doing this:

```erb
<%= link_to(params) do %>
 Link text
<% end %>
```

fofr:
Would be good to show a before/after screenshot.

app/assets/images/arrow-up.svg
```diff
@@ -0,0 +1 @@
+<svg width="14" height="17" viewBox="0 0 14 17" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><g transform="translate(-885.000000, -1380.000000) translate(886.000000, 1380.000000) translate(0.000000, 1.000000)" stroke-width="2" stroke="#005EA5" fill="none" fill-rule="evenodd"><polyline points="12 6.61354167 6.10565074 0.719192411 0.312351484 6.51249167"/><path d="M6,1.52415365 L6,15" stroke-linecap="square"/></g></svg>
```
fofr: Is this file needed?

andysellick: Oops, good point

fofr:
Tested this on old IE and it looks fine:
screen shot 2017-05-22 at 09 54 09

Tested on mobile and the arrow looks too big (font-size is reduced):
screen shot 2017-05-22 at 09 54 53

fofr:
@nickcolley Can you add updated screenshots?

nickcolley:
@fofr Andy is doing some additional work on this

nickcolley:
I think we need to https://github.com/alphagov/government-frontend/blob/master/app/views/shared/_sidebar_contents_with_body.html.erb add this here too

fofr:
I was hoping we could keep the design updates and the "adding to new formats" as separate PRs. One shouldn't block the other.

nickcolley:
image

I think the nesting might be off, since it looks like we're missing the grids gutter padding

andysellick:
Nesting should be fixed now.

nickcolley:
I've noticed for some of the pages the contents link will spill over the end, I wonder if we can address this? It feels really odd at the moment.

screen shot 2017-05-25 at 14 47 14

nickcolley:
Checking the live site, the issue I've raised above is already an issue, so we're going to ship these changes and try and fix the existing behaviour separately.

andysellick:
Fix added to static... https://github.com/alphagov/static/pull/1067

app/views/content_items/document_collection.html.erb
```diff
@@ -42,6 +46,9 @@
       </ol>
     <% end %>
   </div>
+  <div data-sticky-element class="govuk-sticky-element">
```
fofr: Use of the class name `govuk-sticky-element` hints that it's a component, which it isn't.

fofr:
How does this look on right to left content?

andysellick:
I can't find a page with right to left content where this is used. Any suggestions?

andysellick:
Seems to be okay on RTL. Have added some more example links to the card.

PR #372, opened 2017-05-23
Add a meta tag to show if the content item has history

thomasleese:
This will be used by the analytics to include that information as a custom dimension.

[Trello Card](https://trello.com/c/uHiHx5RQ/56-custom-dimension-to-identify-session-where-content-history-is-available)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -12,6 +12,9 @@
   <% if @content_item.description %>
     <meta name="description" content="<%= strip_tags(@content_item.description) %>" />
   <% end %>
+  <% if @content_item.respond_to?(:history) %>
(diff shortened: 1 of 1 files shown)


nickcolley:
Hey @thomasleese can you add a test alongside this? 👍

fofr:
These analytics meta tags are usually added to the component: https://github.com/alphagov/static/blob/master/app/views/govuk_component/analytics_meta_tags.raw.html.erb

thomasleese:
Oh, thanks, I completely missed that!

thomasleese:
Hmm, actually, I don't think I can use that as I want to use the `history` method of the various presenters, otherwise I would have to duplicate the checks in https://github.com/alphagov/government-frontend/blob/master/app/presenters/updatable.rb . The analytics meta tags component is only given the raw content item.

PR #373, opened 2017-05-24
Use memcached in production.

danielroseman:
Memcached is now available on the frontend boxes, so use that instead of the in-process cache.


fofr:
How are we measuring the affect of this?

PR #376, opened 2017-05-31
Reduce timeout from 15s to 10s for Webchat availability polling

nickcolley:
HMRC (Jonathan Lewis I CDIO Digital Services) are trying to reduce the
amount of people going through to their webchat, they've asked us to
increase the timeout so that there's less chance of a user seeing a stale
'available' state which would allow them to proceed.

We want to make sure we can revert and deploy this if HMRC decide this has impacted them negatively.

Marked as [Do not merge] while we talk with HMRC about when we can deploy this.

### DIFF
--- a/app/assets/javascripts/webchat/library.js
+++ b/app/assets/javascripts/webchat/library.js
@@ -7,7 +7,7 @@
 
 
   function Webchat (options) {
-    var POLL_INTERVAL = 15 * 1000
+    var POLL_INTERVAL = 10 * 1000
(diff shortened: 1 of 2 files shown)


fofr:
PR description says “increase timeout”, PR title says “reduce timeout”.

What is the affect of decreasing the polling interval?

nickcolley:
My mistake, they should both be reduce.

They wanted to make sure users were seeing unavailable more often so that they don't go through to the chat and end up in their queue.

PR #381, opened 2017-06-16
Fix intermittent test failure on guides

fofr:
The `JSON.parse` method omitted `content_store_has_item` which meant the content store wasn’t always stubbed with the example – sometimes this test would fail, sometimes it would pass, depending on the tests that ran before it.

Avoid risk that this could happen again by updating `setup_and_visit_content_item` to take URL params.


nickcolley:
🐛  💥  🔫

PR #383, opened 2017-06-21
Add link tracking to HTML Publication contents lists

fofr:
* Track clicks on contents lists
* We can’t modify HTML Pub contents lists easily, instead write a new
module to add event listeners on the parent and track links via click
events within it
* Use category and action defined on parent container
* Use the target of the link as the label

When we move the position of contents lists, and if we use the standard ContentsList functionality, then this custom tracking can be removed.

https://trello.com/c/z3YCn9zf/424-implement-ga-tracking-on-html-publication-content-list

@nickcolley @andysellick


nickcolley:
Tested it using GA Debugger locally, looks good. 👍

PR #384, opened 2017-06-21
Move HTML Publication contents list from top to left

andysellick:
Goes alongside this static PR: https://github.com/alphagov/static/issues/1073

* Contents moved into left nav area and uses ContentsList for output
* Contents adjusted so number and text of link are wrapped in separate spans for better styling and test added
* Contents styled with indented numbering if numbering present
* Contents link overlap fixed with some z-indexing, background and box shadow on contents list, so the contents link disappears behind
* Content heading numbering size reduced and repositioned to be inline
* Publication header (blue box) padding made even and publication date font size reduced and moved up
* Print view fixed following layout changes, print styles updated and hidden crown copyright text moved out of header and below contents list to maintain original print layout
* Right to left content checked and amended following layout changes.

https://trello.com/c/eDpJIHQG/410-html-publication-move-content-list-to-left-hand-side-3

**Before:**

screen shot 2017-06-21 at 16 26 58

**After:**

screen shot 2017-06-23 at 08 28 17

### DIFF
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -58,6 +58,7 @@
     @include grid-column( 1 / 4, $full-width: desktop );
 
     .direction-rtl & {
+      float: right;
       direction: rtl;
       text-align: start;
     }
@@ -111,6 +112,10 @@
 
   .contents-wrapper {
     @include grid-column( 3 / 4, $full-width: desktop );
+
+    .direction-rtl & {
+      float: right;
+    }
   }
 
   .offset-one-quarter {
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -130,18 +130,6 @@
     }
   }
 
-  h3 .number,
-  h4 .number,
-  h5 .number,
-  h6 .number {
-    margin-right: 0.1em;
-
-    .direction-rtl & {
-      margin-right: 0;
-      margin-left: 0.1em;
-    }
-  }
-
   .print-meta-data {
     display: none;
   }
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -55,7 +55,7 @@
   }
 
   .contents-list-wrapper {
-    @include grid-column( 1/4, $full-width: desktop );
+    @include grid-column( 1 / 4, $full-width: desktop );
 
     .direction-rtl & {
       direction: rtl;
@@ -110,7 +110,7 @@
   }
 
   .contents-wrapper {
-    @include grid-column( 3/4, $full-width: desktop );
+    @include grid-column( 3 / 4, $full-width: desktop );
   }
 
   .offset-one-quarter {
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -29,19 +29,16 @@
   }
 
   .publication-header {
+    @include core-19;
     color: $white;
     background: $light-blue;
     padding: $gutter;
     margin-bottom: $gutter * 2;
   }
 
-  .headings {
-    @include core-16;
-  }
-
   // Based on the govuk-title component
   .html-publication-title {
-    margin-bottom: 10px;
+    margin-bottom: $gutter;
 
     .context {
       @include core-27;
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -29,8 +29,6 @@
   }
 
   .publication-header {
-    @include white-links;
-    @include core-19;
     color: $white;
     background: $light-blue;
     padding: $gutter;
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -1,6 +1,8 @@
 @import "grid_layout";
 
 .html-publication {
+  @include sidebar-with-body;
+
   .publication-external {
     @extend %grid-row;
     margin-bottom: $gutter-two-thirds;
@@ -16,9 +18,28 @@
     }
   }
 
+  .publication-header {
+    @include white-links;
+    @include core-19;
+    color: $white;
+
+    background: $light-blue;
+    padding: $gutter;
+    margin-bottom: $gutter;
+
+    &.direction-rtl {
+      direction: rtl;
+      text-align: start;
+    }
+  }
+
+  .headings {
+    @include core-16;
+  }
+
   // Based on the govuk-title component
   .html-publication-title {
-    @include responsive-vertical-margins;
+    margin-bottom: 10px;
 
     .context {
       @include core-27;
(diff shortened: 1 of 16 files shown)


test/presenters/html_publication_presenter_test.rb
```diff
@@ -10,7 +10,11 @@ def schema_name
     assert_equal schema_item("published")['links']['parent'][0]['document_type'], presented_item("published").format_sub_type
     assert_equal schema_item("published")['title'], presented_item("published").title
     assert_equal schema_item("published")['details']['body'], presented_item("published").body
-    assert_equal schema_item("published")['details']['headings'], presented_item("published").contents
+  end
+
+  test 'presents a list of contents extracted from headings in the body' do
+    #assert_equal schema_item("published")['details']['headings'], presented_item("published").contents
```
nickcolley: Nit: Commented code

test/wraith/config.yaml
```diff
@@ -73,6 +73,18 @@ paths:
   contact_9: /government/organisations/hm-revenue-customs/contact/gambling-duties
   contact_10: /government/organisations/hm-revenue-customs/contact/welsh-language-helpline-for-debt-management
 
+  # Most popular HTML publications
```
nickcolley: 👍

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -31,85 +55,81 @@
...
     }
 
-    ol {
-      padding-left: $gutter;
-
-      .direction-rtl & {
-        padding-right: $gutter;
-        padding-left: 0;
-      }
+    .contents-list {
```
nickcolley: Is this contents list different enough to have it's own CSS?

test/helpers/contents_list_helper_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+class ContentsListHelperTest < ActionView::TestCase
+  test "it wraps the number and text in span elements" do
```
fofr: What happens if a header contains something like: "Number 4."

app/assets/stylesheets/views/_html-publication.scss
andysellick: I think so, yes. I looked for alternatives but couldn't find any suitable option. It could have been done as a variant on the existing dash-list but that would have meant having a version like '.dash-list.without-dashes', which seemed confusing.

test/helpers/contents_list_helper_test.rb
andysellick: It will only match the pattern of a number followed by a period at the start of the text of the link. I've now added a test for this specific example though.

app/assets/stylesheets/views/_html-publication.scss
fofr: I think we could probably refactor this in a later PR. eg `contents-list-dashed`, `contents-list-numbered`. I agree with Nick about keeping this together but don't think it should block the initial PR.

fofr:
Are there screenshots showing right-to-left content? What's a good example to check with? (Should we have an example in the wraith config too?)

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -14,11 +21,28 @@
         padding-bottom: $gutter-one-third;
       }
     }
+
+    .direction-rtl & {
```
fofr: Is it worth putting a comment here explaining the need to reset direction?

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -14,11 +21,28 @@
...
+      direction: ltr;
+    }
+  }
+
+  .publication-header {
+    @include white-links;
```
fofr: Does the `publication-header` still contain `white-links`?

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -14,11 +21,28 @@
...
+    @include core-19;
```
fofr: Should this be `core-16`? Then the `.headings` styles could be removed?

app/helpers/contents_list_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ContentsListHelper
+  def insert_spans(text)
+    no_tags = strip_tags(text) #just the text of the link
+    number = /^[0-9]*[.]/.match(no_tags)
+
+    if number
+      link_text = no_tags.sub number.to_s, '' #remove the number from the text
+      text.sub! no_tags, '<span class="contents-number">' + number.to_s + '</span><span class="contents-text">' + link_text.strip + '</span>'
+      text = text.squish.html_safe
```
fofr: Would this be equivalent:
```ruby
"<span class=\"contents-number\">#{number}</span><span class=\"contents-text\">#{link_text.strip}</span>".squish.html_safe
```

Not sure why we need the `sub!` before it.

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -31,85 +55,81 @@
...
-      font-weight: normal;
+        .contents-text {
+          margin-left: 0;
+          margin-right: 1.5em;
+        }
+      }
     }
+  }
+
+  h3 .number,
```
fofr: Shouldn't these styles be in the component?

test/integration/guide_test.rb
```diff
@@ -30,6 +30,7 @@ class GuideTest < ActionDispatch::IntegrationTest
 
   test "draft access tokens are appended to part links within navigation" do
     setup_and_visit_content_item('guide', '?token=some_token')
+
```
nickcolley: my bad 👼

andysellick:
> Are there screenshots showing right-to-left content? What's a good example to check with? (Should we have an example in the wraith config too?)

There are some examples of right to left content in the trello card. INCOMPLETE...

test/wraith/config.yaml
andysellick: Added.

test/helpers/contents_list_helper_test.rb
```diff
@@ -0,0 +1,21 @@
...
+    input = '<a data-track-category="contentsClicked" data-track-action="leftColumnH2" data-track-label="run-an-effective-welfare-system-that-enables-people-to-achieve-financial-independence-by-providing-assistance-and-guidance-into-employment" href="#run-an-effective-welfare-system-that-enables-people-to-achieve-financial-independence-by-providing-assistance-and-guidance-into-employment">1. Run an effective welfare system that enables people to achieve financial independence by providing assistance and guidance into employment</a>'
```
andysellick: They're not - I'd simply copied and pasted a real world example. Will trim it down to be more readable.

app/helpers/contents_list_helper.rb
andysellick: It wouldn't, unfortunately. The reason is that the value of text is a link tag, so we're replacing the text within that tag with the text containing the spans. It would need to be something more like

> <a href="#"><span class= ... (your suggestion) ...</span></a>

...except we don't know what the link will be.

test/presenters/html_publication_presenter_test.rb
```diff
@@ -21,7 +24,7 @@ def schema_name
 
   test 'presents no contents when headings is an empty list' do
```
andysellick: I don't know. Definitely agree we should remove it if not needed.

app/views/shared/_sidebar_contents.html.erb
```diff
@@ -1,10 +1,19 @@
...
+
 <% if contents.any? %>
   <nav role="navigation">
     <h2><%= t("content_item.contents") %></h2>
-    <ol class="dash-list" data-module="track-click">
+    <ol class="<%= list_class %>" data-module="track-click">
       <% contents.each do |heading| %>
         <li>
-          <%= heading %>
+          <% if format_numbers %>
```
andysellick: Refactored.

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -31,89 +54,94 @@
...
 
-  .publication-header {
-    @include white-links;
-    @include core-19;
-    color: $white;
-
-    background: $light-blue;
-    padding: $gutter $gutter $gutter / 3;
-    margin-bottom: $gutter;
+  .contents-list-wrapper {
```
fofr: I'd like to use classes more like the grid ones we use elsewhere, eg: `column-quarter-desktop-only`, which would decouple the contents list from the grid. This can come as a later PR.

app/views/content_items/html_publication.html.erb
```diff
@@ -32,32 +29,30 @@
...
-        <%= t("content_item.contents") %>
-      </h2>
-      <%= @content_item.contents %>
-    </nav>
+<div
+  class="grid-row sidebar-with-body"
+  data-module="sticky-element-container"
+>
+  <% if @content_item.contents.any? %>
+    <div class="contents-list-wrapper">
```
fofr: It's not really a `contents-list-wrapper` but a grid class.

fofr:
Fixed bug with contents list in de0cc2e

screen shot 2017-06-26 at 14 27 40
screen shot 2017-06-26 at 14 30 53

nickcolley:
LGTM 👍

PR #386, opened 2017-06-23
Fix world location links

gpeng:
World location links are becoming base_path-less and will no longer support translations. We still need to generate a link to these in the frontend end. 

Initially, in this commit the link will continue to point to the Whitehall representation of the world location.

In a subsequent commit, once the routing is in place, the `WorldLocationBasePath` class will be updated to generate a link to the new `/world/<slug>` taxon.

This PR needs https://github.com/alphagov/govuk-content-schemas/pull/629 to be deployed before CI will pass.

Part of [Trello](https://trello.com/c/6tio72jl/221-split-world-locations-from-international-delegations-in-whitehall)

### DIFF
--- a/app/services/world_location_base_path.rb
+++ b/app/services/world_location_base_path.rb
@@ -1,0 +1,41 @@
+#This is intended as a temporary solution while dependencies on other apps are
+#resolved. We are migrating world locations to a taxonomy based model but the
+#following problems have arisen that this approach temporarily solves:
+#
+# * we require world locations to be associated with content in Whitehall
+#   for email subscriptionas to work so we can't just get rid of them
+# * we also need them as not all content will be part of the taxonomy
+#   (e.g. news about the country)
+# * the taxon will take the base path (/world/<country-slug>) and so the
+#   WorldLocation can't occupy that
+# * we need to redirect old /government/world/<country-slug> urls to the new
+#   taxon but if we leave WorldLocation with that path it will a) overwrite
+#   the redirect route every time it is updated and b) we'll be knowingly
+#   linking to a redirect
+#
+#A slightly better solution will be for Whitehall to retrieve the taxon path from
+#publishing API and send that with the world location link as an additional field
+#in the links. We'll need the taxonomy to exist in order to implement this. Thiw
+#will still require frontend code to know about the links but will be similar
+#to prior art we have for some links to be affected by elements with `detail`.
+#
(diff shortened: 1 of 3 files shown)


fofr:
> World location links are becoming base_path-less and will no longer support translations. We still need to generate a link to these in the frontend end.

I don't understand. The first sentence appears to refute the second.

My impression is that the content item now will not provide all of the information a frontend needs to render the content. This is one of the principles the frontends expect to follow.

Is this situation a temporary thing? (Specifically things like this: https://github.com/alphagov/government-frontend/pull/386/files#diff-30c3af0b5bd736575c794f5f225deb70R2 shouldn't be frontend concerns)

gpeng:
Yes, this will 'hopefully' be temporary. We need worldwide content to link to the taxon (at `/world/<country-slug>` ultimately but there are a number of problems currently.

* we require world locations to be associated for email subscriptions
* not all content will be part of the taxonomy (e.g. news about the country)
* the taxon will take the base path (`/world/<country-slug>`) and so the WorldLocation can't occupy that
* we need to redirect old `/government/world/<country-slug>` urls to the new taxon but if we leave WorldLocation with that path it will a) overwrite the redirect route every time it is updated and b) we'll be knowingly linking to a redirect

So this 'solution' moves us on from a few of those issues and allows us to deploy the taxonomy. I fully agree that there is non-frontend 'stuff' in here that we need to work out. One solution that I've discussed with @danielroseman is finding some way of associating the world location link with the taxon within the links but that's complicated and this may be the only scenario that requires that functionality. So, in the meantime this was the best I could come up with the unblock the current work.

app/services/world_location_base_path.rb
```diff
@@ -0,0 +1,20 @@
+class WorldLocationBasePath
+  EXCEPTIONAL_CASES = {
```
fofr: Can we add a comment indicating the temporary nature of this. Keen for this not to hang around too long.

gpeng: We most certainly can. Added in 	0b4e917. We appreciate this isn't ideal. Would very much welcome a chat with you next week to discuss options for a better solution.

fofr:
Thanks for adding comment. Happy to discuss next week. Code looks good but will leave for one of your team members to approve.

gpeng:
Awesome. Thanks @fofr

PR #388, opened 2017-06-26
Move the Postcode line in the presented Postal Address

whoojemaflip:
The postcode was displaying in an odd position in address blocks,
between the locality and region. This moves the postcode down a line,
so it will display at the bottom of the address, just above the country.


nickcolley:
For context this matches https://github.com/alphagov/govspeak/blob/6cb3ddf213eb8c37d76910d50237a6bde3e7e65b/config/address_formats.yml#L110

PR #394, opened 2017-06-28
HTML Publications clean-up

fofr:
Some refactoring following #384, changes that we didn't want to block the release of the feature.

* Separate contents list styles and grid classes
* Remove custom HTML contents tracking
* Tweak alignment of HTML publication title and published date
* Combine HTML contents and dashed list styles – creating three modifiers: dashed, nested and numbered (this highlights the need for a local style guide)

## Apply fixes from HTML publications to all contents
A detailed guide with a long contents list:

Before:
screen shot 2017-06-28 at 11 15 06

After:
screen shot 2017-06-28 at 11 14 59

## HTML publication content styles applied to all contents

Multi-line contents list items are easier to read and there's clearer separation between each contents list item.

| New | Old |
|--|--|
|HTML Publications|All other contents|
|1.3 line height with 7.5px padding| 1.5 line height with 5px padding |

screen shot 2017-06-28 at 11 25 00

## Heading tweaks

Current:
screen shot 2017-06-28 at 11 40 37

New:
screen shot 2017-06-28 at 11 40 43

CSV (and before update):
screen shot 2017-06-28 at 11 41 33

cc @nickcolley @andysellick


nickcolley:
This looks sensible 👍

fofr:
Rebased following #395

app/assets/stylesheets/views/_html-publication.scss
```diff
@@ -51,71 +51,23 @@
...
-
-      @include media(false, $desktop-breakpoint) {
-        box-shadow: none;
-      }
     }
   }
 
-  .contents-wrapper {
+  .column-three-quarters-desktop-only {
     @include grid-column( 3 / 4, $full-width: desktop );
```
andysellick: This looks to be collapsing to a single column at 640px, which I think is lower than HTML publications. It gets pretty compressed (particularly the contents list) just above that (in two column layout). Did you consider collapsing earlier?

test/integration/specialist_document_test.rb
```diff
@@ -107,8 +107,8 @@ def test_meta(component)
   test "renders a nested contents list" do
     setup_and_visit_content_item('aaib-reports')
 
-    assert page.has_css?(".dash-list")
-    within ".dash-list" do
+    assert page.has_css?(".contents-list.contents-list-nested")
```
andysellick: Is this line necessary? Presumably test will fail anyway if not present, and this line isn't repeated at line 121 where another test does exactly the same thing. I guess the test might fail in a different way if the line is simply omitted, but if that's the case then it should also occur before line 121 as well.

app/assets/stylesheets/helpers/_contents-list.scss
```diff
@@ -0,0 +1,91 @@
...
+  // Always render the contents list above a
+  // back to contents link
+  position: relative;
+  z-index: 1;
+  background: $white;
+  box-shadow: 0 20px 15px -10px $white;
+
+  li {
+    padding-top: $gutter / 4;
+    line-height: 1.3;
```
andysellick: Genuine question - should there be a unit of measurement for the line height or will all browsers safely default to em? (which is what I presume it's supposed to be?)

app/assets/stylesheets/helpers/_contents-list.scss
```diff
@@ -0,0 +1,91 @@
...
+      text-decoration: underline;
+    }
+  }
+
+  &.contents-list-nested,
+  &.contents-list-dashed {
+    li {
+      margin-left: $gutter-half + 5;
+      padding-right: $gutter-half + 5;
+      list-style-type: disc;
```
andysellick: I think this line isn't needed - I don't believe there's any circumstances now where the list items should have a disc, it seems to always get overridden.

app/assets/stylesheets/helpers/_contents-list.scss
```diff
@@ -0,0 +1,91 @@
...
+      }
+    }
+  }
+
+  // When nesting contents lists, the top level items have no
+  // list marker and are styled in bold
+  &.contents-list-nested {
+    .nested-contents-item-h2 {
+      margin-left: 0;
+      list-style-type: none;
```
andysellick: Not sure this needs to be defined here - the main styles for this list already set list-style-type to none.

app/assets/stylesheets/helpers/_contents-list.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  &.contents-list-nested,
+  &.contents-list-dashed {
+    li {
+      margin-left: $gutter-half + 5;
+      padding-right: $gutter-half + 5;
+      list-style-type: disc;
+
+      @supports (content: "—") {
```
andysellick: Is there a browser we support that doesn't support content: "-" ?

app/assets/stylesheets/helpers/_contents-list.scss
```diff
@@ -0,0 +1,91 @@
...
+
+    .contents-number {
+      min-width: 1.5em;
+    }
+
+    .contents-text {
+      $contents-text-padding: 0.3em;
+      padding-left: $contents-text-padding;
+
+      .direction-rtl & {
```
andysellick: Has this been tested with proper RTL content? I don't have an example page so all I can do is apply the class to an English page, but if I do that the dashes in the contents list remain on the left hand side. Might be worth checking.

app/assets/stylesheets/helpers/_contents-list.scss
fofr: https://css-tricks.com/almanac/properties/l/line-height/#article-header-id-0

> Unitless line heights are recommended due to the fact that child elements will inherit the raw number value, rather than the computed value. With this, child elements can compute their line heights based on their computed font size, rather than inheriting an arbitrary value from a parent that is more likely to need overriding.

app/assets/stylesheets/helpers/_contents-list.scss
fofr: It is used by browsers that do not support `@supports (content: "—") {`

app/assets/stylesheets/helpers/_contents-list.scss
fofr: I had reason to keep the `@supports` when I switched away from the hack in April – however I haven't been clear in the commit message and I can't remember why. https://github.com/alphagov/government-frontend/commit/827a4af6a90580846f27edf64591f0754c9cae7f

This code isn't really being changed by this PR, just moved. Might be worth assessing as a separate issue.

app/assets/stylesheets/views/_html-publication.scss
fofr: This should be the same as before (ie as added in #384), classes are just renamed.
Did you compare with live?

test/integration/specialist_document_test.rb
fofr: Quite possibly, however this is a find/replace exercise to keep tests passing, rather than an assessment of every contents list test and its purpose.

app/assets/stylesheets/helpers/_contents-list.scss
fofr: This is the same code as was tested successfully in #395, just moved to a new file.

app/assets/stylesheets/helpers/_contents-list.scss
andysellick: Ah, excellent.

app/assets/stylesheets/views/_html-publication.scss
andysellick: I didn't, I'm just assessing it as it is.

fofr:
@andysellick Happy to approve this?

PR #396, opened 2017-07-03
Use govuk_schemas gem for examples too

tijmenb:
This gem can also fetch the examples. It has a slightly nicer interface which means we don't have to mess around with JSON deserialisation.


nickcolley:
This is epic, thanks 👍

PR #398, opened 2017-07-17
Handle `special_route` document type correctly

gpeng:
Content store now falls through to any matching prefix when a content item is not found. This commit adds support for this returning a 404 when a `special_route` is returned by content store. Currently this causes a 500.

This PR requires https://github.com/alphagov/govuk-content-schemas/pull/639 to be deployed.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,11 +1,14 @@
 require 'gds_api/content_store'
 
 class ContentItemsController < ApplicationController
+  class SpecialRouteReturned < StandardError; end
+
   rescue_from GdsApi::HTTPForbidden, with: :error_403
   rescue_from GdsApi::HTTPNotFound, with: :error_notfound
   rescue_from GdsApi::InvalidUrl, with: :error_notfound
   rescue_from ActionView::MissingTemplate, with: :error_406
   rescue_from ActionController::UnknownFormat, with: :error_406
+  rescue_from SpecialRouteReturned, with: :error_notfound
 
   attr_accessor :content_item
 
@@ -28,7 +31,12 @@ private
 
   def load_content_item
     content_item = content_store.content_item(content_item_path)
+    raise SpecialRouteReturned if special_route?(content_item)
     @content_item = present(content_item)
+  end
+
+  def special_route?(content_item)
+    content_item && content_item['document_type'] == "special_route"
   end
 
   def present(content_item)
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -190,6 +190,18 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_response :not_found
   end
 
+  test "returns 404 if content store falls through to special route" do
+    path = 'government/item-not-here'
+
+    content_item = content_store_has_schema_example('special_route', 'special_route')
+    content_item['base_path'] = '/government'
+
+    content_store_has_item("/#{path}", content_item)
+
+    get :show, params: { path: path }
+    assert_response :not_found
+  end
+
   test "returns 403 for access-limited item" do
     path = 'government/case-studies/super-sekrit-document'
     url = content_store_endpoint + "/content/" + path


app/controllers/content_items_controller.rb
```diff
@@ -1,10 +1,14 @@
...
 
 class ContentItemsController < ApplicationController
+  class SpecialRouteReturned < StandardError; end
+
   rescue_from GdsApi::HTTPForbidden, with: :error_403
   rescue_from GdsApi::HTTPNotFound, with: :error_notfound
   rescue_from GdsApi::InvalidUrl, with: :error_notfound
   rescue_from ActionView::MissingTemplate, with: :error_406
+  rescue_from SpecialRouteReturned, with: :error_notfound
+
```
fofr: Rogue whitespace.

fofr:
What is a special route?

fofr:
I don't see any content items with schema_name "special_route" and rendering app "government-frontend" in the content store.

What circumstances lead government-frontend to render this special_route?

gpeng:
`special_route` are for things like `/government` that publishing apps (Whitehall) in that case, reserve.

I'm not sure of 'all' of the circumstances that lead to government frontend getting these requests but one way to recreate is create a publication with an attachment, save and quickly preview. Probably due to caching you often see the router updated but the content isn't there yet. Instead of 404-ing we get a 500 (or quite a few of them if you have a look at errbit, there are > 200 on the first page in production currently).

app/controllers/content_items_controller.rb
gpeng: Fixed in aecea2d

fofr:
For clarity, we believe this happens when:

* A router receives a URL registration with the government-frontend app
* The content-store, for  an unknown reason, isn't serving the content associated with that content item, instead returning a special_route content item
* government-frontend then can't handle that response

app/controllers/content_items_controller.rb
```diff
@@ -27,9 +30,14 @@ def set_guide_draft_access_token
 
   def load_content_item
     content_item = content_store.content_item(content_item_path)
+    raise SpecialRouteReturned if special_route?(content_item)
     @content_item = present(content_item)
   end
 
+  def special_route?(content_item)
+    content_item && content_item['document_type'] == "special_route"
```
fofr: I wonder if a check that inspects whether the returned content item has the rendering_app set to government-frontend would be useful.

gpeng: Yeah... maybe. I think that would be an 'actual exception' though rather than something we know is happening and need to handle.

kevindew:
Hello :wave: @fofr asked me to provide some details.

So the core of why this happens is that an inconsistency occurs between router and content-store: 

If router thinks `/government/a` is registered to `government-frontend` it will send requests to government-frontend app. 
However if `/government/a` is no longer in the content-store. The content store will return with [/government](https://www.gov.uk/api/content/government) -  This is because we have /government as a prefix route of everything /government - rather than content store returning 404 like one would expect.

At the core of this is probably that /government should not be a prefix route for everything. Perhaps this should be on the [Tech Debt trello board](https://trello.com/b/oPnw6v3r/govuk-tech-debt)

But the question is how does content-store and router slip out of snyc:

- On the draft stack, a draft can be discarded and the route is not removed from content store [trello card](https://trello.com/c/2vqZ6OFC/531-remove-routes-when-deleting-from-content-store)

I'm confused though by examples on the live stack such as https://www.gov.uk/government/consultations/independent-custody-visitor-icv-code-of-practice-consultation and https://www.gov.uk/government/consultations/domestic-violence-disclosure-scheme - as these both seem to have entries in live content store. Not sure if they're rendered by multiple content items?

fofr:
@kevindew For the live versions of this error, we think this is happening for URLs with `?locale=en` at the end, but still don't quite understand why.

gpeng:
@fofr @kevindew I'm going to merge this unless you've any objections. It'll reduce the noise a bit in errbit for now and we can take it back out if we make changes to content store at some point.

PR #402, opened 2017-07-19
[Do not merge] Enable sass linting in Jenkins

fofr:
Jenkinsfile docs suggest that sass linting is on by default:
```
“but aren't linting your SASS yet (you should), you can disable linting”
“sassLint Whether or not to run the SASS linter. Default: true”
```

But the logic says:
`if (hasAssets() && hasLint() && options.sassLint)`

If the option is omitted, it will default to false.

Before this PR, the Sass linter does not run:
https://ci.integration.publishing.service.gov.uk/job/government-frontend/job/display-curated-links-in-sidebar/5/console

You can see it does run now:
https://ci.integration.publishing.service.gov.uk/job/government-frontend/job/enable-sass-linting/1/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,5 +4,5 @@ REPOSITORY = 'government-frontend'
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject()
+  govuk.buildProject(sassLint: true)
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,5 +4,5 @@ REPOSITORY = 'government-frontend'
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject(sassLint: true)
+  govuk.buildProject()
 }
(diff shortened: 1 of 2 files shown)


tijmenb:
Hmm, this seems like a bug in the puppet code, we intended to make it hard to _not_ run the sass-linter.

I think the default value was inadvertently removed in https://github.com/alphagov/govuk-puppet/commit/43af3ededf1fc289d63bbac3969176bed76ce147. We should fix it there.

fofr:
@tijmenb Yep, see: https://github.com/alphagov/govuk-puppet/pull/6131

fofr:
Closing. Puppet change has fixed this: https://github.com/alphagov/govuk-puppet/pull/6131

Reverting the change in a commit, CI runs the sass linter and fails the build as exected:
https://ci.integration.publishing.service.gov.uk/job/government-frontend/job/enable-sass-linting/3/console

```
bundle exec govuk-lint-sass app/assets/stylesheets
app/assets/stylesheets/helpers/_available-languages.scss:1:1 [W] IdSelector: Avoid using id selectors
app/assets/stylesheets/helpers/_available-languages.scss:2:10 [W] ColorVariable: Color literals like `#000` should only be used in variable declarations; they should be referred to via variable everywhere else.
```

PR #403, opened 2017-07-20
Create print link component

andysellick:
- replaces existing print link in guides and travel advice pages
- accepts parameters for href content and link text
- deliberately errors if href not provided to component
- props to Vanita

screen shot 2017-07-20 at 15 52 52

https://trello.com/c/SXdjvsAh/29-convert-print-link-to-component


app/assets/stylesheets/components/_print-link.scss
```diff
@@ -0,0 +1,14 @@
+.app-c-print-link {
+  margin-top: 2em;
+  margin-bottom: 2em;
+
+  a {
```
fofr: Can we avoid styling elements and use a class?

app/assets/stylesheets/components/_print-link.scss
```diff
@@ -0,0 +1,14 @@
...
+    background: image-url("icon-print.png") no-repeat 0 50%;
```
fofr: I wonder how we document when an image is part of a component.

PR #406, opened 2017-07-21
Reduce timeout from 10s to 5s for Webchat availability polling

nickcolley:
Graham Orr from HMRC:
"We have previously requested that all webchat links have their update time reduced from 15 to 10 seconds as an experiment.

The experiment was a success and we would now like the webchat update time to be lowered from 10 seconds down to 5 seconds."

Do not merge until Monday 07/08/17

### DIFF
--- a/app/assets/javascripts/webchat/library.js
+++ b/app/assets/javascripts/webchat/library.js
@@ -7,7 +7,7 @@
 
 
   function Webchat (options) {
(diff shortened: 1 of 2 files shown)


fofr:
Updated PR title to be 10s down to 5s as PR description suggested.

fofr:
Rebased the commit to have the correct timeout.

PR #407, opened 2017-07-21
Create download link component

vanitabarrett:
- Replaces the existing download map link in travel advice pages.
- Takes parameters for href and link text
- Deliberately errors if no href is provided
- Documented in the component guide (shown below):

screen shot 2017-07-21 at 15 32 42
screen shot 2017-07-21 at 15 33 08

### DIFF
--- a/app/assets/stylesheets/components/_download-link.scss
+++ b/app/assets/stylesheets/components/_download-link.scss
@@ -1,0 +1,21 @@
+.app-c-download-link {
+  display: inline-block;
+  @include bold-19;
+  margin-top: $gutter-half;
+  margin-bottom: $gutter-half;
+
+  @include media(tablet) {
+    margin-top: $gutter;
+  }
+}
+
+
+// scss-lint:disable SelectorFormat
+.app-c-download-link__icon {
+  margin-right: 0.5em;
+  height: 1.31578947368em; // 25 / 19
+  width: 1.31578947368em;
+  vertical-align: middle;
+}
+// scss-lint:enable SelectorFormat
+
--- a/app/views/components/_download-link.html.erb
+++ b/app/views/components/_download-link.html.erb
@@ -1,0 +1,17 @@
+<%
+  link_text ||= "Download File"
+%>
+
+<a class="app-c-download-link" href="<%= href %>">
+  <svg class="app-c-download-link__icon"
+        aria-hidden="true"
+        viewBox="0 0 25 25"
+        height="25"
+        width="25"
+        xmlns="http://www.w3.org/2000/svg">
+    <g>
+      <path d="M23 5h2v20H5v-2h18V5z"/>
+      <path d="M0 0h20v20H0z"/>
+    </g>
+  </svg><%= link_text %>
+</a>
--- a/app/views/components/docs/download-link.yml
+++ b/app/views/components/docs/download-link.yml
@@ -1,0 +1,16 @@
+name: Download link
+description: A link with a file download icon
+accessibility_criteria: |
+  The download link must:
+  - accept focus
+  - be focusable with a keyboard
+  - be usable with a keyboard
+  - indicate when it has focus
+  The download icon must:
+  - be presentational and ignored by screen readers
+fixtures:
+  default:
+    href: '/download-this'
+  custom_link_text:
+    href: '/download-map'
+    link_text: 'Download Map (PDF)'
--- a/app/views/shared/_travel_advice_summary.html.erb
+++ b/app/views/shared/_travel_advice_summary.html.erb
@@ -14,8 +14,8 @@
   <figure class="map">
     <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>">
     <% if content_item.map_download_url %>
-      <figcaption class="map-download">
-        <a href="<%= content_item.map_download_url %>">Download map (PDF)</a>
+      <figcaption>
+        <%= render 'components/download-link', href: content_item.map_download_url, link_text: "Download map (PDF)" %>
       </figcaption>
     <% end %>
   </figure>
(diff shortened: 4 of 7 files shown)


app/views/components/_download-link.html.erb
```diff
@@ -0,0 +1,9 @@
+<%
+  download_link_text ||= "Download File"
```
nickcolley: Do we need to extra verbosity here, could we just use 'text' since the we know it's a download link as the component is called download link?

app/assets/stylesheets/components/_download-link.scss
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  // scss-lint:disable SelectorFormat
+  .app-c-download-link__link {
+    display: block;
+    @include bold-19;
+    min-height: 2.5em;
+    padding-left: 2.5em;
+    background: image-url("icon-file-download.png") no-repeat scroll 0 0;
+  // scss-lint:enable SelectorFormat
```
nickcolley: Nit: I think this should be re enabled at 21

app/assets/stylesheets/components/_download-link.scss
```diff
@@ -0,0 +1,22 @@
+.app-c-download-link {
+  margin-top: $gutter-half;
+
+  @include media(tablet) {
+    margin-top: $gutter;
+  }
+
+  // scss-lint:disable SelectorFormat
+  .app-c-download-link__link {
```
nickcolley: If this is block level, do we need the wrapper element?

app/views/components/_download-link.html.erb
fofr: I think the API should match `print_link`

app/views/components/docs/download-link.yml
```diff
@@ -0,0 +1,16 @@
...
+  - be usable with a keyboard
+  - indicate when it has focus
+  The download icon must:
+  - be presentational and ignored by screen readers
+fixtures:
+  default:
+    href: '/download-this'
+  custom_link_text:
+    href: '/download-map'
+    download_link_text: 'Download Map (PDF)'
```
fofr: This file should have a new line at the end. This is something you can configure in your editor.

app/views/shared/_travel_advice_summary.html.erb
```diff
@@ -14,8 +14,8 @@
   <figure class="map">
     <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>">
     <% if content_item.map_download_url %>
-      <figcaption class="map-download">
-        <a href="<%= content_item.map_download_url %>">Download map (PDF)</a>
+      <figcaption>
+        <%= render partial: 'components/download-link', locals: {href: content_item.map_download_url, download_link_text: "Download map (PDF)"} %>
```
fofr: The partial render could be a bit simpler, the guide recommends using:
```erb
<%= render 'components/download-link', href: content_item.map_download_url, download_link_text: "Download map (PDF)" %>
```

app/views/components/_download-link.html.erb
```diff
@@ -0,0 +1,9 @@
...
+%>
+
+<div class="app-c-download-link">
```
fofr: I agree with Nick, I don't think we need the `div` wrapper here.

fofr: (Same could have been said about print link)

andysellick: Is the wrapper needed to ensure clearing is done correctly? I'm not sure, simply asking. We can't make the link block level because the focus state would look weirdly wide.

app/views/shared/_travel_advice_summary.html.erb
andysellick: Should we be translating the download link text?

app/views/components/_download-link.html.erb
vanitabarrett: @andysellick @fofr If we make the link block level, the focus state looks like this:
screen shot 2017-07-24 at 09 40 32

rather than this?
screen shot 2017-07-24 at 09 40 45

fofr: Does the component need to be block level?

app/views/shared/_travel_advice_summary.html.erb
fofr: Probably don't need to add this now.

app/views/components/_download-link.html.erb
nickcolley: When seeing the focus state it looks oddly big, I wonder if that min-height is what's making it bigger than you'd expect?

vanitabarrett: I think we were using the wrapper and/or making it block level so we could add spacing (margin-top) between the map and download map link. I'm not sure whether we want to keep this spacing or if this will be effected by our spacing changes later on?

Removing min-height does make the focus state smaller, but it still takes up a lot of the width of the page - is this expected behaviour?
screen shot 2017-07-24 at 10 20 57

nickcolley: We could probably use `display: inline-block` in this case.

nickcolley:
If we remove the min-height I think we'll need some `margin-bottom` to compensate?

test/components/download_link_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  test "fails to render a download link when no href is given" do
+    assert_raise do
+      render_component({})
+    end
+  end
+
+  test "fails to render a download link when no href is given but custom text is" do
+    assert_raise do
+      render_component(download_link_text: 'link text')
```
nickcolley: Hmm I think we can remove this test since it seems to be redundant since we changed to `link_text` and it didnt fail.

PR #408, opened 2017-07-24
Add SRI again (again)

h-lame:
This is version 3 of #331 and #400.  In both cases we added SRI and then realised that it broke something so had to revert them (#368 and #401 respectively).  Hopefully this version will last a bit longer.

The problem last time was that the BOM removal gem was overzealous and removed 2 extra characters from the CSS file when it removed the BOM.  We fixed this in the gem (see https://github.com/alphagov/asset_bom_removal-rails/pull/6 for the full details) and so this PR repeats #400 and bumps the included gem version.

For: https://trello.com/c/oFhtO0Gm/146-enable-subresource-integrity-sri-on-government-frontend-l

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem 'govuk_ab_testing', '~> 2.0'
 gem 'htmlentities', '4.3.4'
 gem 'statsd-ruby', '1.3.0', require: 'statsd'
 gem 'dalli'
+gem 'asset_bom_removal-rails', '~> 1.0.0'
 
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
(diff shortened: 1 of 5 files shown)


h-lame:
Thanks for looking at the error @fofr - I don't have access to heroku so couldn't check.  Another test vs. production difference in rails autoloading bites.  I've a fix for the gem here: https://github.com/alphagov/asset_bom_removal-rails/issues/8

h-lame:
Rebased to use v1.0.2 of asset_bom_removal-rails which should work in production mode.

fofr:
I've tested the Heroku preview app in Firefox < 52 and it seems to work fine. Assets look good too.
https://government-frontend-pr-408.herokuapp.com/foreign-travel-advice/thailand/safety-and-security

Might be worth running the visual regressions tests to check things haven't inadvertently changed.

h-lame:
Ran the regression tests and failures fall into a bunch of false positive categories:

1. "something went wrong" pages on heroku
2. pages with a beta banner on production that isn't appearing on heroku
3. pages that redirect to other content when requested through production

So I think we're ok to deploy.

PR #410, opened 2017-07-26
Move description above documents

alextea:
When a consultation has many attachments, the description is below the
documents and users may miss important information.

This was brought up in the Snook discovery into Consultations.

## Before
consultations-before

## After
consultations-after


nickcolley:
This is also more consistent with the short text pages 👍 

cc @miaallers

PR #413, opened 2017-07-27
Add notice component

andysellick:
- component replaces notice, consultation notice and withdrawal notices
- component must have a title
- accepts also either a description_text simple string to be wrapped in a para, or description_govspeak, a chunk of HTML to be rendered by govspeak

screen shot 2017-07-27 at 09 24 57

https://trello.com/c/pmyrvkHi/33-1-convert-notice-to-a-component


andysellick:
Have added a test but it's incomplete, setting to do not merge.

test/integration/notice_test.rb
```diff
@@ -0,0 +1,30 @@
+require 'test_helper'
+
+class NoticeGovspeakTest < ActionDispatch::IntegrationTest
```
nickcolley: I think we should put this in a 'components' directory so it's not confused with the other template level integration tests.

test/components/notice_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'component_test_helper'
+
+class NoticeTest < ComponentTestCase
+  def component_name
+    "notice"
+  end
+
+  test "fails to render when no parameters are given" do
```
nickcolley: We need to figure out what do with these tests, ideally we'd only have the integration test and that'd cover all options.

test/integration/notice_test.rb
fofr: Or call it `NoticeComponentGovspeakTest`

andysellick: There's already a notice_test.rb in the test/components/ directory, it contains one test to check that nothing is rendered if no parameters are passed, which the integration test can't check. Could I combine the files? Would that work/be sensible?

app/assets/stylesheets/components/_notice.scss
```diff
@@ -8,7 +8,8 @@
     padding: $gutter-half;
   }
 
-  h2 {
+  // scss-lint:disable SelectorFormat
```
nickcolley: We can get rid of these selectors now, also the nesting is not necessary anymore

test/components/notice_test.rb
andysellick: Not sure that's possible - the integration test couldn't test that the component errors if a required parameter isn't passed.

andysellick: Current solution is that I've got the integration test housed in the component test - so now there's just one test file. It works, and I've commented the file to explain.

nickcolley:
Good to merge when conflicts are resolved 👍

app/assets/stylesheets/components/_notice.scss
```diff
@@ -0,0 +1,28 @@
...
+  clear: both;
+  @include responsive-bottom-margin;
+  padding: $gutter-two-thirds;
+  border: 5px solid $govuk-blue;
+
+  @include media(mobile) {
+    padding: $gutter-half;
+  }
+
+  p:last-child {
```
fofr: As @nickcolley pointed out, this should be explicit that it's styling something within the govspeak component.

PR #416, opened 2017-07-28
Allow component guide to render on Heroku

fofr:
We need the component guide to be available in Heroku preview apps to check components in pull requests and to link to versions of the guide.

* We don't want the component guide to render in production mode
* Heroku review apps have the production environment set
* Make a special case for including the guide when a HEROKU env
  variable has been set
* This allows us to link users to a PR's component guide or master’s

https://government-frontend-pr-416.herokuapp.com/component-guide


boffbowsh:
I think this logic will result in different Gemfile.locks depending on how it's run. I think the logic needs to apply to the `require:` kwarg instead.

fofr:
@boffbowsh Updated.

PR #417, opened 2017-07-28
Create translation navigation component

vanitabarrett:
The translation nav component:
- Replaces the existing 'available languages' partial in government frontend
- Takes parameters for a translation content item. This should contain base_href, locale, text, and an active state if applicable
- Deliberately errors if no translations are provided
- Documented in the component guide (shown below):

screen shot 2017-07-28 at 16 47 02

## **Usage**
The old 'available languages' partial was used in multiple templates in government-frontend. **Examples:**
- [Case Study](https://www.gov.uk/government/case-studies/the-role-of-women-in-afghanistan)
- [Corporate Report](https://www.gov.uk/government/publications/your-charter)
- [Detailed Guide](https://www.gov.uk/guidance/hm-land-registry-registration-services-fees)
- [Publication Scheme](https://www.gov.uk/government/organisations/department-for-work-pensions/about/publication-scheme)
- [Speech](https://www.gov.uk/government/speeches/we-seek-to-align-uk-interests-with-ambitions-of-karnataka)
- [World Location News](https://www.gov.uk/government/world-location-news/changes-to-secure-english-language-test-providers-for-uk-visas)


fofr:
This component can be previewed at:
https://government-frontend-pr-417.herokuapp.com/component-guide/translation-nav

app/views/components/docs/translation-nav.yml
```diff
@@ -0,0 +1,21 @@
+name: Translation Nav
+description: A navigation item for available language translations.
+body: The active property should be used to indicate the currently selected language.
+accessibility_criteria: |
+  The translation nav must:
+  - accept focus
```
fofr: For the markdown to render as a list there needs to be a line break before the first `- item`

fofr:
Can we add a fixture for an extreme example? Something with _lots_ of translations?

fofr:
We need a right to left example:
https://www.gov.uk/government/case-studies/the-role-of-women-in-afghanistan.ur

app/assets/stylesheets/components/_translation-nav.scss
```diff
@@ -1,19 +1,17 @@
-.available-languages {
+// scss-lint:disable SelectorFormat
```
fofr: Shouldn't need this comment now the linter has been updated.

app/assets/stylesheets/components/_translation-nav.scss
```diff
@@ -38,4 +36,5 @@
       }
     }
   }
-}
+} // scss-lint:enable SelectorFormat
+
```
fofr: Stray newline.

app/views/components/_translation-nav.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="app-c-translation-nav">
+  <ul class="app-c-translation-nav__list">
+    <% translations.each.with_index do |translation, i| %>
+      <li class="app-c-translation-nav__list__language<%= ' last' if i == translations.length-1 %>">
+        <% if translation[:active].present? %>
+          <%= translation[:text] %>
+        <% else %>
+          <%= link_to translation[:text], translation[:base_path], lang: translation[:locale] %>
```
fofr: Should the link have the `rel=alternate` attribute too?

app/views/components/_translation-nav.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="app-c-translation-nav">
+  <ul class="app-c-translation-nav__list">
+    <% translations.each.with_index do |translation, i| %>
+      <li class="app-c-translation-nav__list__language<%= ' last' if i == translations.length-1 %>">
```
fofr: Is the `last` class being used?

app/views/components/_translation-nav.html.erb
```diff
@@ -0,0 +1,13 @@
...
+        <% if translation[:active].present? %>
```
fofr: `translation[:active]` should be a boolean, so we can simply do: `if translation[:active]`

app/views/content_items/detailed_guide.html.erb
```diff
@@ -7,9 +7,7 @@
     <%= render 'govuk_component/title', @content_item.title_and_context %>
   </div>
   <div class="column-third">
-    <%= render 'shared/available_languages',
-      translations: @content_item.available_translations
-    %>
+    <%= render 'components/translation-nav', translations: @content_item.available_translations %>
```
fofr: If `available_translations` returned `{ translations: … }` this could be simplified to:
```erb
<%= render 'components/translation-nav', translations: @content_item.available_translations %>
```

This might make it easy to refactor the API in the future.

app/views/components/_translation-nav.html.erb
fofr: If it's not we could also remove the `with_index` part of the loop

app/assets/stylesheets/components/_translation-nav.scss
```diff
@@ -1,19 +1,17 @@
...
 
-  ul {
+  .app-c-translation-nav__list {
     @extend %contain-floats;
     list-style: none;
     margin-left: -$gutter-one-third;
     margin-right: -$gutter-one-third;
 
-    li {
+    .app-c-translation-nav__list__language {
```
fofr: Should the `.last` be a BEM class modifier?

fofr: Or we could use `:last-child` and simplify the logic in the template?

app/views/components/_translation-nav.html.erb
nickcolley: This is not the case since the presenter only adds active when it's needed

app/views/content_items/detailed_guide.html.erb
nickcolley: I think this is more consistent with the other components, left as is, as from a glance you know what data is being passed in.

app/assets/stylesheets/components/_translation-nav.scss
nickcolley: Since we're using BEM, we don't want to nest since it'll create long selectors in the output CSS.

app/views/components/_translation-nav.html.erb
fofr: When it's not there it's falsy.

```
irb(main):001:0> translations = {}
=> {}
irb(main):002:0> translations[:active]
=> nil
irb(main):003:0> translations[:active] = true
=> true
irb(main):004:0> translations[:active]
=> true
irb(main):005:0>
```

vanitabarrett:
@fofr Is it possible to include a fixture for right-to-left? The component adjusts according to a class of 'direction-rtl' or 'direction-ltr' being applied to a parent, at the moment there is no way of setting direction directly on the component itself.

fofr:
@vanitabarrett Good point. We need to write a story so we can provide context to fixtures.

For now can you provide screenshots showing how it renders right to left and a TODO note regarding adding a right to left fixture. I'd prefer this component wasn't blocked on the fixture context story.

test/components/translation_nav_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    assert_select ".app-c-translation-nav__list__language", text: "English"
+    assert_select ".app-c-translation-nav__list__language a[href=\"/en\"]",
+                    false, "An active item should not link to a page translation"
+  end
+
+  test "renders an inactive translation nav item with href, base path and text" do
+    render_component(
+      translations: [
```
nickcolley: If we test with our fixture examples, it'll be more realistic and we can also test the active/not active states.

nickcolley:
@vanitabarrett we try to follow https://github.com/alphagov/styleguides/blob/master/git.md for the git commit message conventions, for example 'Refactoring translation component tests' could be 'Add more realistic test cases based on fixtures' so when someone looks through the commits they'll be able to find what's relevant quickly.

nickcolley:
Links to test:

https://government-frontend-pr-417.herokuapp.com/component-guide/translation-nav
https://government-frontend-pr-417.herokuapp.com/government/world-location-news/changes-to-secure-english-language-test-providers-for-uk-visas

test/components/translation_nav_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  test "renders an active translation nav item with a text description" do
+    render_component(
+      translations: [
+        {
+          locale: 'en',
+          base_path: '/en',
+          text: 'English',
+          active: true
+        },
+        {
```
nickcolley: We need to have assertion for all the translations, otherwise we could break this and the tests would still pass.

test/components/translation_nav_test.rb
```diff
@@ -0,0 +1,82 @@
...
+          locale: 'hi',
+          base_path: '/hi',
+          text: 'हिंदी',
+        }
+      ]
+    )
+
+    assert_select ".app-c-translation-nav__list__language a[href=\"/en\"]",
+                    false, "An active item should not link to a page translation"
```
nickcolley: We should try to do positive assertions, rather than asserting something does not exist since this test does not confirm that there is an active element formatted how we expect, just that it's not a link. For example it could not be there at all and the tests would pass.

test/components/translation_nav_test.rb
```diff
@@ -0,0 +1,82 @@
...
+          locale: 'zh',
+          base_path: '/zh',
+          text: '中文'
+        }
+      ]
+    )
+
+    assert_select ".app-c-translation-nav__list__language a[lang=\"fr\"]", text: "French"
+    assert_select ".app-c-translation-nav__list__language a[href=\"/fr\"]"
+    assert_select ".app-c-translation-nav__list__language a[href=\"/hi\"]",
```
nickcolley: Same as above

test/components/translation_nav_test.rb
```diff
@@ -0,0 +1,90 @@
...
+      ]
+    )
+
+    assert_select ".app-c-translation-nav__list__language a[lang=\"fr\"]", text: "French"
+    assert_select ".app-c-translation-nav__list__language a[href=\"/fr\"]"
+
+    assert_select ".app-c-translation-nav__list__language a[lang=\"zh\"]", text: "中文"
+    assert_select ".app-c-translation-nav__list__language a[href=\"/zh\"]"
+
+    assert_select ".app-c-translation-nav__list__language", text: "हिंदी"
```
nickcolley: Does this ensure the active state does not use an anchor?

I think we need to test both that it exists but that it's not an anchor.

vanitabarrett: @nickcolley What's the best way of testing it's not an anchor using a positive assertion?

app/views/content_items/detailed_guide.html.erb
fofr: Take a look at how government-frontend handles the title, metadata and document_footer components.

app/views/shared/_translations.html.erb
```diff
@@ -1,7 +1,6 @@
-<% if @content_item.available_translations.any? %>
+<% if @content_item.available_translations.any? && @content_item.available_translations.length > 1 %>
```
fofr: Why do we need both `@content_item.available_translations.any?` and `@content_item.available_translations.length > 1`?

test/components/translation_nav_test.rb
nickcolley: We spoke offline in person and decided to add an addition assertion to cover both cases here.

app/views/shared/_translations.html.erb
vanitabarrett: I think because we don't want to render a translation navigation item for only one translation.. there should always be a link to the current language the page is being displayed in?

fofr:
Right to left seems to work as expected:
screen shot 2017-08-03 at 08 18 36

PR #419, opened 2017-07-31
Add heading component

andysellick:
- requires text to be passed or errors
- accepts optional id parameter
- accepts optional heading value (1 to 6) to change heading level

screen shot 2017-07-31 at 16 23 23

https://trello.com/c/yV8MYIvK/41-2-convert-heading-to-component


fofr:
I think this component needs a bit of context about when and where it should be used.
"Heading" is a little vague.

app/views/components/_heading.html.erb
```diff
@@ -0,0 +1,11 @@
+<%
+  id = "id=#{id}" if id
+  heading_level ||= 1
```
nickcolley: Should we default to 2 since we can make the assumption that we'll only want one H1 and the title component will do that most of the time?

andysellick:
I've added a bit more info into the yml file, hopefully should be clearer.

PR #421, opened 2017-08-01
Add guidance on using components

fofr:
* Indicate where components live
* Indicate which component guides to use


nickcolley:
🚢

PR #424, opened 2017-08-02
Add banner component

vanitabarrett:
https://government-frontend-pr-424.herokuapp.com/component-guide/banner

The banner component:
* Replaces the existing history-notice and consultation-banner.
* Takes parameters for title, text and aside.
* Providing the banner with an aside will change the layout to use the grid
* Deliberately errors if no text is provided. Other parameters are optional.
* Documented in the component guide (shown below):

screen shot 2017-08-02 at 11 54 14
screen shot 2017-08-02 at 11 54 24
screen shot 2017-08-02 at 11 54 37


nickcolley:
Seems to be some odd behaviour with the padding on the grid:

screen shot 2017-08-02 at 13 32 07

fofr:
Spacing looks a little odd when there's just a title and text:
screen shot 2017-08-02 at 13 35 32
screen shot 2017-08-02 at 13 36 21

app/assets/stylesheets/views/_consultation.scss
```diff
@@ -55,3 +15,7 @@
     }
   }
 }
+
+.consultation-date {
```
fofr: As this isn't a BEM class, it should probably still live within the `.consultation` scope.

app/views/components/_banner.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <div class="column-two-thirds app-c-banner__main">
+      <h2 class="app-c-banner__title"><%= title %></h2>
+      <p class="app-c-banner__desc"><%= text  %></p>
+    </div>
+  </div>
+ </div>
+<% else %>
+  <div class="app-c-banner">
+    <% if title %>
+      <h2 class="app-c-banner__title"><%= title %></h2>
```
fofr: You could probably have a block at the top for title and description to avoid repetition:

```erb
<% title_block = capture do %>
  <h2 class="app-c-banner__title"><%= title %></h2>
<% end %>
```

Then
```erb
<div class="column-two-thirds app-c-banner__main">
  <%= title_block %>
  <p class="app-c-banner__desc"><%= text  %></p>
</div>
```

app/views/components/_banner.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<% if aside %>
+ <div class="app-c-banner app-c-banner--grid">
+  <div class="grid-row app-c-banner__aside">
+    <div class="column-third">
+      <p><%= aside %></p>
+    </div>
+
+    <div class="column-two-thirds app-c-banner__main">
+      <h2 class="app-c-banner__title"><%= title %></h2>
```
fofr: If `aside` is provided but without a `title`, will this error?

app/views/components/docs/banner.yml
```diff
@@ -0,0 +1,18 @@
+name: Banner
+description: A page banner, designed to highlight important information
+body: Passing an aside to the banner will mean the banner renders in a grid layout.
+accessibility_criteria: |
+  The banner must:
+
+  - be visually distinct from other content on the page
+  - meet WCAG AA standard for colour contrast
```
fofr: It'd be good to include what the contrast ratio should be.

app/views/content_items/consultation.html.erb
```diff
@@ -83,41 +83,28 @@
...
-    </div>
-  </div>
-</div>
+<% consultation_desc = capture do %>
+  <%= @content_item.description %>
+  <% if @content_item.held_on_another_website? %>
+    <strong>This consultation <% if @content_item.closed? %>was<% else %>is being<% end %> held on <a href="<%= @content_item.held_on_another_website_url %>">another website</a>.</strong>
+  <% end %>
+<% end %>
+<%= render 'components/banner', text: consultation_desc, title: 'Summary', aside: consultation_date %>
```
fofr: Could we have a fixture that shows a consultation banner with some equivalent HTML passed through?

app/views/shared/_history_notice.html.erb
```diff
@@ -1,5 +1,4 @@
-<% if content_item.historically_political? %>
-  <div class="history-notice">
-    <h2>This was published under the <%= content_item.publishing_government %></h2>
-  </div>
+<% if content_item.historically_political?
+    text = 'This was published under the ' + content_item.publishing_government %>
+  <%= render 'components/banner', text: text %>
```
fofr: I think this would be a little cleaner with [ruby string interpolation](https://stackoverflow.com/questions/10076579/string-concatenation-vs-interpolation-in-ruby):

```erb
<%= render 'components/banner', text: "This was published under the #{content_item.publishing_government}" %>
```

app/views/components/_banner.html.erb
vanitabarrett: @fofr At the moment, no, the only argument that is required is 'text'. I was trying to make the banner as general as possible, and couldn't see a reason why providing an aside and just text (with no title) would be a problem?

app/views/content_items/consultation.html.erb
vanitabarrett: @fofr Wasn't sure how to do this in the .yml file?

app/views/components/_banner.html.erb
```diff
@@ -0,0 +1,33 @@
...
+<% content_block = capture do %>
+  <% if title %>
+    <h2 class="app-c-banner__title"><%= title %></h2>
+    <p class="app-c-banner__desc"><%= text  %></p>
+  <% elsif aside %>
+    <p class="app-c-banner__desc"><%= text  %></p>
+  <% else %>
+    <p class="app-c-banner__text"><%= text  %></p>
+  <% end %>
+<% end %>
```
fofr: You could simplify this further to something like:

```erb
<%
  title ||= false
  aside ||= false
  large_text = !title && !aside
%>

<% content_block = capture do %>
  <% if title %>
    <h2 class="app-c-banner__title"><%= title %></h2>
  <% end %>
  <p class="app-c-banner__<% if large_text %>text<% else %>desc<% end %>">
    <%= text %>
  </p>
<% end %>
```

app/views/components/docs/banner.yml
```diff
@@ -0,0 +1,18 @@
+name: Banner
+description: A page banner, designed to highlight important information
+body: Passing an aside to the banner will mean the banner renders in a grid layout.
+accessibility_criteria: |
+  The banner must:
+
+  - be visually distinct from other content on the page
+  - meet WCAG AA standard for colour contrast (4.5:1 for normal text and 3:1 for large text)
```
fofr: We've had trouble in the past interpreting what normal text and large text are. I think we should be explicit about one ratio for all text:

* all text must have a contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA

For context, see the discussion here: https://github.com/alphagov/govuk_frontend_toolkit/issues/316

app/views/content_items/consultation.html.erb
fofr: HTML can be passed in within a string. Something like:

```yaml
aside: "This consultation ran from<br><span class=\"consultation-date\"><time>9:30am on 13 June 2017</time> to <time>11:45pm on 11 July 2017</time>"
```

vanitabarrett: I tried this before and it just prints out the HTML tags along with the text? Does using `capture do` in Ruby behave differently, because that definitely works when passing through consultation dates to the banner component?

vanitabarrett: Created story for this here: https://trello.com/c/3ov9FAlO/144-gem-allow-fixtures-containing-html

PR #431, opened 2017-08-07
Don't render translations when only one available

fofr:
Fix issue where “English” is showing by itself on all English detailed guides.

Update translation nav component to only output navigation if there’s
more than one available translation. If there's only one then it's
the current page and the text is redundant.

Fixes:
screen shot 2017-08-07 at 17 16 42


fofr:
Re-opening with fix to component instead

PR #434, opened 2017-08-09
Move the breadcrumb outside of the main element

andysellick:
- skip to content link now bypasses the breadcrumb
- change made to keep us consistent following the same change to Elements

https://trello.com/c/IsxWIQXQ/21-2-investigate-work-needed-to-have-breadcrumbs-consistent-with-govuk-elements

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -27,9 +27,9 @@
   <% end %>
 
   <div id="wrapper" class="<%= wrapper_class %>">
+    <%= render_phase_label @content_item, content_for(:phase_message) %>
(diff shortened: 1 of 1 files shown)


fofr:
For reference:
https://github.com/alphagov/govuk_elements/pull/478

fofr:
Can you confirm that this has no effect on the rendering of the pages, and that the AB tested breadcrumb which is slightly different also looks the same as it does on live.

andysellick:
I've tested every example page listed on https://github.com/alphagov/government-frontend with this change, including several with both the A and B variants (although I didn't notice a difference in the breadcrumb on those). Also tested with the JS appended banners (e.g. email survey banner). All looked ok.

PR #436, opened 2017-08-10
Create share links component

vanitabarrett:
The share links component:
* Replaces the existing share buttons
* Takes parameters for title and share urls
* Deliberately errors if no share urls are provided
* Is documented in the component guide (shown below)

[Example in News Article](https://government-frontend-pr-436.herokuapp.com/government/news/fast-tracking-uk-innovation-apply-for-business-funding)
[Example in Consultation page](https://government-frontend-pr-436.herokuapp.com/government/consultations/soft-drinks-industry-levy)

screen shot 2017-08-10 at 15 58 50
screen shot 2017-08-10 at 15 58 56


fofr:
Can you also include links/examples of its use in context on page?

fofr:
Images don't seem to be working on the review app:
https://government-frontend-pr-436.herokuapp.com/component-guide/share-links

app/views/components/_share-links.html.erb
```diff
@@ -0,0 +1,18 @@
+<% title ||= 'Share this page' %>
+
+<div class="app-c-share-links-wrapper">
+  <h2><%= title %></h2>
+  <div class="app-c-share-links" data-module="track-share-button-clicks">
+    <%= link_to share_urls[:facebook],
```
fofr: I think I'd prefer `facebook_href` to `share_urls[:facebook]` – ie an API similar to our other link components

test/components/share_links_test.rb
```diff
@@ -0,0 +1,29 @@
+require 'component_test_helper'
+
+class ShareLinksTest < ComponentTestCase
+  def component_name
+    "share-links"
+  end
+
+  test "fails to render share links when no share_urls are given" do
```
fofr: Do we need to better handle the case where only one share link is provided? eg Facebook only?

app/views/components/_share-links.html.erb
vanitabarrett: @fofr I kept this the same as the old share buttons which seem to generate the social media URLs from the current page and pass them into the partial/component via the content item?

fofr: I think this:
https://github.com/alphagov/government-frontend/blob/aaib-dates/app/presenters/shareable.rb#L4-L9

Could become:
```ruby
def share_links
    {
      title: 'Share this page',
      facebook_href: facebook_share_url,
      twitter_href: twitter_share_url
    }
  end
```

Like this:
https://github.com/alphagov/government-frontend/blob/master/app/presenters/metadata.rb#L5

vanitabarrett: Would you then suggest passing them in separately to the component, e.g (ignore syntax mistakes, haven't tried it yet):
`<%= render 'components/share-links', facebook_href: @content_item.share_links[facebook_href] twitter_href:.....etc %>`

instead of pulling out the links from the share links object inside the component itself?

fofr:
```
<%= render 'components/share-links', @content_item.share_links %>
```

test/components/share_links_test.rb
fofr: Test description should be "renders nothing when no share links provided"

test/components/share_links_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    assert_select ".app-c-share-links .app-c-share-links__link[href=\"/facebook\"]"
+    assert_select ".app-c-share-links .app-c-share-links__link[href=\"/twitter\"]"
+  end
+
+  test "renders a default title if no custom title is provided" do
+    render_component(facebook_href: '/facebook', twitter_href: '/twitter')
+    assert_select ".app-c-share-links-wrapper h2", text: "Share this page"
+  end
+
+  test "renders a print link with custom link text correctly" do
```
fofr: Copy and paste error, these aren't print links.

app/views/content_items/news_article.html.erb
```diff
@@ -6,5 +6,12 @@
 <%= render 'shared/history_notice', content_item: @content_item %>
 <%= render 'components/lead-paragraph', text: @content_item.description %>
 <%= render 'shared/sidebar_image_with_body', content_item: @content_item %>
-<%= render 'shared/share_buttons', share_urls: @content_item.share_urls %>
+
+<div class="grid-row sidebar-with-body dont-print">
```
fofr: The `dont-print` class should be in the component.

app/views/shared/_share_buttons.html.erb
```diff
@@ -1,18 +0,0 @@
-<div class="grid-row sidebar-with-body dont-print">
```
fofr: We could keep this partial to avoid repetition of the grid classes in the content items. Maybe rename to something like `share_links_offset_in_grid` and then include the component call?

app/views/components/_share-links.html.erb
```diff
@@ -0,0 +1,28 @@
+<%
+  title ||= 'Share this page'
+  facebook_href ||= false
+  twitter_href ||= false
+%>
+<% if facebook_href || twitter_href %>
+<div class="app-c-share-links-wrapper">
```
fofr: Can the outer class be `app-c-share-links` for consistency with other components?

app/views/components/_share-links.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  <h2><%= title %></h2>
+
+  <div class="app-c-share-links" data-module="track-share-button-clicks">
```
fofr: We could probably refactor out this container by:
* moving data module to parent
* applying margin bottom to the title
* applying font size to the links

app/views/components/docs/share-links.yml
```diff
@@ -0,0 +1,23 @@
+name: Share links
+description: Social media links, linking to facebook and twitter share pages
```
andysellick: I think it would be worth adding a body here to explain a few things:

- the tracking
- the js-share-facebook/twitter classes on the elements (presumably linked to the tracking?)
- that the URL passed must be a complete share URL, i.e. the component does not automatically take a URL and process it into a link that when clicked will share something on Facebook

andysellick: Suggestion for alternative description: links to allow the user to share the current page on Facebook or Twitter

test/components/share_links_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    render_component(facebook_href: '/facebook', twitter_href: '/twitter', title: 'Share this article')
+    assert_select ".app-c-share-links-wrapper h2", text: "Share this article"
+  end
+end
```
andysellick: Do we need a test for if only one share link provided?

app/assets/stylesheets/components/_share-links.scss
```diff
@@ -0,0 +1,60 @@
...
+  display: inline-block;
+  width: $share-button-width;
+  height: $share-button-height;
+  margin-right: $gutter-one-third;
+  background-color: $link-colour;
+  background-image: image-url("social-icons-white.png");
+  background-repeat: no-repeat;
+  vertical-align: top;
+
+  &--facebook {
```
fofr: I preferred @vanitabarrett’s original. I think it's more readable.

fofr:
I'm seeing some weird underlines in the links on Chrome:
screen shot 2017-08-14 at 22 06 47

fofr:
Please add links to the component being used in context, eg:
* consultation with share links
* news article
* a right to left example as the component guide can't include that for now

app/views/components/_share-links.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <% if facebook_href %>
+      <%= link_to facebook_href,
+        target: "_blank",
+        rel: "noopener noreferrer",
+        class: "app-c-share-links__link js-share-facebook" do %>
+        <span class="app-c-share-links__icon app-c-share-links__icon--facebook"></span><span class="visually-hidden">Share on </span>Facebook<% end %>
```
vanitabarrett: Turns out this is on one line to prevent whitespace problems - otherwise you end up with the problem @fofr noticed below, with slightly longer underlines on the text.

andysellick: Oh wow, I almost created a bug via a PR comment. Is there really no solution to this, or is this just one of those annoying browser things? What if you float the icon?

fofr: I wouldn't change implementation to satisfy a line length.

app/views/components/docs/share-links.yml
```diff
@@ -0,0 +1,34 @@
+name: Share links
+description: Links to allow the user to share the current page on Facebook or Twitter
+body: |
+  Pass complete share URls to the component - the component will not process a URL into a share link itself.
```
fofr: Minor: URls should be URLs. Maybe drop the hyphen and start a new sentence too.

app/views/components/docs/share-links.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  The component also  contains Javascript classes (js-share-facebook and js-share-twitter) which are used to track the number of clicks on the share links.
```
fofr: I'd rephrase this slightly:
The component will track interactions with the share links to Google Analytics using [Social Interactions](https://developers.google.com/analytics/devguides/collection/analyticsjs/social-interactions)

PR #439, opened 2017-08-11
Turn up slimmer to 11, update nokogiri to 1.8 and minor Rails bump

fofr:
* Bump slimmer to v11 to pickup latest nokogiri
* Requires a bump to wraith major version
* Requires minor bump to rails-dom-testing
* Bumping only rails-dom-testing and minitest surfaced a bug in Rails: rails/rails#29022
  * Bump rails patch version to address bug fix
* Fix assert_nil deprecation warnings in updated minitest


fofr:
Tests are failing because of https://github.com/searls/jasmine-rails/issues/213

PR #444, opened 2017-08-14
Create document list component

andysellick:
- accepts a content item and outputs a list of lists and information from it
- accepts an optional parameter to add tracking to the links
- possible point of contention: for the date object had to add a check/convert to date object part in order to get component guide to render. Open to better solution if one can be found.

(was called collection list component until part way through development when it became clear the name wasn't clear enough)

https://trello.com/c/JlNtwQ5G/75-2-convert-collection-list-to-document-list-component

screen shot 2017-08-14 at 08 59 19

### DIFF
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -14,7 +14,7 @@ class DocumentCollectionPresenter < ContentItemPresenter
 
   def groups
     groups = content_item["details"]["collection_groups"].reject { |group|
-      group_document_links(group).empty?
+      group_documents(group).empty?
     }
 
     groups.map { |group|
@@ -23,13 +23,26 @@ class DocumentCollectionPresenter < ContentItemPresenter
     }
   end
 
-  def group_document_links(group)
-    group_documents(group).map do |link|
+  def group_document_links(group, group_index)
+    group_documents(group).each_with_index.map do |link, link_index|
       {
-        public_updated_at: Time.zone.parse(link["public_updated_at"]),
-        document_type: link["document_type"],
-        title: link["title"],
-        base_path: link["base_path"]
+        link: {
+          text: link["title"],
+          path: link["base_path"],
+          data_attributes: {
+            track_category: 'navDocumentCollectionLinkClicked',
+            track_action: "#{group_index + 1}.#{link_index + 1}",
+            track_label: link["base_path"],
+            track_options: {
+              dimension28: group['documents'].count.to_s,
+              dimension29: link["title"]
+            }
+          }
+        },
+        metadata: {
+          public_updated_at: Time.zone.parse(link["public_updated_at"]),
+          document_type: link["document_type"],
+        },
       }
     end
   end
--- a/app/views/components/_document-list.html.erb
+++ b/app/views/components/_document-list.html.erb
@@ -1,0 +1,30 @@
+<%
+  items ||= []
+%>
+<% if items.any? %>
+  <ol class="app-c-document-list">
+    <% items.each do |item| %>
+      <li class="app-c-document-list__item">
+        <h3 class="app-c-document-list__item-title">
+          <%=
+            link_to(
+              item[:link][:text],
+              item[:link][:path],
+              data: item[:link][:data_attributes]
+            )
+          %>
+        </h3>
+        <ul>
+          <li class="app-c-document-list__attribute">
+            <time datetime="<%= item[:metadata][:public_updated_at].iso8601 %>">
+              <%= l(item[:metadata][:public_updated_at], format: :short_ordinal) %>
+            </time>
+          </li>
+          <li class="app-c-document-list__attribute">
+            <%= t("content_item.schema_name.#{item[:metadata][:document_type]}", count: 1) %>
+          </li>
+        </ul>
+      </li>
+    <% end %>
+  </ol>
+<% end %>
--- a/app/views/components/docs/document-list.yml
+++ b/app/views/components/docs/document-list.yml
@@ -1,0 +1,78 @@
+name: Document list
+description: An ordered list of links to documents including document type and when updated.
+body: |
+  Outputs a list of links to documents, based on an array of document data. This must include:
+
+  * link text
+  * link href
+  * last updated date object
+  * document type
+
+  Tracking can be added to the links by supplying optional data attributes for each.
+
+  Documents are presented in an ordered list as the component expects that the ordering of the documents is relevant.
+accessibility_criteria: |
+  The component must:
+
+  * inform the user how many items are in the list
+
+  Links in the component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * indicate when it has focus
+  * change in appearance when touched (in the touch-down state)
+  * change in appearance when hovered
+  * be usable with touch
+  * be usable with speech
+  * have visible text
+fixtures:
+  default:
+    items:
+    - link:
+        text: 'Alternative provision'
+        path: '/government/publications/alternative-provision'
+      metadata:
+        public_updated_at: 2016-06-27 10:29:44
+        document_type: 'statutory_guidance'
+    - link:
+        text: 'Behaviour and discipline in schools: guide for governing bodies'
+        path: '/government/publications/behaviour-and-discipline-in-schools-guidance-for-governing-bodies'
+      metadata:
+        public_updated_at: 2015-09-24 16:42:48
+        document_type: 'statutory_guidance'
+    - link:
+        text: 'Children missing education'
+        path: '/government/publications/children-missing-education'
+      metadata:
+        public_updated_at: 2016-09-05 16:48:27
+        document_type: 'statutory_guidance'
+  with_data_attributes_on_links:
+    items:
+    - link:
+        text: 'School behaviour and attendance: parental responsibility measures'
+        path: '/government/publications/parental-responsibility-measures-for-behaviour-and-attendance'
+        data_attributes:
+          track_category: 'navDocumentCollectionLinkClicked'
+          track_action: 1.1
+          track_label: '/government/publications/parental-responsibility-measures-for-behaviour-and-attendance'
+          track_options:
+            dimension28: 2
+            dimension29: 'School behaviour and attendance: parental responsibility measures'
+      metadata:
+        public_updated_at: 2017-01-05 14:50:33
+        document_type: 'statutory_guidance'
+    - link:
+        text: 'School exclusion'
+        path: '/government/publications/school-exclusion'
+        data_attributes:
+          track_category: 'navDocumentCollectionLinkClicked'
+          track_action: 1.2
+          track_label: '/government/publications/school-exclusion'
+          track_options:
+            dimension28: 2
+            dimension29: 'School exclusion'
+      metadata:
+        public_updated_at: 2017-07-19 15:01:48
+        document_type: 'statutory_guidance'
--- a/app/views/content_items/_document_collection_body.html.erb
+++ b/app/views/content_items/_document_collection_body.html.erb
@@ -9,35 +9,8 @@
         content: group["body"],
         direction: page_text_direction %>
   <% end %>
-  <ol class="group-document-list" data-module="track-click">
-    <% @content_item.group_document_links(group).each_with_index do |link, link_index| %>
-      <li class="group-document-list-item">
-        <h3 class="group-document-list-item-title">
-          <%=
-            link_to(
-              link[:title],
-              link[:base_path],
-              data: {
-                track_category: 'navDocumentCollectionLinkClicked',
-                track_action: "#{group_index + 1}.#{link_index + 1}",
-                track_label: link[:base_path],
-                track_options: {
-                  dimension28: @content_item.group_document_links(group).count.to_s,
-                  dimension29: link[:title]
-                }
-              }
-            )
-          %>
-        </h3>
-        <ul class="group-document-list-item-attributes">
-          <li>
-            <time datetime="<%= link[:public_updated_at].iso8601 %>">
-              <%= l(link[:public_updated_at], format: :short_ordinal) %>
-            </time>
-          </li>
-          <li><%= t("content_item.schema_name.#{link[:document_type]}", count: 1) %></li>
-        </ul>
-      </li>
-    <% end %>
-  </ol>
+
+  <div data-module="track-click">
+    <%= render 'components/document-list', items: @content_item.group_document_links(group, group_index) %>
+  </div>
 <% end %>
--- a/test/components/document_list_test.rb
+++ b/test/components/document_list_test.rb
@@ -1,0 +1,109 @@
+require 'component_test_helper'
+
+class DocumentListTest < ComponentTestCase
+  def component_name
+    "document-list"
+  end
+
+  test "renders nothing when no parameters are given" do
+    assert_empty render_component({})
+  end
+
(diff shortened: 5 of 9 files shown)


app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
+<%
+  group_index ||= false
+  track_clicks = "data-module=track-click" if group_index
+  dimension28 = items.count.to_s
+%>
+<ol class="app-c-document-list" <%= track_clicks %>>
+  <% items.each_with_index do |link, link_index| %>
+    <li class="app-c-document-list__item">
+      <h3 class="app-c-document-list__item-title">
```
vanitabarrett: This should probably be app-c-document-list__item__title to follow BEM, as the title is part of the item list.

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
+<%
+  group_index ||= false
+  track_clicks = "data-module=track-click" if group_index
+  dimension28 = items.count.to_s
+%>
+<ol class="app-c-document-list" <%= track_clicks %>>
```
vanitabarrett: Should this use `ul`, not `ol`? (not sure how these items are ordered)

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        <%=
+          link_to(
+            link[:title],
+            link[:base_path],
+            data: data
+          )
+        %>
+      </h3>
+      <ul>
+        <li class="app-c-document-list__attribute">
```
vanitabarrett: To me, it seems weird that the list attributes are part of an unordered list as it doesn't really feel like a list, as it's more a set of specific related attributes/data?

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,43 @@
+name: Document list
+description: A list of links to documents including document type and when updated.
+body: |
+  The component expects data in a specific format from a content item.
```
vanitabarrett: This is a bit vague - are you referring to the date format here? If so, that probably needs to be made clearer.

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,43 @@
...
+  The component must:
+
+  * inform the user how many items are in the list
+
+  Links in the component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * indicate when it has focus
```
vanitabarrett: See Leonie's suggestions for additional criteria for contents list component - may also apply here:

- Change in appearance when touched (in the touch-down state)
- Change in appearance when hovered
- Usable with touch
- Usable with speech
- Have visible text
- The lists must convey the content structure.

app/views/components/_document-list.html.erb
andysellick: It should be an OL, yes. Documents are provided from the content store in alphabetical order, which means pairs of documents occur in the right order (see the 'Assessment' section of https://www.gov.uk/government/collections/statutory-guidance-schools)

app/views/components/_document-list.html.erb
andysellick: It's a series of attributes relating to the document, so an unordered list seems like the correct approach.

The class might be app-c-document-list__item-attribute but I think that's getting unnecessarily verbose.

app/views/components/docs/document-list.yml
andysellick: I think we need to have a wider discussion about some of Leonie's suggestions. There's a danger we may end up over-specifying these criteria. Most of these are about links, so we could just define criteria for links and say they apply to all links unless otherwise specified.

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,43 @@
+name: Document list
+description: A list of links to documents including document type and when updated.
+body: |
+  The component expects data in a specific format from a content item.
+
+  If a group_index parameter is included the component adds link tracking attributes to each link.
+acceptance_criteria: |
```
vanitabarrett: Should be 'accessibility_criteria'

app/views/components/docs/document-list.yml
nickcolley: For now we can duplicate where necessary and we make a card to deduplicate them when patterns emerge.

app/views/components/_document-list.html.erb
fofr: I think the original is right. The _element_ is "item-title".

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  dimension28 = items.count.to_s
+%>
+<ol class="app-c-document-list" <%= track_clicks %>>
+  <% items.each_with_index do |link, link_index| %>
+    <li class="app-c-document-list__item">
+      <h3 class="app-c-document-list__item-title">
+        <%
+          data = {}
+          if group_index
+            data = {
```
fofr: I don't like this being included. It's very specific for a component.

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            link[:base_path],
+            data: data
+          )
+        %>
+      </h3>
+      <ul>
+        <li class="app-c-document-list__attribute">
+          <%
+            document_updated_at = link[:public_updated_at]
+            if not document_updated_at.is_a?(ActiveSupport::TimeWithZone)
```
fofr: `unless document_updated_at.is_a?(ActiveSupport::TimeWithZone)`

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
...
+              document_updated_at = ActiveSupport::TimeZone['UTC'].parse(document_updated_at)
+            end
+          %>
+          <time datetime="<%= document_updated_at.iso8601 %>">
```
fofr: I feel like this time element could be its own component.

app/views/components/_document-list.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            data: data
+          )
+        %>
+      </h3>
+      <ul>
+        <li class="app-c-document-list__attribute">
+          <%
+            document_updated_at = link[:public_updated_at]
+            if not document_updated_at.is_a?(ActiveSupport::TimeWithZone)
+              document_updated_at = ActiveSupport::TimeZone['UTC'].parse(document_updated_at)
```
fofr: Have you double checked this does all the right things re: time zones and day light savings? Not sure this is correct.

app/views/components/_document-list.html.erb
andysellick: @nickcolley my understanding is that a classname in BEM should have at most one instance of -- and one instance of __ i.e. block__element--modifier. Here app-c-document-list is the block, __item and __item-title are the elements. The BEM documentation isn't super clear but I think this link explains it: http://getbem.com/faq/#what-would-be-a-class-name-for-an-element-inside-another-element-block__el1__el2-

app/views/components/_document-list.html.erb
andysellick: Have added some extra information into the yml file.

app/views/components/docs/document-list.yml
andysellick: I'll add all but the last one - it seems to be relevant to the contents list, where there may be nested lists, but not relevant here, where there's only ever one list.

app/views/components/_document-list.html.erb
andysellick: @fofr what would you suggest?

app/views/components/_document-list.html.erb
andysellick: I wondered about this - whether to go with GMT and try to switch to BST or something complex. It seemed like a rabbit hole and then I realised that only the date is used, not the time.

app/views/components/_document-list.html.erb
andysellick: @fofr are you saying we should do this now or are you just making a point?

andysellick:
Moved time logic into a helper.

app/views/components/_document-list.html.erb
vanitabarrett: I've seen something similar (which I think the link above is implying) but a bit less restrictive - more that if an element inside a block can be moved around freely without relying on any parent CSS, then it doesn't need to be nested.

app/views/components/_document-list.html.erb
fofr: Double check this tracking is still needed?
Some sort of `track-group-link-clicks` JS module on the containing group?

app/helpers/document_list_helper.rb
```diff
@@ -0,0 +1,7 @@
+module DocumentListHelper
+  def convert_string_to_date_object(updated)
+    unless updated.is_a?(ActiveSupport::TimeWithZone)
+      updated = ActiveSupport::TimeZone['UTC'].parse(updated)
```
fofr: I think this should be `Time.zone.parse. This will use the app’s timezone rather than UTC (eg London): https://github.com/alphagov/government-frontend/blame/master/app/presenters/updatable.rb#L36

More context in this commit:
https://github.com/alphagov/government-frontend/commit/737211f87e9a9dbf4bcbb3b518a91fec317df50a

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,50 @@
...
+  * indicate when it has focus
+  * Change in appearance when touched (in the touch-down state)
+  * Change in appearance when hovered
+  * Usable with touch
+  * Usable with speech
+  * Have visible text
+fixtures:
+  default:
+    items:
+    - public_updated_at: 'Mon, 27 Jun 2016 10:29:44 BST +01:00'
```
fofr: If you remove the quotes Rails should make this a date automagically:
https://stackoverflow.com/questions/2656196/ruby-how-do-i-define-a-datetime-field-in-yaml

app/views/components/_document-list.html.erb
andysellick: I've checked with Paul C and apparently the tracking is still needed but stopped working on the 9th of August. Will work with him to determine a cause then look at an alternative implementation.

app/views/content_items/_document_collection_body.html.erb
```diff
@@ -9,35 +9,28 @@
...
-      </li>
-    <% end %>
-  </ol>
+
+  <div data-module="track-click">
+    <%
+      items = Array.new
+      @content_item.group_document_links(group).each_with_index do |link, link_index|
+        items[link_index] = {
+          text: link[:title],
```
fofr:
```ruby
items = []
@content_item.group_document_links(group).each_with_index do |link, link_index|
  items << {
```

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,63 @@
...
+      public_updated_at: 2017-01-05 14:50:33
+      document_type: 'statutory_guidance'
+      data_attributes:
+        track_category: 'navDocumentCollectionLinkClicked'
+        track_action: 1.1
+        track_label: '/government/publications/parental-responsibility-measures-for-behaviour-and-attendance'
+        track_options:
+          dimension28: 2
+          dimension29: 'School behaviour and attendance: parental responsibility measures'
+    - text: 'School exclusion'
```
fofr: Spoke about a different structure offline:

```yaml
items:
- link:
    path:
    text:
    data_attributes:
  metadata:
    public_updated_at:
    document_type:
```

app/views/content_items/_document_collection_body.html.erb
fofr: Could move this into the collections presenter

andysellick: There's already a method named group_document_links, could be confusing? How about create_document_links_list?

test/components/document_list_test.rb
```diff
@@ -0,0 +1,117 @@
+require 'component_test_helper'
+
+class DocumentListTest < ComponentTestCase
+  def component_name
+    "document-list"
+  end
+
+  test "fails to render a document list when no parameters are given" do
```
fofr: This is not a failure. "renders nothing when no parameters are given"

test/components/document_list_test.rb
```diff
@@ -0,0 +1,117 @@
...
+    li = ".app-c-document-list__item-title"
+
+    assert_select "#{li} a[href='/government/publications/parental-responsibility-measures-for-behaviour-and-attendance']", text: "School behaviour and attendance: parental responsibility measures"
+    assert_select ".app-c-document-list__attribute time[datetime='2017-01-05T14:50:33+00:00']", text: "5 January 2017"
+    assert_select ".app-c-document-list__attribute", text: "Statutory guidance"
+
+    assert_select "#{li} a[href='/government/publications/school-exclusion']", text: "School exclusion"
+    assert_select ".app-c-document-list__attribute time[datetime='2017-07-19T16:01:48+01:00']", text: "19 July 2017"
+
+    #check date is rendered correctly when it is GMT
```
fofr: Not sure we need these timezone tests given the component simply handles what it’s given

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,73 @@
+name: Document list
+description: An ordered list of links to documents including document type and when updated.
+body: |
+  Accepts a group of documents from a content item and outputs a list of links to those documents. The component expects the public_updated_at value as a date object, but can also convert a correctly formatted date string.
```
fofr: No longer does any conversion – this need updating.

app/views/components/docs/document-list.yml
```diff
@@ -0,0 +1,73 @@
...
+
+  If a group_index parameter is included the component adds link tracking attributes to each link.
```
fofr: Same here

app/views/components/docs/document-list.yml
fofr: Could the first item in the list not be capitalised. According to GOV.UK styles each list item must be read as a sentence starting from the line before the start of the list: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps

eg:
* Links in the component must be usable with touch
* Links in the component must have visible text

So:

Links in the component must:

* be usable with touch
* have visible text

app/views/components/docs/document-list.yml
andysellick: Good spot - this whole section needs rewriting now.

app/views/components/docs/document-list.yml
andysellick: Thanks.

test/components/document_list_test.rb
andysellick: Ok.

app/views/components/docs/document-list.yml
andysellick: Copy and paste fail :)

app/presenters/document_collection_presenter.rb
```diff
@@ -23,13 +23,26 @@ def groups
...
+    group_documents(group).each_with_index.map do |link, link_index|
       {
-        public_updated_at: Time.zone.parse(link["public_updated_at"]),
-        document_type: link["document_type"],
-        title: link["title"],
-        base_path: link["base_path"]
+        link: {
+          text: link["title"],
+          path: link["base_path"],
+          data_attributes: {
```
nickcolley: Liking this change nice work @andysellick @fofr

PR #446, opened 2017-08-14
Return 406 status for unsupported formats

fofr:
When a request is made for an unsupported format, eg text/javascript, the lack of a respond_to block meant we attempted to render it without a layout.

This led to a body being sent to slimmer without a #wrapper element,
causing the error:
`> undefined method 'to_html' for nil:NilClass`
in the BodyInserter processor

* Add a respond_to block to respond only to known formats
* Catch unknown format errors and return 406 (not acceptable)

This matches: https://github.com/alphagov/finder-frontend/pull/271

`format.html` must come before `format.atom` in the block. If a browser sends an unspecific accepts header, eg `*/*`, Rails falls through to the first format/variant declared in `respond_to`.

See also: https://github.com/gds-attic/multipage-frontend/pull/23

### DIFF
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -72,6 +72,41 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_redirected_to content_item['base_path']
   end
 
+  test "returns HTML when an unspecific accepts header is requested (eg by IE8 and below)" do
+    request.headers["Accept"] = "*/*"
+    content_item = content_store_has_schema_example('travel_advice', 'full-country')
+
+    get :show, params: {
+      path: path_for(content_item)
+    }
+
+    assert_match(/text\/html/, response.headers['Content-Type'])
+    assert_response :success
+    assert_select '#wrapper'
+  end
+
+  test "returns a 406 for XMLHttpRequests" do
+    request.headers["X-Requested-With"] = "XMLHttpRequest"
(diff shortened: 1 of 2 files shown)


test/controllers/content_items_controller_test.rb
```diff
@@ -72,6 +72,41 @@ class ContentItemsControllerTest < ActionController::TestCase
...
+    get :show, params: {
+      path: path_for(content_item)
+    }
+
+    assert_match(/text\/html/, response.headers['Content-Type'])
+    assert_response :success
+    assert_select '#wrapper'
+  end
+
+  test "returns a 406 for XMLHttpRequests" do
```
boffbowsh: I think the only reason this is ok at the moment is because we don't have CORS headers that allow use of our HTML content from other sites anyway, and we know we're not (currently) loading content via Ajax. Otherwise I would think that GETting HTML in a Javascript context would be acceptable. 

I'm not sure what to think about this.

fofr: I think the test name is misleading, sorry.

If you make an XMLHttpRequest with "Accepts: text/html", you will still receive an HTML response.

fofr: (Aside: Passing an empty `Accepts: ` header in minitest appears to get ignored.)

fofr: Fixed in https://github.com/alphagov/government-frontend/pull/447/commits/f1ba8adf2b705432638a6f58afd8bda1a91feb14

PR #448, opened 2017-08-15
Create figure component

vanitabarrett:
The image component:
* Replaces existing responsive sidebar images
* Takes parameters for href, alt text and caption
* Deliberately errors if no href or alt text is given
* Is documented in the component guide (shown below)

Note: rendering of this component in the component guide currently uses an image URL from a GOV.UK page - if there are any other suggestions for images to use as an example in the guide, feel free to suggest.

screen shot 2017-08-15 at 11 20 31
screen shot 2017-08-15 at 11 20 42

Examples:
* [Case Study](https://government-frontend-pr-448.herokuapp.com/government/case-studies/2013-elections-in-swaziland)
* [News Article](https://government-frontend-pr-448.herokuapp.com/government/news/the-personal-independence-payment-amendment-regulations-2017-statement-by-paul-gray)
* [Travel Advice](https://government-frontend-pr-448.herokuapp.com/foreign-travel-advice/nepal)

### DIFF
--- a/app/assets/stylesheets/components/_figure.scss
+++ b/app/assets/stylesheets/components/_figure.scss
@@ -1,0 +1,11 @@
+.app-c-figure {
+  @include responsive-bottom-margin;
+}
+
+.app-c-figure__image {
+  max-width: 100%;
+}
+
+.app-c-figure__figcaption {
+  @include core-16;
+}
--- a/app/views/components/docs/figure.yml
+++ b/app/views/components/docs/figure.yml
@@ -1,0 +1,24 @@
+name: Figure
+description: A figure containing an image that never exceeds the component’s width
+body: |
+  A figure should be used for images that are referenced within a page, but which can be moved around without affecting the flow of the page [(see guidance here)](https://developer.mozilla.org/en/docs/Web/HTML/Element/figure)
+
+  Examples:
+
+  - [Case Study containing an image](/government/case-studies/2013-elections-in-swaziland)
+  - [Travel Advice containing a map (image)](/foreign-travel-advice/nepal)
+accessibility_criteria: |
+  The figure must:
+
+  - provide an informative text description, as alt text or caption
+
+fixtures:
+  default:
+    src: 'https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg'
+  figure_with_alt_text:
+    src: 'https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg'
+    alt: 'An image of the lords chamber'
+  figure_with_caption:
+    src: 'https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg'
+    alt: 'An image of the lords chamber'
+    caption: 'The Lords Chamber'
--- a/app/views/shared/_sidebar_image.html.erb
+++ b/app/views/shared/_sidebar_image.html.erb
@@ -1,6 +1,3 @@
 <% if image %>
-  <figure class="sidebar-image">
-    <img src="<%= image["url"] %>" alt="<%= image["alt_text"] %>">
-    <% if image["caption"].present? %><figcaption><%= image["caption"] %></figcaption><% end %>
-  </figure>
+  <%= render 'components/figure', src: image["url"], alt: image["alt_text"], caption: image["caption"] %>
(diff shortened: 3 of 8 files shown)


fofr:
Can you include links to some example formats where this is used? And screenshots showing usage in context.

test/components/figure_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    "figure"
+  end
+
+  test "fails to render a figure when no data is given" do
+    assert_raise do
+      render_component({})
+    end
+  end
+
+  test "fails to render a figure when a href is provided but alt text isn't" do
```
fofr: Not sure about this. Ideally – yes, but we aren't in control of the content and this may silently hide images that content editors expect to see. 🤔

app/assets/stylesheets/components/_figure.scss
```diff
@@ -0,0 +1,11 @@
+.app-c-figure {
+  @include responsive-bottom-margin;
+
+  img {
```
fofr: This should be `.app-c-figure__image`

app/views/components/docs/figure.yml
```diff
@@ -0,0 +1,20 @@
+name: Figure
+description: A responsive image
```
fofr: There should be some indication of what makes it responsive.
eg A figure containing an image that never exceeds the component’s width

app/views/components/docs/figure.yml
```diff
@@ -0,0 +1,20 @@
...
+body: |
```
fofr: We might want to include details about:

- alt text should be included
- when a figure should be used, maybe linking to https://developer.mozilla.org/en/docs/Web/HTML/Element/figure

fofr:
The spacing below the map seems to have changed, compare:
https://government-frontend-pr-448.herokuapp.com/foreign-travel-advice/nepal
https://www.gov.uk/foreign-travel-advice/nepal

test/components/figure_test.rb
vanitabarrett: @fofr Should we make the component less restricted then, so it only errors if no src is passed in?

fofr: Reluctantly, yes – for now. The schema allow images without alt attributes: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json#L202

We might be able to add some logging to see how many images are being rendered without an alt attribute.

vanitabarrett:
Have removed logic making alt_text a required field - can now render with just src.

Also have checked the travel advice pages (with larger bottom margin on map) on desktop and mobile, and all looks ok to me

test/components/figure_test.rb
```diff
@@ -0,0 +1,26 @@
+require 'component_test_helper'
+
+class FigureTest < ComponentTestCase
+  def component_name
+    "figure"
+  end
+
+  test "fails to render a figure when no data is given" do
```
fofr: Test name should be "renders nothing when no data given" (nothing is failing)

app/views/shared/_travel_advice_summary.html.erb
```diff
@@ -10,13 +10,14 @@
...
-    <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>">
+  <% download_caption = capture do %>
     <% if content_item.map_download_url %>
-      <figcaption>
         <%= render 'components/download-link', href: content_item.map_download_url, link_text: "Download map (PDF)" %>
-      </figcaption>
     <% end %>
-  </figure>
+  <% end %>
+  <%= render 'components/figure', src: content_item.map["url"], alt_text: content_item.map["alt_text"], caption: download_caption %>
```
fofr: If there is no `map_download_url`, will this render an empty caption field?

vanitabarrett: From my testing, it will render just the image (no caption)

app/views/shared/_sidebar_image.html.erb
```diff
@@ -1,6 +1,3 @@
 <% if image %>
-  <figure class="sidebar-image">
-    <img src="<%= image["url"] %>" alt="<%= image["alt_text"] %>">
-    <% if image["caption"].present? %><figcaption><%= image["caption"] %></figcaption><% end %>
-  </figure>
+  <%= render 'components/figure', src: image["url"], alt_text: image["alt_text"], caption: image["caption"] %>
```
nickcolley: Can we just use 'alt' rather than 'alt_text'? This would mirror the dom api closer.

Issue #443, opened 2017-08-11
Translation Component needs to provide semantic information for screen reader users


vanitabarrett:
It has been suggested that the translation component needs to provide additional information to screen readers in order to give context, e.g:

```
<section aria-label="translations">...</section>
```

**This label would need to to take a string that can be localised to each different language version depending on the language of the page.**

Issue #442, opened 2017-08-11
Notice Component uses heading element with no following content


vanitabarrett:
The notice component accepts text with no description. When this happens, the text is rendered in a heading element. This is incorrect use of a heading element.

Leonie has suggested the following:
```
<section aria-label="Notice">
    <p>Statistics release cancelled</p>
</section>
```

fofr:
Fixed by #451

Issue #441, opened 2017-08-11
Banner component can contain information before the heading, making it inaccessible to screenreaders


vanitabarrett:
The way the banner component has been implemented means that if an aside is provided, the aside  is above the banner heading in the DOM. This makes it highly likely that screen reader users will miss the aside content. All content needs to follow the heading.

**Current:**
<img width="412" alt="screen shot 2017-08-11 at 16 10 07" src="https://user-images.githubusercontent.com/29889908/29219404-d6ba0eee-7eaf-11e7-8126-3210d1ce4fa2.png">

fofr:
Fixed by #470

PR #449, opened 2017-08-15
Improve accessibility and accessibility documentation of components

fofr:
Updates to accessibility acceptance criteria following a review from @LJWatson, with corresponding updates to improve accessibility of components.

The updated components and guide are available for [review on Heroku](https://government-frontend-pr-449.herokuapp.com/component-guide).

* Update accessibility acceptance criteria for banner, notice, contents list and link components
* Put space between contents number and text for better screen reader pronunciation
* Put notices and banners in `<section>` elements and include aria-label to give an accessible name

This doesn't include fixes for issues we're tracking separately in #441, #442 and #443

### DIFF
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -3,7 +3,6 @@
   aside ||= false
   large_text = !title && !aside
 %>
-
 <% content_block = capture do %>
   <% if title %>
     <h2 class="app-c-banner__title"><%= title %></h2>
@@ -12,23 +11,19 @@
     <%= text %>
   </p>
 <% end %>
-
-<% if aside %>
- <div class="app-c-banner app-c-banner--grid">
-  <div class="grid-row">
-    <div class="column-third">
-      <div class="app-c-banner__aside">
-        <p><%= aside %></p>
+<section class="app-c-banner<% if aside %> app-c-banner--grid<% end %>" aria-label="Notice">
+  <% if aside %>
+    <div class="grid-row">
+      <div class="column-third">
+        <div class="app-c-banner__aside">
+          <p><%= aside %></p>
+        </div>
+      </div>
+      <div class="column-two-thirds app-c-banner__main">
+        <%= content_block %>
       </div>
     </div>
-
-    <div class="column-two-thirds app-c-banner__main">
-      <%= content_block %>
-    </div>
-  </div>
- </div>
-<% else %>
-  <div class="app-c-banner">
-      <%= content_block %>
-  </div>
-<% end %>
+  <% else %>
+    <%= content_block %>
+  <% end %>
+</section>
--- a/app/views/components/docs/banner.yml
+++ b/app/views/components/docs/banner.yml
@@ -1,6 +1,9 @@
 name: Banner
 description: A page banner, designed to highlight important information
-body: Passing an aside to the banner will mean the banner renders in a grid layout.
+body: |
+  Passing an aside to the banner will mean the banner renders in a grid layout.
+
+  Real world example: [Consultation with history banner](/government/consultations/child-trust-fund-consultation-on-allowing-the-transfer-of-savings-from-a-child-trust-fund-to-a-junior-isa)
 accessibility_criteria: |
   The banner must:
 
--- a/app/views/components/docs/banner.yml
+++ b/app/views/components/docs/banner.yml
@@ -5,7 +5,8 @@ accessibility_criteria: |
   The banner must:
 
   - be visually distinct from other content on the page
-  - have a contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
+  - have an accessible name that describes the banner as a notice
+  - have a text contrast ratio higher than 4.5:1 against the background colour to meet [WCAG AA](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast)
 fixtures:
   default:
     text: 'This was published under the 2010 to 2015 Conservative and Liberal Democrat coalition government'
--- a/app/views/components/docs/download-link.yml
+++ b/app/views/components/docs/download-link.yml
@@ -1,5 +1,7 @@
 name: Download link
 description: A link with a file download icon
+body: |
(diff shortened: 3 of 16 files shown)


app/views/components/docs/banner.yml
```diff
@@ -5,7 +5,8 @@ accessibility_criteria: |
   The banner must:
 
   - be visually distinct from other content on the page
-  - have a contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
+  - have an accessible name that describes the banner as a notice
+  - have a text contrast ratio higher than 4.5:1 against the background colour to meet [WCAG AA](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast)
```
andysellick: The actual wording of the WCAG requirement is "of at least". Also doesn't this qualify as "large text" and therefore only need to have a contrast ratio of 3:1?

app/views/components/docs/download-link.yml
```diff
@@ -1,12 +1,19 @@
 name: Download link
 description: A link with a file download icon
+body: |
+  For usability the provided link text should indicate the file type and size of the download.
```
andysellick: Can we update one of the examples on this page to include this information?

app/views/components/docs/download-link.yml
```diff
@@ -1,12 +1,19 @@
...
   The download link must:
 
   - accept focus
   - be focusable with a keyboard
   - be usable with a keyboard
   - indicate when it has focus
+  - change in appearance when touched (in the touch-down state)
+  - change in appearance when hovered
+  - be usable with touch
+  - be usable with speech
```
andysellick: I'm wondering if this is clear enough - obviously you mean with a screen reader, but it doesn't actually say that. Is there a WCAG page we can simply link to here to give a bit more information?

app/views/components/docs/download-link.yml
```diff
@@ -1,12 +1,19 @@
...
+  - have visible text
```
andysellick: Do we need to specify what this means (i.e. presumably contrast ratio? Font size?) or can we link to an appropriate WCAG rule?

app/views/components/docs/notice.yml
```diff
@@ -5,9 +5,7 @@ body: |
 
   The component accepts either a simple string description_text parameter that it wraps in a paragraph, or a description_govspeak parameter that is rendered through govspeak for more complex HTML layout.
 accessibility_criteria: |
-  The notice must:
-
-  - have a border of AA contrast ratio 4.5:1
+  The notice border colour must have a contrast ratio of more than 4.5:1 with its background to be visually distinct.
```
andysellick: Contrast ratio for text is "at least" rather than "more than" but is this level of contrast necessary for non-text elements? What's the WCAG guidance here?

app/views/components/docs/print-link.yml
```diff
@@ -7,10 +7,13 @@ accessibility_criteria: |
   - be focusable with a keyboard
   - be usable with a keyboard
   - indicate when it has focus
+  - change in appearance when touched (in the touch-down state)
+  - change in appearance when hovered
+  - be usable with touch
+  - be usable with speech
+  - have visible text
```
andysellick: See previous comments for the same criteria.

app/views/components/_banner.html.erb
```diff
@@ -12,23 +11,19 @@
...
-<% if aside %>
- <div class="app-c-banner app-c-banner--grid">
-  <div class="grid-row">
-    <div class="column-third">
-      <div class="app-c-banner__aside">
-        <p><%= aside %></p>
+<section class="app-c-banner<% if aside %> app-c-banner--grid<% end %>" aria-label="Notice">
+  <% if aside %>
+    <div class="grid-row">
+      <div class="column-third">
```
andysellick: Probably out of scope of this PR, and we may have discussed this already, but is the component isolated if it relies on external CSS? (in this case the grid)

app/views/components/docs/banner.yml
fofr: We avoid confusion over the definition of large text by sticking with 4.5:1 for everything.

app/views/components/docs/download-link.yml
fofr: We can change speech to dictation.

app/views/components/docs/notice.yml
fofr: There's no guidance for WCAG border colours. But we know this contrast is necessary for text so can infer it's usefulness in distinguishing things from other things.

app/views/components/_banner.html.erb
fofr: Out of scope, yes. Might want to track with an issue.

LJWatson:
@fofr
>"We can change speech to dictation.">

Note that dictation is not the same as speech recognition. Dictation software enables you to dictate content, speech recognition software enables you to control your OS, applications etc. using voice commands. If a different term is needed, then "voice" would be more appropriate.

@andysellick
>"Contrast ratio for text is "at least" rather than "more than" but is this level of contrast necessary for non-text elements? What's the WCAG guidance here?">

WCAG 2.0 doesn't apply this SC to non-text content (though this will change in WCAG 2.1). That said, in this context I suggest we focus on what this pattern needs to do - and I think one of those things is to be obviously visible to everyone.

fofr:
I've changed "speech" to "voice commands" in c964336 and linked to https://www.w3.org/WAI/perspectives/voice.html

app/views/components/docs/download-link.yml
vanitabarrett: Not sure if this is what you're referring to anyway, but I think Leonie's referring to speech recognition software like [Dragon](https://www.nuance.com/dragon/accessibility-solutions.html)?

app/views/components/_banner.html.erb
nickcolley: In this case the grid classes would be better suited to being `.govuk-o-grid__row` `.govuk-o-grid__column` etc. Since objects are shared across components.

PR #450, opened 2017-08-15
Update rails and other gems

tomnatt:


tomnatt:
@nickcolley Good point. Tested and looks ok.

PR #451, opened 2017-08-17
Only use heading in notice if description present

vanitabarrett:
The notice component was not accessible as headings were being used without any content following.
A check is now done to see if description is present before placing the title in a heading or paragraph.

Fixes issue #442

### DIFF
--- a/app/views/components/_notice.html.erb
+++ b/app/views/components/_notice.html.erb
@@ -1,14 +1,20 @@
 <% if defined?(title) %>
   <%
-    description_text ||= ""
-    description_govspeak ||= ""
+    description_text ||= false
+    description_govspeak ||= false
   %>
   <section class="app-c-notice" aria-label="Notice">
-    <h2 class="app-c-notice__title"><%= title %></h2>
-    <% if description_text.present? %>
+    <% if description_text || description_govspeak %>
+      <h2 class="app-c-notice__title"><%= title %></h2>
+    <% else %>
+      <span class="app-c-notice__title"><%= title %></span>
+    <% end %>
+
+    <% if description_text %>
       <p class="app-c-notice__description"><%= description_text %></p>
     <% end %>
-    <% if description_govspeak.present? %>
+
+    <% if description_govspeak %>
       <%= render 'govuk_component/govspeak', content: description_govspeak %>
     <% end %>
   </section>
--- a/app/views/components/docs/notice.yml
+++ b/app/views/components/docs/notice.yml
@@ -5,7 +5,10 @@ body: |
 
   The component accepts either a simple string description_text parameter that it wraps in a paragraph, or a description_govspeak parameter that is rendered through govspeak for more complex HTML layout.
 accessibility_criteria: |
-  The notice border colour must have a contrast ratio of more than 4.5:1 with its background to be visually distinct.
+  The notice must:
+
+  - have a border colour contrast ratio of more than 4.5:1 with its background to be visually distinct.
+  - always render headings with associated description content, so there are no isolated heading elements inside the component
 fixtures:
   default:
     title: 'Statistics release cancelled'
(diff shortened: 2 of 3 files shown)


fofr:
Is it worth adding an accessibility acceptance criteria to cover this case?

vanitabarrett:
@fofr I considered that, but as you can now pass any content to the notice and it should be accessible (in terms of this fix), I wondered if it was needed?

fofr:
I think it's needed – it’s something we did that broke accessibility, I think we should add a criteria that says "headings are always rendered with associated content" or something similar, and include a unit test to confirm it.

andysellick:
Have we tested that a screen reader interprets this component in the same way when the content is a paragraph and not a heading? My concern is that it might get overlooked depending on the configuration - imagine the situation if a heading was one colour and a paragraph was another.

app/views/components/_notice.html.erb
```diff
@@ -3,8 +3,15 @@
     description_text ||= ""
     description_govspeak ||= ""
   %>
+
   <section class="app-c-notice" aria-label="Notice">
-    <h2 class="app-c-notice__title"><%= title %></h2>
+
```
andysellick: A bit of inconsistent line spacing in this file - line gaps 6 and 8 seem unnecessary, but there probably should be one between 16 and 18.

app/views/components/_notice.html.erb
```diff
@@ -3,8 +3,15 @@
...
+    <% if description_text.present? || description_govspeak.present? %>
+      <h2 class="app-c-notice__title"><%= title %></h2>
+    <% else %>
+      <p class="app-c-notice__title"><%= title %></p>
+    <% end %>
+
     <% if description_text.present? %>
       <p class="app-c-notice__description"><%= description_text %></p>
     <% end %>
```
andysellick: Is this entire file indented?

app/views/components/_notice.html.erb
```diff
@@ -3,8 +3,15 @@
     description_text ||= ""
     description_govspeak ||= ""
   %>
+
   <section class="app-c-notice" aria-label="Notice">
-    <h2 class="app-c-notice__title"><%= title %></h2>
+
+    <% if description_text.present? || description_govspeak.present? %>
```
andysellick: Not sure if this would work, but I think if you declare description_text ||= false you could then just do if description_text (same with description_govspeak)

vanitabarrett:
@andysellick I'm not sure I understand your comment about screen readers interpreting the component in the same way - can you give another example?

app/views/components/_notice.html.erb
vanitabarrett: We've definitely done this for other components, so I see no reason why it wouldn't work here. Was there any particular reason why you chose to use `.present?` when you implemented the component?

app/views/components/_notice.html.erb
vanitabarrett: It's inside an `if` statement?

andysellick:
@vanitabarrett I just want to make sure we've tested this change properly. Voiceover on Mac announces a heading as a heading, so if in some circumstances this component isn't a heading, I want to be sure that people using assistive technology don't get confused or miss out on what is being conveyed. I'm trying to think of another example that would be the equivalent for a sighted person - imagine if a website had warning messages that were the same level of severity but worded differently (say one was "Warning" and one was "Caution") but they meant exactly the same thing, that would be a little confusing, wouldn't it?

vanitabarrett:
@andysellick I think the fact that the component is now wrapped in a section with aria-label 'Notice' means that the same information/level of information is always conveyed regardless of whether a heading or a paragraph is contained within it? I've tested it with the various notices across GOV.UK and I think the information is clear.

app/views/components/_notice.html.erb
andysellick: My ruby experience may have grown since then. @fofr should be able to rule whether this suggestion is worth changing or not.

app/views/components/_notice.html.erb
andysellick: Sorry, I was confused by the way github displays pages. Looking at the complete file it looked like there was an extra tab on the left of every line, but there isn't, so never mind.

app/views/components/_notice.html.erb
fofr: This would be a sensible change.

app/views/components/_notice.html.erb
```diff
@@ -1,14 +1,20 @@
...
+    description_text ||= false
+    description_govspeak ||= false
   %>
   <section class="app-c-notice" aria-label="Notice">
-    <h2 class="app-c-notice__title"><%= title %></h2>
-    <% if description_text.present? %>
+    <% if description_text || description_govspeak %>
+      <h2 class="app-c-notice__title"><%= title %></h2>
+    <% else %>
+      <p class="app-c-notice__title"><%= title %></p>
```
nickcolley: Nit: Should this be a `span`? Since we want it to have no semantic meaning?

(Also less likely to be impacted by anyone doing pesky p globals 😱 )

PR #452, opened 2017-08-17
Test component guide

nickcolley:
This PR prepares government-frontend to be able to pick up throw accessibility errors from the component guide.

Related PR: https://github.com/alphagov/govuk_publishing_components/pull/33

### DIFF
--- a/test/components/all_components.rb
+++ b/test/components/all_components.rb
@@ -1,4 +1,4 @@
-require 'component_test_helper'
+require 'test_helper'
 
 # TODO: Move this logic to the GOVUK Publishing Components gem
 # and have the tests visit each component.
--- a/test/components/all_components.rb
+++ b/test/components/all_components.rb
@@ -6,13 +6,10 @@ class AllComponentsTest < ActionDispatch::IntegrationTest
   test "renders all component guide pages without erroring" do
     visit '/component-guide'
     components = all(:css, '.component-list a').map do |el|
-      {
-        text: el[:text],
-        href: el[:href]
-      }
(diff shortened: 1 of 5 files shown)


test/components/all_components.rb
```diff
@@ -0,0 +1,18 @@
...
+    visit '/component-guide'
+    components = all(:css, '.component-list a').map do |el|
+      {
+        text: el[:text],
+        href: el[:href]
+      }
+    end
+    components.each do |component|
+      visit component[:href] + '/preview'
+    end
```
fofr: Could this be:
```ruby
components = all(:css, '.component-list a').each do |el|
  visit "#{el[:href]}/preview"
end
```

nickcolley: I think it suffered from the issue we ran into where the instances can't be visited since they 'no longer exist'

fofr:
```ruby
component_previews = all(:css, '.component-list a').map do |el|
  "#{el[:href]}/preview"
end

component_previews.each do |component|
  visit component
end
```

nickcolley: That's better, I think I was tired when I wrote this because there's also unneeded data in there too. Will update

test/components/all_components.rb
```diff
@@ -0,0 +1,15 @@
+require 'component_test_helper'
```
fofr: I think this can be `test_helper`

fofr:
Rebased with master and updated gem to pick up aXe tests.

Needed to add a806be8 for tests to fail as expected when running locally.

PR #460, opened 2017-08-22
Raise errors for missing translations in dev and test

fofr:
Raise errors for missing translations in dev and test

* Catch missing translations
* Fix incorrect reference to translation: Change `publication` to `publications`


nickcolley:
Small but great change 👍  🤖  💥

PR #466, opened 2017-08-31
Move page title to layout rather than individual views

dwhenry:
This allows us to have a more consistent approach for
page titles.

This work corrects `html_publications` which previously
missed the `[withdrawn]` notice when they had been withdrawn.

https://trello.com/c/Mn8Yk9Z6/278-add-withdrawn-to-page-title

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,13 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title><%= yield :title %> - GOV.UK</title>
+  <title>
+    <%if content_for?(:page) %>
+      <%= yield :page %> - GOV.UK
+    <% else %>
+      <%= @content_item.page_title %> - GOV.UK
+    <% end %>
+  </title>
   <!--[if gt IE 8]><!--><%= stylesheet_link_tag "application", integrity: true, crossorigin: 'anonymous' %><!--<![endif]-->
   <!--[if IE 6]><%= stylesheet_link_tag "application-ie6" %><script>var ieVersion = 6;</script><![endif]-->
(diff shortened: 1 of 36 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,13 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title><%= yield :title %> - GOV.UK</title>
+  <title>
+    <%if content_for?(:page) %>
```
fofr: This should be `:title`

fofr:
Do you have some examples of withdrawn pages that were missing the title before?

app/views/layouts/application.html.erb
dwhenry: yes it should: fixed.

dwhenry:
We have taken some screen shots of the examples in google and will do again once this is released\

https://gov-uk.atlassian.net/wiki/spaces/GS/pages/166494324/Add+withdrawn+to+title+of+html+publications+content

fofr:
Titles look ok in the Heroku review app:

https://government-frontend-pr-466.herokuapp.com/government/publications/rates-and-allowances-inheritance-tax-thresholds/inheritance-tax-thresholds

https://government-frontend-pr-466.herokuapp.com/foreign-travel-advice/thailand/print

Do we need any tests to avoid regressions?

fofr:
Do we need any tests to avoid regressions?

dwhenry:
Given the likelihood of regression here is low, I think the overhead of creating the test is too big to justify at this time. Let me know if you disagree. :)

I did look into doing it and we would need to create a new example file in content-schemas for withdrawn html_publiscation and then test the hide title field.

PR #469, opened 2017-09-01
Ensure Withdrawn is included on all withdrawn content

dwhenry:
This cover a few edge cases where custom page_titles
where being generated

https://trello.com/c/Mn8Yk9Z6/278-add-withdrawn-to-page-title


test/presenters/guide_presenter_test.rb
```diff
@@ -17,9 +17,9 @@ def schema_name
       end
     end
 
-    test "presents part titles with their index" do
```
fofr: Shouldn't the old test remain?

dwhenry: umm... yes, this has been reverted..

PR #470, opened 2017-09-01
Improve banner accessibility

andysellick:
- ordering of aside and main banner content changed so aside occurs in the dom after the main content, so screen readers should read the heading first (unless you're using VoiceOver in Chrome, which has weird issues that we've decided to ignore as they're out of our control)
- changed aside element to an aside
- appearance of component should not have changed at all

https://government-frontend-pr-470.herokuapp.com/component-guide/banner

screen shot 2017-09-01 at 11 15 27

https://trello.com/c/J6907ojp/93-2-banner-component-is-inaccessible-to-screen-reader-users

### DIFF
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -14,13 +14,13 @@
 <section class="app-c-banner<% if aside %> app-c-banner--grid<% end %>" aria-label="Notice">
   <% if aside %>
     <div class="grid-row">
+      <div class="app-c-banner__main">
+        <%= content_block %>
+      </div>
       <div class="column-third">
         <div class="app-c-banner__aside">
           <p><%= aside %></p>
         </div>
-      </div>
-      <div class="column-two-thirds app-c-banner__main">
-        <%= content_block %>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


app/views/components/_banner.html.erb
```diff
@@ -14,13 +14,13 @@
 <section class="app-c-banner<% if aside %> app-c-banner--grid<% end %>" aria-label="Notice">
   <% if aside %>
     <div class="grid-row">
+      <div class="app-c-banner__main">
+        <%= content_block %>
+      </div>
       <div class="column-third">
-        <div class="app-c-banner__aside">
+        <aside class="app-c-banner__aside">
```
fofr: I'm not sure this should be an aside, despite the variable name. The content we're putting in there is not tangential, it's important and directly related to the main content. (https://developer.mozilla.org/en/docs/Web/HTML/Element/aside)

nickcolley: Might be worth changing the class

fofr: And the input variable name.

PR #473, opened 2017-09-04
Redirect to base_path/ for first part of guides

Rosa-Fox:
[Trello](https://trello.com/c/ST5fkS0a/285-implement-redirect-from-default-chaptered-guides-pages)
- Internal links within a guide now point to the `base_path`, not to the `base_path` + `slug` of the first part. However, the longer URL is still accessible without redirecting to its shorter equivalent if entered manually into the address bar.
- This is not ideal for analytics, as we have to deal with two URLs for the same page.
- This PR solves this issue by redirecting the slug of the first part to /base_path.

### DIFF
--- a/test/presenters/parts_test.rb
+++ b/test/presenters/parts_test.rb
@@ -125,6 +125,21 @@ class PartsTest < ActiveSupport::TestCase
     refute @parts.has_valid_part?
   end
 
+  test 'invalid when slug for first part is present in URL' do
+    class << @parts
+      def part_slug
+        'first-slug'
+      end
+
+      def requested_content_item_path
+        base_path + '/' + part_slug
+      end
+    end
+
+    assert @parts.requesting_a_part?
+    refute @parts.has_valid_part?
+  end
+
   test 'defaults to first part as current part when parts exist but no part requested' do
     presenting_first_part_in_content_item
 
--- a/app/presenters/parts.rb
+++ b/app/presenters/parts.rb
@@ -17,7 +17,7 @@ module Parts
   end
 
   def has_valid_part?
-    !!current_part
+    !!current_part && current_part != parts.first
   end
 
   def current_part_title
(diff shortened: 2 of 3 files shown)


dwhenry:
I have some concerns about this solution. 

My understanding from the trello card and the PR message was that we wanted to add a redirect if we hit the URL for the first part of the guide.

The tests don't really show clearly the purpose of this code, which is to result in a redirect to the base_path when an invalid part is detected. It would be nice to have an integration test to show this happens. I also feel you could simplify this implementation down to:

```
  def has_valid_part?
    !!current_part && current_path != parts.first
  end
```

Rather than mess with how parts are detected, etc..

My other concern is that we already have code that does a redirect (see below link for an example): 
https://www.gov.uk/apply-for-student-finance/how-to-apply -> redirects already

So it feels like the better solution would be to find the existing code and fix it (broken example):
https://www.gov.uk/standard-visitor-visa/overview -> does not redirect

This way we don't have two separate ways of doing this redirection.

Rosa-Fox:
Thanks @dwhenry. I have simplified the implementation and added a controller test to check the redirect.

test/presenters/parts_test.rb
```diff
@@ -125,6 +125,21 @@ def requested_content_item_path
     refute @parts.has_valid_part?
   end
 
+  test 'does not request a part when valid url of first part is manually entered into the address bar' do
```
dwhenry: The test title does not quite match the implementation as it does "request a part", it isn't a valid part.

fofr:
I think this might break travel advice.

dwhenry:
@fofr both:
https://government-frontend-pr-473.herokuapp.com/foreign-travel-advice/canada/safety-and-security
and
https://government-frontend-pr-473.herokuapp.com/foreign-travel-advice/canada
work as I would expect, Are there other travel advise pages we should check?

PR #482, opened 2017-09-13
Add shared accessibility criteria to components

andysellick:
- replaces all instances of link accessibility criteria with the shared version pulled from the gem
- all other existing component specific accessibility criteria should remain

screen shot 2017-09-13 at 17 16 25

https://trello.com/c/qS1pfVle/132-2-accessibility-criteria-universal-criteria-for-all-components

https://government-frontend-pr-482.herokuapp.com/component-guide

### DIFF
--- a/app/views/components/docs/translation-nav.yml
+++ b/app/views/components/docs/translation-nav.yml
@@ -12,15 +12,8 @@ accessibility_criteria: |
   - [identify the language of the text](https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-other-lang-id.html#meaning-other-lang-id-examples-head)
 
     [Watch a screen reader pronounce text differently based on lang attribute](https://bit.ly/screenreaderpronunciation)
-  - accept focus
-  - be focusable with a keyboard
-  - indicate when it has focus
-  - be usable with a keyboard
-  - change in appearance when touched (in the touch-down state)
-  - change in appearance when hovered
-  - be usable with touch
-  - be usable with [voice commands](https://www.w3.org/WAI/perspectives/voice.html)
-  - have visible text
+shared_accessibility_criteria:
+  - link
 examples:
   default:
     data:
--- a/app/views/components/docs/back-to-top.yml
+++ b/app/views/components/docs/back-to-top.yml
@@ -1,17 +1,7 @@
 name: Back to top link
 description: An anchor link intended to allow users to return to the top of the content.
-accessibility_criteria: |
-  The link must:
-
-  - accept focus
-  - be focusable with a keyboard
-  - be usable with a keyboard
-  - indicate when it has focus
-  - change in appearance when touched (in the touch-down state)
-  - change in appearance when hovered
-  - be usable with touch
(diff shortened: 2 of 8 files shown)


app/views/components/docs/translation-nav.yml
```diff
@@ -11,16 +11,9 @@ accessibility_criteria: |
...
-  - accept focus
-  - be focusable with a keyboard
-  - indicate when it has focus
-  - be usable with a keyboard
-  - change in appearance when touched (in the touch-down state)
-  - change in appearance when hovered
-  - be usable with touch
-  - be usable with [voice commands](https://www.w3.org/WAI/perspectives/voice.html)
-  - have visible text
+  [Watch a screen reader pronounce text differently based on lang attribute](https://bit.ly/screenreaderpronunciation)
```
fofr: This should be indented still.

fofr: See: https://government-frontend-pr-482.herokuapp.com/component-guide/translation-nav

app/views/components/docs/notice.yml
```diff
@@ -9,6 +9,8 @@ accessibility_criteria: |
 
   - have a border colour contrast ratio of more than 4.5:1 with its background to be visually distinct.
   - always render headings with associated description content, so there are no isolated heading elements inside the component
+shared_accessibility_criteria:
+  - link
```
vanitabarrett: Should this be included in this component? The notice component itself doesn't contain any links (as far as I know)?

app/views/components/docs/translation-nav.yml
andysellick: It looked weird when I didn't. Is it meant to be part of the same list item as 'identify the language of the text'?

app/views/components/docs/notice.yml
andysellick: Yeah, I wasn't sure - but one of the examples in the guide has links in. They don't get added directly, I guess. I might be jumping the gun as this might fit more into the bracket of accessibility guidance for inputs.

app/views/components/docs/translation-nav.yml
maxgds: A really picky grammar point: the header and bullet points should agree in terms of whether we're referring to single or multiple, so either
Links in the component must:
... indicate when *they have* focus

or 

*(A|Any)* link in the component must:
...  indicate when it has focus

maxgds: Looks like it to me - it's indicating why you should ID the language.

app/views/components/docs/notice.yml
vanitabarrett: I think if we add link accessibility criteria for all components that may, at some point, include a link being passed to them in some HTML, this might mean we have to add it to the majority of components, like banner, lead para etc... To me, that doesn't seem right, but that's just my personal opinion 🙂

fofr: I agree with @vanitabarrett

PR #490, opened 2017-09-21
Capitalise 'cyhoeddwyd' and 'diweddarwyd' 

tuzz:
https://trello.com/c/tWvPKudB/162-welsh-template-change-capitalisation-of-one-word


fofr:
In en.yml updated is also capitalised. Should it be here too?

tuzz:
@fofr: Good spot! I think so. Updated.

PR #492, opened 2017-09-21
Replace airbrake with sentry

h-lame:
For: https://trello.com/c/6RdsLgOY/51-update-government-frontend


fofr:
Should we include the equivalent of this too:
https://github.com/alphagov/collections/pull/383

fofr:
Can I suggest `ContentParts` rather than `PartsPresenterMethods`, in keeping with other modules. Maybe a comment alluding to clash too? How did this error present itself?

app/controllers/content_items_controller.rb
```diff
@@ -30,7 +28,7 @@ def set_guide_draft_access_token
   end
 
   def load_content_item
-    content_item = content_store.content_item(content_item_path)
+    content_item = Services.content_store.content_item(content_item_path)
```
fofr: :+1: :100:

app/presenters/guide_presenter.rb
```diff
@@ -13,7 +13,7 @@ def page_title
 
   def has_parts?
     unless parts.any?
-      Airbrake.notify("Guide with no parts",
+      GovukError.notify("Guide with no parts",
         error_message: "Guide rendered without any parts at #{base_path}")
```
tijmenb: This needs wrapping in `extra: {}`

h-lame: Oh, didn't know that, will do.

h-lame:
@fofr - the error presented itself in all the tests for the routes and controllers that included the module failing with method not found errors.  I see what you mean about the names of the other modules, `PartsPresenterMethods` does stick out like a sore thumb.  Looking at them though I'm surprised we don't have the same problem with them as some have fairly generic names (`Body` and `Metadata` in particular).  Perhaps it makes sense to introduce a namespace for them to make sure they're unique to us.  Something like `ContentItemAttributes::` maybe?

fofr:
@h-lame I agree, we should namespace in a later PR. Wonder why the Faraday module isn't namespaced.

h-lame:
@fofr - looks like there's no github hosted ruby code with a class called `ContentParts` so I'll use that for now (see: https://github.com/search?l=ruby&q=ContentParts&type=Code&utf8=%E2%9C%93).  Will do a namespace in a later PR.

test/presenters/parts_presenter_methods_test.rb
```diff
@@ -1,6 +1,6 @@
 require 'test_helper'
 
-module PartsStubs
```
fofr: Filename also needs updating

h-lame: Oops! Just spotted this myself; sorry!

PR #493, opened 2017-09-25
Fix auto loading for Services

h-lame:
For: https://trello.com/c/6RdsLgOY/51-update-government-frontend

If we move `/lib/services.rb` to `/app/lib/services.rb` then it will be
available via rails autoloading by default because it lives in a
subfolder of `/app`.  When it lives in `/lib` we have to either require
it explicitly when we want to use it, or add `/lib` to the auto load
or eager load paths.  As `/lib` is already configured to be in those
paths but we're still getting load errors (see:
https://sentry.io/govuk/app-government-frontend/issues/355340514/) it
seems we've not set that up properly and moving it into `/app` will be
the safest option.


h-lame:
Deployed this branch and tested it and it has fixed the problem; there may be alternate solutions though.

kevindew:
You can resolve this by using `eager_load_paths` rather than `autoload_paths`. This commit provides some more info: https://github.com/alphagov/publishing-api/commit/2e291cd0d3780ca9f1e08aaac43e9f0dbf6e12c7

Davidslv:
It feels to me that it's the order rails is loading the files, which it does in an alphabetic order, also the `app` path is eager loaded (https://github.com/rails/rails/blob/v5.1.3/railties/lib/rails/engine/configuration.rb#L38-72)
I could only find one place where Services is being used: https://github.com/alphagov/government-frontend/blob/5cbe9eab2f30fd908083db7ad645da612f80debc/app/controllers/content_items_controller.rb#L31

I would try to do `require 'lib/services'` on top of the controller and see if that works instead

h-lame:
@kevindew - I'll try moving it back and using `eager_load_paths` instead of `autoload_paths`.  I note that a handful of other apps have `app/lib/services.rb` too ([Publisher](https://github.com/alphagov/publisher/blob/master/app/lib/services.rb), [Finder Frontend](https://github.com/alphagov/finder-frontend/blob/master/app/lib/services.rb), [Content Tagger](https://github.com/alphagov/content-tagger/blob/master/app/lib/services.rb) so I wonder if we need to decide what the solution is across all the apps for consistency.

@Davidslv - I'm not a fan of using an explicit require in rails apps because it's not really what you expect to have to do.  Rails is supposed to deal with constant loading without explicit requires (at least from your app's own code) so every time I end up doing it it feels like I've gone off the rails way and given up.

Davidslv:
@h-lame I agree with you, I assumed the `eager_load_paths` was set already and it wasn't working. I see now by @kevindew comment that you exploring it.

kevindew:
@Davidslv that's interesting, I thought it was only the root level of lib that wasn't eager loaded not all of it. I assume that is done so you don't load up classes into memory (like those for rake tasks) that aren't always used in your app.

@h-lame yeah that sounds like something to think about. I tend to think the presence of `app/lib` is a smell since it's basically same purpose as `lib`

h-lame:
Using `eager_load_paths` doesn't work because `lib/generators/format/format_generator.rb` is now eager loaded and it breaks when required at startup with:

    uninitialized constant Generators::NamedBase (NameError)
    /path/to/lib/generators/format/format_generator.rb:3:in `<top (required)>'

To fix we need to require the whole `Rails::Generators` framework by adding `require 'rails/generators'` to the file.  I'll add this missing require to `lib/generators/format/format_generator.rb` so it can be loaded again, but I don't think we want the whole rails generators framework in memory for the web app, so I'm going to go back to the `app/lib` version of this PR.

kevindew:
Ah drat! I wonder if `require` is the way to go - but agree doesn't feel very Rails

h-lame:
@kevindew - thanks for approving.

I think the problem with the generator might highlight why rails now points us at `app/lib` and `lib` as being two very different things.  Code in `app/lib` doesn't fit in a traditional app folder like models, controllers, presenters, workers, etc, but is absolutely required for the app to function.  Code in `lib` is related to the app, but you don't need it for the web-app to work (rake tasks or one-off scripts, generators, support classes for rake tasks, etc...).  It's a fuzzy line for sure, but it makes a mild amount of sense.  The rails way seems to be "don't write `require` statements for code you've written" and using `app` folders, which are all now auto/eager/whatever loaded, is how you achieve that.  For now, in rails 5+, anyway, until they change things again.

kevindew:
@h-lame no problem. Yeah this all feels a bit messier than I anticipated. Have you seen any documentation about the app/lib directory? my understanding was that it was just something people added in on impulse and not something that was sanctioned.

PR #499, opened 2017-10-04
Create publication header component

vanitabarrett:
Replace the existing html publication header with a component which makes use of govuk-title component.

In component guide:

Review app component guide:
https://government-frontend-pr-499.herokuapp.com/component-guide

Visual regression results:
https://government-frontend-pr-499.surge.sh/gallery.html

### DIFF
--- a/app/views/components/_publication-header.html.erb
+++ b/app/views/components/_publication-header.html.erb
@@ -1,0 +1,9 @@
+<header class="app-c-publication-header">
+  <%= render partial: 'govuk_component/title', locals: {
+    title: title,
+    context: context,
+    inverse: true,
+    margin_bottom: 0
+  } %>
+  <p class="app-c-publication-header__last-changed"><%= last_changed %></p>
+</header>
--- a/test/components/publication_header_test.rb
+++ b/test/components/publication_header_test.rb
@@ -1,0 +1,43 @@
+require 'component_test_helper'
+
+class PublicationHeaderTest < ComponentTestCase
+  def component_name
+    "publication-header"
+  end
+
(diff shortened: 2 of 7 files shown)


app/views/components/_publication-header.html.erb
```diff
@@ -0,0 +1,9 @@
+<header class="app-c-publication-header" id="contents">
```
fofr: Should that ID still be there?

vanitabarrett: Good spot, I'll look into it. Thanks 👍

test/components/publication_header_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'component_test_helper'
+
+class PublicationHeaderTest < ComponentTestCase
+  def component_name
+    "publication-header"
+  end
+
+  test "fails to render a publication-header when no text is given" do
```
fofr: > when no text is given
Should this be title? title and context?
It looks like all properties are required?

app/views/components/_publication-header.html.erb
vanitabarrett: @fofr It's being used as an anchor for back to top link at the moment, see this page: https://government-frontend-pr-499.herokuapp.com/government/publications/budget-2016-documents/budget-2016 . 

I guess that should really be moved out of the component itself, maybe into a wrapper element within the template?

fofr: Other pages link to the contents container:
https://www.gov.uk/guidance/hmrc-tools-and-calculators#contents

PR #395, opened 2017-06-29
Allow HTML pub contents to render cleanly with custom numbers

fofr:
Publishers can enter custom numbers. Add support for the forms:

* Numbers without periods (1)
* Numbers with decimals (1.2)
* Allow long numbers to be spaced correctly and not wrap badly (100.)
* Ignore numbers bigger than 999 (eg 2014)
* Ignore contents items that are just numbers

A list of HTML publications with custom numbering:
https://gist.github.com/fofr/622c0c52dfee41cde9402c56ec5d67ea

## Hover and focus states

screen shot 2017-06-29 at 17 04 31

## Examples

https://www.gov.uk/government/publications/xbrl-guide-for-uk-businesses/xbrl-guide-for-uk-businesses

| Old | New |
|-|-|
|screen shot 2017-06-29 at 16 49 17|screen shot 2017-06-29 at 16 51 16|
|screen shot 2017-06-29 at 16 59 00|screen shot 2017-06-29 at 16 58 49|

https://www.gov.uk/government/publications/archived-case-reports/archived-case-reports

| Old | New |
|-|-|
|screen shot 2017-06-29 at 16 53 04|screen shot 2017-06-29 at 16 53 26|

https://www.gov.uk/government/publications/british-forces-air-strikes-in-iraq-monthly-list/raf-air-strikes-in-iraq-and-syria-december-2015

| Old | New |
|-|-|
|screen shot 2017-06-29 at 16 54 27|screen shot 2017-06-29 at 16 54 49|

https://www.gov.uk/government/publications/vietnam-country-of-concern--2/vi-t-nam-bao-cao-nhan-quy-n-2014-c-a-b-ngo-i-giao-anh

| Old | New |
|-|-|
|screen shot 2017-06-29 at 16 56 59|screen shot 2017-06-29 at 16 57 37|

## Examples where new is not as good

https://www.gov.uk/government/publications/fifth-meeting-of-the-leadership-for-libraries-taskforce/minutes-of-the-fifth-meeting-of-the-leadership-for-libraries-taskforce

| Old | New |
|-|-|
|screen shot 2017-06-29 at 17 11 13|screen shot 2017-06-29 at 17 12 41|

PR #500, opened 2017-10-06
Don't wrap numbers without a period

thomasleese:
We've had a ticket come in where someone was writing dates in the headings (e.g. 2 April 2017) and this was being parsed as a numbered list. Since the guidance for govspeak is to include the period after numbers you want to be recognised as numbers, I don't think it makes sense for numbers without a period to be accepted.

It's also unclear why numbers without periods are currently being accepted.

### DIFF
--- a/app/helpers/contents_list_helper.rb
+++ b/app/helpers/contents_list_helper.rb
@@ -7,7 +7,7 @@ module ContentsListHelper
     # Must be followed by a space
     # May contain a period `1.`
     # May be a decimal `1.2`
-    number = /^\d{1,3}(\.?|\.\d{1,2})(?=\s)/.match(content_item_text)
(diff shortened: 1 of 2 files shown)


fofr:
Is there a link to some examples?

fofr:
There's a lot of context here, along with examples of the problems we needed to fix by rendering numbers without periods in contents lists: https://github.com/alphagov/government-frontend/pull/395

fofr:
Essentially numbers without periods are accepted because some content is written that way. The component aims to be lenient in what it accepts.

The examples of the contents list in the component guide should probably state this.

thomasleese:
That's disappointing, but I figured there was a reason why it was like that! I'll close this PR then.

thomasleese:
Just to note that there is currently a workaround of using a `&nbsp;` instead of a ordinary space between the number and the month in the heading, i.e. `2&nbsp;April 2017`.

PR #502, opened 2017-10-10
Add clarity to stats announcement forthcoming notice

bevanloon:
Add a helper to improve the clarity of the forthcoming notification on Statistical Announcements.

Statistical Announcements can have three different date formats:

* Exact e.g. `10th January 2018 9:30am (provisional)`
* One month e.g. `January 2018 (provisional)`
* Two month e.g. `January to February 2018 (provisional)`

To improve the clarity of the notice, we want to add `on`, `in` or `between` before the date component. In addition, when we are working with a two month date range, we want to substitute `and` for the `to` (`...between January and February...`) so we do not imply that the publication is only available between the two months.

Example of an exact date:
https://government-frontend-pr-502.herokuapp.com/government/statistics/announcements/12-week-maternal-assessment-for-q4-201617-and-annual-data-201617--3

Example of a one month date:
https://government-frontend-pr-502.herokuapp.com/government/statistics/announcements/nhs-outcomes-framework-indicators-nov-2018-release

Example of a two month range:
https://government-frontend-pr-502.herokuapp.com/government/statistics/announcements/housing-statistics-1-april-2017-to-30-september-2017

[Trello](https://trello.com/c/vRY4O6vp/208-clarify-statistical-announcement-forthcoming-notice)

### DIFF
--- a/app/helpers/statistics_announcement_helper.rb
+++ b/app/helpers/statistics_announcement_helper.rb
@@ -1,0 +1,27 @@
+module StatisticsAnnouncementHelper
+  def on_in_between_for_release_date(date)
+    return "on #{date}" if date_is_exact_format?(date)
+    return "in #{date}" if date_is_one_month_format?(date)
+    return "between #{replace_on_with_and(date)}" if date_is_two_month_format?(date)
+    date
+  end
+
+private
+
+  def replace_on_with_and(date_in_two_month_format)
+    re = /\s(to)\s/
+    date_in_two_month_format.sub(re, " and ")
+  end
+
+  def date_is_two_month_format?(date)
+    date =~ /\A(\w+)\s(to)\s(\w+)/
+  end
+
+  def date_is_one_month_format?(date)
+    date =~ /\A(\w+)\s(\d{1,4})/
+  end
+
+  def date_is_exact_format?(date)
+    date.downcase =~ /\A(\d{1,2})\s(\w+)\s(\d{4})\s(\d{1,2}:\d{1,2})(am|pm)/
+  end
+end
--- a/app/presenters/statistics_announcement_presenter.rb
+++ b/app/presenters/statistics_announcement_presenter.rb
@@ -1,8 +1,9 @@
 class StatisticsAnnouncementPresenter < ContentItemPresenter
   include ContentItem::Metadata
   include ContentItem::TitleAndContext
+  include StatisticsAnnouncementHelper
 
-  FORTHCOMING_NOTICE = "These statistics will be available".freeze
+  FORTHCOMING_NOTICE = "These statistics will be released".freeze
 
   def release_date
     content_item["details"]["display_date"]
@@ -58,7 +59,7 @@ class StatisticsAnnouncementPresenter < ContentItemPresenter
   end
 
   def forthcoming_notice_title
-    "#{FORTHCOMING_NOTICE} #{release_date}"
+    "#{FORTHCOMING_NOTICE} #{on_in_between_for_release_date(release_date)}"
   end
 
   def forthcoming_publication?
--- a/test/helpers/statistics_announcement_helper_test.rb
+++ b/test/helpers/statistics_announcement_helper_test.rb
@@ -1,0 +1,21 @@
+require 'test_helper'
+
+class StatisticsAnnouncementHelperTest < ActionView::TestCase
+  test "returns 'on' if the date is an exact format" do
+    assert_equal "on 10 January 2017 9:30am", on_in_between_for_release_date("10 January 2017 9:30am")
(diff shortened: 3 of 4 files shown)


fofr:
@kevindew I think the schema is correct, it's a string not a date.

bevanloon:
@kevindew - the dates for these are strings because we can set them to be one of three types (exact, one month, two month). 

Screenshot from whitehall where these get created:

screen shot 2017-10-10 at 09 51 02

kevindew:
@fofr I meant along the lines that it's a string that we rely on the formatting so it not having to match a pattern or similar - e.g. if you had "to be determined" in there you'd get odd output.

app/helpers/statistics_announcement_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+  def date_is_two_month_format?(date)
+    date =~ /\A(\w+)\s(to)\s(\w+)/
+  end
+
+  def date_is_one_month_format?(date)
+    date =~ /\A(\w+)\s(\d{1,4})/
+  end
+
+  def date_is_exact_format?(date)
+    date.downcase =~ /\A(\d+)\s(\w+)\s(\d{4})\s(\d{1,2}:\d{1,2})(am|pm)/
```
fofr: That's some funky regexin’

bevanloon: :) yeah anyone with better regex-fu who fancies neatening these up gets :1st_place_medal:  from me.  I'm slightly worried there's an edge case I haven't thought of that fails the regex.

fofr:
@kevindew Should we publishing the new string rather than risking constructing it in the frontend? Or using dates to more reliably construct in frontend? 🤔

app/helpers/statistics_announcement_helper.rb
kevindew: Do you only care at this point whether it's a valid date? you might be able to skip this regex and just do Date.parse or a similar method at this point

bevanloon: The problem I had with `Date.parse` is that it seems very forgiving:

```
irb(main):001:0> Date.parse("November to December 2017")
=> Fri, 01 Dec 2017
irb(main):002:0> Date.parse("January 2018")
=> Mon, 01 Jan 2018
irb(main):003:0> Date.parse("13 December 2017 9:30am")
=> Wed, 13 Dec 2017
```

so I can't rely on it to determine the format of the string since they are all valid dates.

kevindew:
@fofr that's a tough question really. I think if it's considered the content of the document (e.g. we can live with it not being consistent with old things and new things) then it's probably worth constructing within publishing app and just output verbatim.

Whereas if it's considered data then it should probably be just an object that has means to represent exact and relative dates with the logic for transforming it being always handled frontend. 

I guess at the moment we're in the middle of that so we're caught with a tight coupling of string formatting of a publishing app and rendering app which seems somewhat fragile.

app/helpers/statistics_announcement_helper.rb
kevindew: Yup sure, though you'll have already checked for the first two scenarios with your simpler regex so it'd only be for dealing with the last of those scenarios.

bevanloon:
If this gets modelled into the publishing app, then we are looking at a slightly larger piece of work, I think. We will need to change the schema to handle a new field and update whitehall to push the notice into it. As @kevindew said, this could leave us in an inconsistent state unless we republish all statistical announcements.

The advantage is that we could consider the "notice" as part of the document as @kevindew notes and do away with the logic in the frontend to decide whether to display the "cancelled" or "forthcoming" state (as long as we republish all old statistical announcements to bring them up to snuff).

On the other hand, I don't think we should shy away from string formatting if it makes sense. We are in the frontend where strings are the bread and butter. :)

keithiopia:
The wording looks good (though I'm biased).

app/helpers/statistics_announcement_helper.rb
bevanloon: so having thought about this a bit and reading through the rest of the comments around the issue, I noticed the code didn't take into account a completely unexpected format being passed in. If some string got passed in here that failed all three conditions, we would get a front end message that said "These statistics will be released". I've updated to return any unexpected format back so we at least append the date to the sentence as we do now without an "in", "on" or "between".

To take that into account, I think I still need to check for this format specifically - what do you think @kevindew?

kevindew: good catch 👍 

I don't think it's necessary to check that one specifically, but it's more in the realm of two ways to skin a cat than offering a massive advantage to use `Date.Parse`. Only seems like something worth doing if there are risks the formatting could be slightly different - which seems unlikely.

PR #504, opened 2017-10-16
Remove lead paragraph component

maxgds:
- Delete lead paragraph files
- Update views to point to new location (coming from static)
- Update test to check input instead of output due to testing limitations
  around static

Review app component guide:
https://government-frontend-pr-504.herokuapp.com/component-guide


lib/generators/format/templates/format.html.erb
```diff
@@ -6,4 +6,4 @@
   </div>
 </div>
 
-<%%= render 'components/lead-paragraph', text: @content_item.description %>
+<%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
```
fofr: This should still be `<%%` – it's a peculiar style used to escape a percent in ERB.
eg https://stackoverflow.com/questions/2322413/how-do-i-escape-erb-code-in-fixtures

maxgds: Fixed.

PR #508, opened 2017-10-30
Hebrew is a right-to-left language

gpeng:
This commit adds `rtl` to the he.yml locale as Hebrew should read right
to left.

Fixes [Zendesk](https://govuk.zendesk.com/agent/tickets/2496478)

Example with the issue: 

https://www.gov.uk/government/news/foreign-secretary-article-on-the-balfour-declaration

image

After: 

image

### DIFF
--- a/config/locales/he.yml
+++ b/config/locales/he.yml
@@ -1,4 +1,6 @@
 he:
+  i18n:
+    direction: rtl
   language_names:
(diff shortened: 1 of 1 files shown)


fofr:
😱 

@gpeng Does this need to be done in any other apps too?

PR #506, opened 2017-10-27
[DO NOT MERGE] Remove education AB test from government-frontend

vanitabarrett:
Disabled the new breadcrumb and taxonomy related links. Essentially, rollback the B variant and always show A instead.

We will revisit this later and add extra logic to govuk_navigation helpers to only show the B variant on pages tagged to the new taxonomy. This PR strips away the AB logic to make it easier to do that work.

Example page in Education AB test: https://www.gov.uk/career-skills-and-training

**Before:**
screen shot 2017-10-27 at 11 44 34

**After:**
screen shot 2017-10-27 at 11 44 50

Review app component guide:
https://government-frontend-pr-506.herokuapp.com/component-guide

Visual regression results:
https://government-frontend-pr-506.surge.sh/gallery.html

Part of:
https://trello.com/c/ZQNGFeUa/65-remove-ab-test-from-frontend-and-government-frontend

PR #510, opened 2017-11-01
[Do not merge] Show taxonomy-based navigation in more cases 

tijmenb:
This is the current decision tree for when to show the taxonomy-based navigation:

image

This PR implements the following new logic:

image

[Diagrams on the Wiki](https://gov-uk.atlassian.net/wiki/spaces/GFED/pages/187924634/When+we+show+what+on+the+content+pages)

## Examples

- Page tagged to mainstream browse: https://government-frontend-pr-510.herokuapp.com/career-skills-and-training
- Page not tagged to browse, but tagged to new taxonomy: https://government-frontend-pr-510.herokuapp.com/government/publications/financial-support-for-full-time-students-of-higher-education-in-2016-to-2017
- Page not tagged to new taxonomy: https://government-frontend-pr-510.herokuapp.com/government/publications/the-olympic-and-paralympic-legacy-inspired-by-2012-second-annual-report

Supersedes https://github.com/alphagov/government-frontend/pull/506

https://trello.com/c/ZQNGFeUa


app/models/navigation_type.rb
```diff
@@ -0,0 +1,35 @@
+class NavigationType
+  NEW_NAVIGATION_CONTENT_ITEM_SCHEMAS =
+    %w{answer contact guide detailed_guide document_collection publication}.freeze
+
+  def initialize(content_item)
+    @content_item = content_item
+  end
+
+  def should_present_new_navigation_view?
```
fofr: `should_present_taxonomy_navigation`?

app/models/navigation_type.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def content_is_tagged_to_a_taxon?
+    @content_item.dig("links", "taxons").present?
+  end
+
+  def content_is_tagged_to_browse_pages?
+    @content_item.dig("links", "mainstream_browse_pages").present?
+  end
+
+  def content_schema_has_new_navigation?
```
fofr: As mentioned in person, we should probably move away from `new_navigation`. Our "new navigation" will become our tested variants.

tijmenb:
@fofr renamed everything in a separate commit.

vanitabarrett:
This changes means there are less related links on the side compared to the B-variant that was running in the AB test.

**AB Test (B Variant):**
screen shot 2017-11-01 at 13 48 43


**This change:**
screen shot 2017-11-01 at 13 48 37

app/views/shared/_breadcrumbs.html.erb
```diff
@@ -1,5 +1,3 @@
-<% # THIS IS PART OF EDUCATION NAVIGATION A/B TESTING - PLEASE KEEP IN SYNC WITH +new_navigation.erb %>
```
fofr: Do we still need a warning about keeping templates in sync?
Seems like this logic could be moved into a presenter? Maybe in a future PR.

tijmenb: I think people who'll modify those templates will be aware of the different variants, so I don't think we need the warning.

PR #515, opened 2017-11-02
Add the TrafficSignsSummary A/B test

alecgibson:
The TrafficSignsSummary A/B test will test a content improvement to the
summary of the [Know your traffic signs][1] publication.

[1]: https://www.gov.uk/government/publications/know-your-traffic-signs

Visual regression results:
https://government-frontend-pr-515.surge.sh/gallery.html

## Dependencies:

- [x] https://github.com/alphagov/cdn-configs/pull/9
- [x] https://github.com/alphagov/fastly-configure/pull/44

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -16,6 +16,7 @@ class ContentItemsController < ApplicationController
     load_content_item
     set_up_navigation
     setup_tasklist_ab_testing
+    set_up_traffic_signs_summary_ab_testing
     set_expiry
     set_access_control_allow_origin_header if request.format.atom?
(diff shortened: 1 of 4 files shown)


alecgibson:
[Add redirects for a/b testing content (Transport)](https://trello.com/c/TzBmz1gi/765-add-redirects-for-a-b-testing-content-transport)

fofr:
FYI, we'll soon be running a set of multivariate tests across a large set of government-frontend pages as part of the navigation work. I don't think this test will be in the same scope as ours, but there are others planned we should probably co-ordinate to avoid conflicts.

alecgibson:
@fofr sure thing. This is a tiny test on a single page. Should only be running for a month.

PR #516, opened 2017-11-03
Use the same metadata for all pages

fofr:
* Don’t hide the `part of` bits of metadata on new taxonomy pages
* We will be iterating what nav is shown in future multivariate tests
* Part of includes both trivially related things and important related
things – we need to do better in distinguishing them

Review app vs live:
https://government-frontend-pr-516.herokuapp.com/government/publications/national-apprenticeship-awards-apprentice-categories
https://www.gov.uk/government/publications/national-apprenticeship-awards-apprentice-categories?ABTest-EducationNavigation=B

https://trello.com/c/zuBOPznV/74-remove-ab-test-from-government-frontend

### DIFF
--- a/app/presenters/content_item/metadata.rb
+++ b/app/presenters/content_item/metadata.rb
@@ -15,12 +15,6 @@ module ContentItem
       }
     end
 
-    def metadata_for_taxonomy_navigation
-      data = metadata
-      data.delete(:part_of)
-      data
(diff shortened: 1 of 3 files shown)


fofr:
Review app seems to say it’s deployed correctly, but there's no app at the URL. Nothing in the logs 🤔

PR #513, opened 2017-11-02
Add sidebar tasklist to answers and guide pages

Davidslv:
Trello: https://trello.com/c/QUBFYtcf/217-add-a-b-test-functionality-into-the-government-frontend-app-%F0%9F%9A%97

screen shot 2017-11-02 at 15 41 47
screen shot 2017-11-02 at 16 00 04
screen shot 2017-11-02 at 15 59 39
screen shot 2017-11-02 at 15 59 14

PR #524, opened 2017-11-07
Set up Self Assessment A/B test

emmabeynon:
This is set up for an A/B test we are conducting on the Self Assessment page to test alternative login methods.  This test is running until 16/11 after which we plan to remove this code.

[Trello](https://trello.com/c/zfJiG9v8/145-sa-a-b-test-make-changes-to-government-frontend)


emmabeynon:
@fofr Thanks - now removed 👍

fofr:
Modelling and services have added some logic for AB testing in https://github.com/alphagov/government-frontend/pull/513

Is it possible for this test to be configured in a similar way?

ie All test details in a concern?

Frontend design will be adding a further test to government-frontend – we ought to have a consistent way of managing these tests and a way to prevent clashes.

cc @sihugh @Davidslv

leenagupte:
@fofr @sihugh @Davidslv As we talked about yesterday, this test will only run for one week. We have a placeholder [card])(https://trello.com/c/xdmgIGIO/163-tear-down-a-b-test) to remove this code in week 8. 

If the test was running for longer, and we'd had more time to implement it, we would have spent more time figuring out how to make the new routes more generic and moved the A/B test set up into a separate class.

app/controllers/content_items_controller.rb
```diff
@@ -22,6 +23,40 @@ def show
...
+
+  def lost_account_details
+    set_up_self_assessment_ab_test
+    content_item = Services.content_store.content_item(content_item_path)
+    @content_item = ContentItemPresenter.new(content_item, content_item_path)
+    render template: 'content_items/signin/lost-account-details'
+  end
+
+  def sign_in_options
+    if params["sign-in-option"] == "government-gateway"
```
Davidslv: This could be a case/when statement

app/controllers/content_items_controller.rb
```diff
@@ -22,6 +23,40 @@ def show
...
+  def choose_sign_in
+    set_up_self_assessment_ab_test
+    content_item = Services.content_store.content_item(content_item_path)
+    @content_item = ContentItemPresenter.new(content_item, content_item_path)
+    @error = params[:error]
+    render template: 'content_items/signin/choose-sign-in'
+  end
+
+  def not_registered
+    set_up_self_assessment_ab_test
```
Davidslv: you could extract

```
    set_up_self_assessment_ab_test
    @content_item = ContentItemPresenter.new(content_item, content_item_path)
```

Into another method and call it on the 3 actions you have

leenagupte: Good point. Thanks.

app/controllers/content_items_controller.rb
leenagupte: We could have. Is there a benefit to using a case statement?

app/controllers/content_items_controller.rb
leenagupte: Extracted in commit a1a003851b1c2e4768f5157c2f8d9f6adb041078

app/controllers/content_items_controller.rb
Davidslv: DRY principle, but you don't need to address it as it will all be reverted in a week

leenagupte:
Thanks @Davidslv

fofr:
How do I see this on the Heroku review app?

config/initializers/assets.rb
```diff
@@ -9,6 +9,8 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
 Rails.application.config.assets.precompile += %w(
+  self_assessment_b_variant.css
+  application-ie6.css
```
fofr: Why has `application-ie6.css` been duplicated?

PR #525, opened 2017-11-07
Initial PaaS manifest

stephenharker:
The PaaS manifest file will not have any effect for now.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,0 +1,15 @@
+---
+applications:
+- name: govuk-dev-frontend
+  memory: 512M
+  buildpack: ruby_buildpack
+  instances: 1
+  env:  
+    GOVUK_APP_DOMAIN: www.gov.uk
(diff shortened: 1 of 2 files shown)


manifest.yml
```diff
@@ -0,0 +1,15 @@
...
+applications:
+- name: govuk-dev-frontend
+  memory: 512M
+  buildpack: ruby_buildpack
+  instances: 1
+  env:  
+    GOVUK_APP_DOMAIN: www.gov.uk
+    GOVUK_ASSET_ROOT: https://assets.digital.cabinet-office.gov.uk
+    GOVUK_WEBSITE_ROOT: https://www.gov.uk
+    #PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
```
Davidslv: forgotten?

fofr:
Does this also need a rummager var?
https://github.com/alphagov/government-frontend/blob/master/startup.sh#L10

nickcolley:
What work is this change related to? What will this let us do?

PR #526, opened 2017-11-07
Add content navigation multivariate test

fofr:
First pass at adding the content navigation multivariate test. This adds the logic to respond to test bucket, without any of the markup changes or variations.

Start with 4 variants matching https://github.com/alphagov/cdn-configs/pull/17
original, no nav, taxonomy nav, mainstream nav

* Follow “concern” pattern setup in #513 for task list AB test
* Only content with the document types listed are part of the test
* The content must also be tagged to a taxon
* Use a “universal_navigation” variant for all multivariate tests
except original – they are largely the same, significant differences
with original shouldn’t be repeated

Part of:
https://trello.com/c/sVQSGE2e/16-create-multivariant-test-for-content-pages
https://trello.com/c/1caGTIJn/77-create-no-navigation-test-variant-based-on-first-universal-design

### DIFF
--- a/app/controllers/concerns/content_navigation_ab_testable.rb
+++ b/app/controllers/concerns/content_navigation_ab_testable.rb
@@ -37,7 +37,8 @@ module ContentNavigationABTestable
       GovukAbTesting::AbTest.new(
         "ContentNavigation",
         dimension: CONTENT_NAVIGATION_DIMENSION,
-        allowed_variants: CONTENT_NAVIGATION_ALLOWED_VARIANTS
+        allowed_variants: CONTENT_NAVIGATION_ALLOWED_VARIANTS,
+        control_variant: CONTENT_NAVIGATION_ORIGINAL
       )
   end
 
--- a/app/controllers/concerns/content_navigation_ab_testable.rb
+++ b/app/controllers/concerns/content_navigation_ab_testable.rb
@@ -1,0 +1,86 @@
+module ContentNavigationABTestable
+  CONTENT_NAVIGATION_DIMENSION = 67
+  CONTENT_NAVIGATION_ORIGINAL = "Original".freeze
+  CONTENT_NAVIGATION_UNIVERSAL_NO_NAVIGATION = "UniversalNoNav".freeze
+  CONTENT_NAVIGATION_UNIVERSAL_MAINSTREAM_NAVIGATION = "UniversalMainstreamNav".freeze
+  CONTENT_NAVIGATION_UNIVERSAL_TAXON_NAVIGATION = "UniversalTaxonNav".freeze
+
+  CONTENT_NAVIGATION_ALLOWED_VARIANTS = [
+    CONTENT_NAVIGATION_ORIGINAL,
+    CONTENT_NAVIGATION_UNIVERSAL_NO_NAVIGATION,
+    CONTENT_NAVIGATION_UNIVERSAL_MAINSTREAM_NAVIGATION,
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/content_navigation_ab_testable.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def content_navigation_ab_test_applies?
+    @content_item.tagged_to_a_taxon? && content_document_type_is_guidance?
+  end
+
+  def content_document_type_is_guidance?
+    GUIDANCE_DOCUMENT_TYPES.include? @content_item.document_type
+  end
+
+  def content_navigation_test_original_variant?
```
steventux: Not for this PR but I think we could find a way to reduce the constant + method boilerplate for these helper methods and make something generic enough to give us all the help we need for a given collection of variant names.

app/controllers/content_items_controller.rb
```diff
@@ -89,17 +91,20 @@ def set_expiry
...
     # use a convention to automagically load an alternative partial, view or
     # layout.  For example, if I set a variant of :taxonomy_navigation and we render
     # a partial called _breadcrumbs.html.erb then Rails will attempt to load
     # _breadcrumbs.html+taxonomy_navigation.erb instead. If this file doesn't exist,
     # then it falls back to _breadcrumbs.html.erb.  See:
     # http://edgeguides.rubyonrails.org/4_1_release_notes.html#action-pack-variants
-    if @navigation.should_present_taxonomy_navigation?
-      request.variant = :taxonomy_navigation
+    if should_present_universal_navigation?
+      request.variant = :universal_navigation
```
tijmenb: Are we sure about using variants? I would've expected that the universal navigation is a single template, not variants on the existing per-format templates.

fofr: That might be what we end up with. For now there's still a few too many unknowns. While content will all be in two columns, there will be differences eg Guides with their pagination and page links at top vs Publications that have metadata, vs Answers which have neither.

tijmenb: Cool. I think the dream would be a government-frontend that doesn't know anything about `schema_name` and `document_type` and determines layout dependent on the [features it detects](https://www.joezimjs.com/javascript/feature-detection-vs-browser-detection) (not sure if that's 100% possible though).

app/controllers/concerns/content_navigation_ab_testable.rb
```diff
@@ -0,0 +1,86 @@
+module ContentNavigationABTestable
+  CONTENT_NAVIGATION_TEST_NAME = "ContentNavigation".freeze
+  CONTENT_NAVIGATION_DIMENSION = 67
```
Davidslv: You should freeze this one too

app/controllers/content_items_controller.rb
fofr: Strongly agree about content itself determining features, been wanting that for a while.

Keen not to focus on global layout and template though, prefer adaptable components that can handle whatever is given to them. End result ought to be similar.

app/controllers/concerns/content_navigation_ab_testable.rb
tijmenb: I think this confuses constants and frozen objects. 

Constants aren't actually constants in Ruby, and you can change them (albeit with the warning you're seeing).

Unrelatedly objects in Ruby can also be frozen, which means you can't mutate them.

```

PR #529, opened 2017-11-08
Published dates component

vanitabarrett:
Published dates component to be used within metadata block. 

screen shot 2017-11-08 at 17 47 05

**Also links to expandable full page history**
screen shot 2017-11-08 at 17 48 26

Trello Story: https://trello.com/c/UTBixmPW/85-convert-published-dates-into-a-component-for-universalnonav-variant

Review app component guide:
https://government-frontend-pr-529.herokuapp.com/component-guide/published-dates

Visual regression results:
https://government-frontend-pr-529.surge.sh/gallery.html

### DIFF
--- a/app/views/components/_published-dates.html.erb
+++ b/app/views/components/_published-dates.html.erb
@@ -8,14 +8,14 @@
 <% if published || last_updated %>
 <div class="app-c-published-dates" <% if history%>id="history" data-module="toggle"<% end %>>
   <% if published %>
-    Published <%= published %>
+    <%= t('components.published_dates.published', date: published) %>
   <% end %>
   <% if last_updated %>
-    <% if published %><br /><% end %><%= t('common.last_updated', date: last_updated) %>
+    <% if published %><br /><% end %><%= t('components.published_dates.last_updated', date: last_updated) %>
     <% if link_to_history && history.empty? %>
-      &mdash; <a href="#history"><%= t('common.see_all_updates') %></a>
+      &mdash; <a href="#history"><%= t('components.published_dates.see_all_updates') %></a>
     <% elsif history.any? %>
-      <a href="#full-history" data-controls="full-history" data-expanded="false">&#43;&nbsp;<%= t('common.full_page_history') %></a>
+      <a href="#full-history" data-controls="full-history" data-expanded="false">&#43;&nbsp;<%= t('components.published_dates.full_page_history') %></a>
       <div class="app-c-published-dates__change-history js-hidden" id="full-history">
         <ol>
           <% history.each do |change| %>
--- a/app/views/components/_published-dates.html.erb
+++ b/app/views/components/_published-dates.html.erb
@@ -6,7 +6,7 @@
   link_to_history ||= false
 %>
 
-<div class="app-c-published-dates" <% if history%>id="history"<% end %>>
+<div class="app-c-published-dates" <% if history%>id="history" data-module="toggle"<% end %>>
   Published <%= published %>
   <% if last_updated %>
     <% if published %><br /><% end %>
@@ -28,16 +28,3 @@
     <% end %>
   <% end %>
 </div>
-
-<% if history %>
-  <div class="app-c-published-dates__page-history" id="full-history">
-    <ol>
-      <% history.each do |change| %>
-        <li class="app-c-published-dates__change-item">
-          <time class="app-c-published-dates__change-date" datetime="<%= change[:timestamp] %>" class="timestamp"><%= change[:display_time] %></time>
-          <%= change[:note] %>
-        </li>
-      <% end %>
-    </ol>
-  </div>
-<% end %>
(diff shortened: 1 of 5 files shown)


app/views/components/_published-dates.html.erb
```diff
@@ -0,0 +1,30 @@
...
+<div class="app-c-published-dates" <% if history%>id="history" data-module="toggle"<% end %>>
+  <% if published %>
+    Published <%= published %>
+  <% end %>
+  <% if last_updated %>
+    <br />Last updated <%= last_updated %>
+    <% if link_to_history && !history %>
+      &mdash; <a href="#history">see all updates</a>
+    <% elsif history %>
+      <a href="#full-history" data-controls="full-history" data-expanded="false">&#43; full page history</a>
```
fofr: I noticed on mobile that the `+` wrapped to a new line. Might be good to include a `nbsp;` in `&#43; full`

app/views/components/_published-dates.html.erb
```diff
@@ -0,0 +1,30 @@
+<%
+  published ||= false
+  history ||= false
+  last_updated ||= false
+  link_to_history ||= false
+%>
+
+<div class="app-c-published-dates" <% if history%>id="history" data-module="toggle"<% end %>>
```
fofr: Should the block be wrapped in an `if published || last_updated` so that if nothing is passed in nothing renders?

app/views/components/_published-dates.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <% if published %>
+    Published <%= published %>
+  <% end %>
+  <% if last_updated %>
+    <br />Last updated <%= last_updated %>
```
fofr: If there's no published date this will include an unneeded last updated date. Sadly this can happen on bulk imported specialist documents (like old AAIB reports) – on these documents we show when it was updated but suppress published date.

fofr: Might be worth adding an example showing this too.

app/views/components/docs/published-dates.yml
```diff
@@ -0,0 +1,36 @@
...
+      last_updated: 20th October 2016
+      link_to_history: true
+  display_page_history:
+    description: This will set up an expandable section on the page to let users toggle the display of the page history.
+    data:
+      published: 1st January 1990
+      last_updated: 20th October 2016
+      history:
+      - display_time: 1st January 1990
+        note: First published
```
fofr: Should timestamp also be included here?

test/components/published_dates_test.rb
```diff
@@ -0,0 +1,62 @@
+require 'component_test_helper'
+
```
maxgds: I noticed that other tests check that without input the component doesn't render at all - at the moment yours will render the outer div which is probably undesirable. I tweaked mine to avoid this, you can see my approach here: https://github.com/alphagov/government-frontend/blob/metadata-component/app/views/components/_content-metadata.html.erb - would be good to get @fofr to give an opinion on that approach and apply that or something similar here, and then test for it.

app/views/components/docs/published-dates.yml
```diff
@@ -0,0 +1,36 @@
...
+    - be usable with a keyboard
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      published: 1st January 1990
+  with_last_updated_date:
+    data:
+      published: 1st January 1990
```
fofr: I wonder if these dates should also have timestamps

app/views/components/docs/published-dates.yml
```diff
@@ -0,0 +1,36 @@
+name: Published Dates
```
fofr: Pedant: Can it be "Published dates" not "Published Dates"

app/views/components/docs/published-dates.yml
```diff
@@ -0,0 +1,36 @@
...
+examples:
+  default:
+    data:
+      published: 1st January 1990
+  with_last_updated_date:
+    data:
+      published: 1st January 1990
+      last_updated: 20th October 2016
+  link_to_page_history:
+    description: This will set up a link to a '#history' anchor on the page - this can either be your own defined anchor or another instance of this component, see [Display Page History example](http://government-frontend.dev.gov.uk/component-guide/published-dates/display_page_history)
```
maxgds: I think your example url should be relative.

app/views/components/docs/published-dates.yml
```diff
@@ -0,0 +1,42 @@
...
+      published: 1st January 1990
+  just_last_updated_date:
+    data:
+      last_updated: 20th October 2016
+  with_last_updated_date:
+    data:
+      published: 1st January 1990
+      last_updated: 20th October 2016
+  link_to_page_history:
+    description: This will set up a link to a '#history' anchor on the page - this can either be your own defined anchor or another instance of this component, see [Display Page History example](/component-guide/published-dates/display_page_history)
```
fofr: Doesn't look like the anchor can be altered? I think that's ok though – just this description that needs updating.

PR #530, opened 2017-11-09
Add GA tracking to radio buttons

leenagupte:
Follows on from PR #524 
Trello card: https://trello.com/c/YQOcvmue

## Motivation

Once the SelfAssessmentSignTest A/B test had been deployed to Production we realised that the radio button events aren't being tracked.

We've added a new radio button javascript to do this.

Adding Javascript caused the all of the Rails Minitest tests to fail with Poltergeist errors. These errors prevented us from adding the script to put the correct focus on the error message yesterday (08/11/2017) before the SelfAssessmentSignTest A/B test was deployed to Production. 

When we realised we had to do something about the Javascript errors to enable radio button tracking, we decided to put the code to add focus to the error messages back in. 

[x] Tested in integration


test/test_helper.rb
```diff
@@ -46,6 +46,11 @@ class ActionDispatch::IntegrationTest
 
   def setup
     stub_shared_component_locales
+    # FIXME
+    # We need to disable errors for JavaScript since Static is mocked in our tests.
+    # and jQuery is pulled in from Static, this means we're not currently testing
+    # webchat, this should be resolved even when we remove this after the AB test.
+    page.driver.browser.js_errors = false
```
fofr: Will this break any accessibility errors that the component guide raises?
The component guide raises JS errors when it runs the aXe suite if it finds accessibility problems.

fofr: (We are building components at the moment and will rely on these tests working)

leenagupte: @fofr I'm under the impression that this setting only suppresses errors from the Minitest tests. Will double check with @nickcolley

nickcolley: I think so yes, maybe we could only include the javascript on the interstitial page, and only turn the errors off on that page.

I'm not happy with most of the approach here, unfortunately we have a fixed deadline since we can only test within an agreed window of time based on conversations with HMRC etc

leenagupte:
@fofr thanks for the review, updated this and now this JavaScript only runs on the b variant, removing the need to disable the JavaScript errors globally.

app/assets/javascripts/self_assessment_b_variant/radio-button-tracker.js
```diff
@@ -0,0 +1,35 @@
+;(function (global, document) {
+  'use strict'
+
+  var $ = global.jQuery
+  var GOVUK = global.GOVUK || {}
+
+  GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {}
+  GOVUK.analyticsPlugins.radioButtonTracker = function ($element) {
```
fofr: I'd have written this more like https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/track-click.js

This would:
* avoid boilerplate instantiation/scoping
* follow similar patterns to other JS modules

It's not really an analytics plugin, but a method for tracking events.

fofr: As this is temporary and needs to be merged swiftly, I've approved.

nickcolley: This was modelled after the existing 'External link tracker'

fofr: The external link tracker applies to every link on a page and couldn't be written in a modular way using data attributes. eg `$('body').on('click', externalLinkSelector, trackClickEvent)`

PR #532, opened 2017-11-09
Add a guard to Self Assessment A/B analytics and refactor

leenagupte:
## Motivation

The Self Assessment A/B test was written very quickly and didn't leave the code base in a very healthy state. This PR attempts to address this.

The main changes made here are:

1. Add a guard to the analytics meta tag for the Self Assessment A/B test in application.html.erb
2. Move the set of the A/B test into a concern.
3. Move the controller methods added for the new GET and POST routes in the A/B test to their own controller.

These changes should make it easier to remove this code when the time comes and make it easier for other A/B tests to be added whilst the Self Assessment A/B test is still running.


nickcolley:
Talked to Leena, and we think it's a good idea to do a good check that we didn't break the existing AB test on staging.

And also talk to the other AB people and make sure we didnt break their tests.

app/controllers/content_items_controller.rb
```diff
@@ -61,6 +62,16 @@ def sign_in_options
     end
   end
 
+  def self_assessment_start_page?(content_item)
+    self_assessment_pages = %w(
+      /log-in-file-self-assessment-tax-return
+      /log-in-file-self-assessment-tax-return/choose_sign_in
```
emmabeynon: Should this be `/log-in-file-self-assessment-tax-return/choose-sign-in`?

leenagupte: Yes! Thank you. Will fix.

PR #534, opened 2017-11-10
Add content metadata component

maxgds:
- Presents publisher metadata
- Passes params to published date component to display published and updated metadata

Review app component guide:
https://government-frontend-pr-534.herokuapp.com/component-guide

### DIFF
--- a/app/assets/stylesheets/components/_content-metadata.scss
+++ b/app/assets/stylesheets/components/_content-metadata.scss
@@ -1,0 +1,16 @@
+.app-c-content-metadata {
+  @include responsive-bottom-margin;
+  @include core-16;
+  padding-top: $gutter-half;
+  padding-bottom: $gutter-half;
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+}
+
+.app-c-content-metadata__from {
+  margin-top: $gutter-half;
+}
+
+.app-c-content-metadata__from a {
+  font-weight: bold;
+}
--- a/app/views/components/_content-metadata.html.erb
+++ b/app/views/components/_content-metadata.html.erb
@@ -1,0 +1,21 @@
+<%
+  published ||= false
+  last_updated ||= false
+  from ||= false
+  link_to_history ||= false
+  metadata = [published, last_updated, from]
+%>
+<% if metadata.any? %>
+  <div class="app-c-content-metadata">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <%= render 'components/published-dates', published: published, last_updated: last_updated, link_to_history: link_to_history %>
+        <% if from %>
+          <div class="app-c-content-metadata__from">
+            <%= t("govuk_component.content-metadata.from", default: "From") %>: <%= from.to_sentence.html_safe %>
+          </div>
+        <% end %>
+      </div>
+    </div>
+  </div>
+<% end %>
--- a/app/views/components/docs/content-metadata.yml
+++ b/app/views/components/docs/content-metadata.yml
@@ -1,0 +1,34 @@
+name: Content Metadata
+description: Lists publication dates and publishers.
+shared_accessibility_criteria:
+- link
+
+examples:
+  default:
+    data:
+      from:
+        - <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+  with_history_link:
+    data:
+      from:
+        - <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+      link_to_history: true
+  two_publishers:
+    data:
+      from:
+        - <a href='/government/organisations/department-for-education'>Department for Education</a>
+        - <a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+  multiple_publishers_and_no_link_to_history:
+    data:
+      from:
+        - <a href='/government/organisations/department-for-education'>Department for Education</a>
+        - <a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>
+        - <a href='/government/organisations/department-for-work-pensions'>Department for Work and Pensions</a>
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -9,7 +9,6 @@
     <% end %>
   </div>
 </div>
-
 <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
 <%= render 'shared/metadata', content_item: @content_item %>
 <%= render 'shared/history_notice', content_item: @content_item %>
--- a/test/components/content_metadata_test.rb
+++ b/test/components/content_metadata_test.rb
@@ -1,0 +1,46 @@
+require 'component_test_helper'
+
+class ContentMetadataTest < ComponentTestCase
+  def component_name
+    "content-metadata"
+  end
+
+  test "does not render metadata when no data is given" do
+    assert_empty render_component({})
+  end
+
+  test "renders a from link when from data is given" do
+    render_component(from: ["<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"])
+    assert_select ".app-c-content-metadata__from a[href='/government/organisations/ministry-of-defence']", text: 'Ministry of Defence'
+    assert_select ".app-c-content-metadata__from", text: 'From: Ministry of Defence'
+  end
+
+
+  test "renders two from links when two publishers are given" do
+    render_component(from: ["<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>", "<a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>"])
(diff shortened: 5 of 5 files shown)


app/assets/stylesheets/components/_content-metadata.scss
```diff
@@ -0,0 +1,19 @@
+
```
vanitabarrett: Stray new line

app/assets/stylesheets/components/_content-metadata.scss
```diff
@@ -0,0 +1,19 @@
...
+  padding-bottom: $gutter-half;
+
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+}
+
+.app-c-content-metadata__from {
+  margin-top: $gutter-half;
+
+  a {
```
vanitabarrett: In previous components we've tended to avoid type selectors like this and replaced with a class, e.g: `.app-c-content-metadata__from-link`. Also tend to avoid nesting BEM selectors.

app/views/components/_content-metadata.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  published ||= false
+  last_updated ||= false
+  from ||= false
+  metadata = [published, last_updated, from]
+%>
+<% if metadata.any? %>
+  <div class="app-c-content-metadata">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <%= render 'components/published-dates', published: published, last_updated: last_updated %>
```
vanitabarrett: This doesn't allow for a published dates component with [link_to_history](https://government-frontend.herokuapp.com/component-guide/published-dates/link_to_page_history) enabled

app/views/content_items/detailed_guide.html.erb
```diff
@@ -9,7 +9,7 @@
     <% end %>
   </div>
 </div>
-
+<%= @content_item.updated.inspect %>
```
vanitabarrett: 👀

app/views/components/docs/content-metadata.yml
```diff
@@ -0,0 +1,14 @@
...
+shared_accessibility_criteria:
+- link
+
+examples:
+  default:
+    data:
+      from:  ["<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"]
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+  multiple_publishers:
```
vanitabarrett: Can we add another example with more than 2? Perhaps that would involve renaming this to `two_publishers`

vanitabarrett: This may be personal preference, but I'd also include the published dates in these examples to see how all the elements of the component work together. You could always include a separate example, e.g `no_published_dates` to show an example on it's own.

test/components/content_metadata_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'component_test_helper'
+
+class ContentMetaDataTest < ComponentTestCase
```
vanitabarrett: Nitpick - should be `ContentMetadataTest`

test/components/content_metadata_test.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  test "renders a from link when from data is given" do
+    render_component(from: ["<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"])
+    assert_select ".app-c-content-metadata__from a[href='/government/organisations/ministry-of-defence']", text: 'Ministry of Defence'
+    assert_select ".app-c-content-metadata__from", text: 'From: Ministry of Defence'
+  end
+
+  test "renders a sentence when multiple publishers are given" do
+    render_component(from: ["<a href='/government/organisations/department-for-education'>Department for Education</a>", "<a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>"])
+    assert_select ".app-c-content-metadata__from", text: 'From: Department for Education and Education Funding Agency'
```
vanitabarrett: Can we also assert that two links are rendered with the correct href?

app/assets/stylesheets/components/_content-metadata.scss
```diff
@@ -0,0 +1,19 @@
+
+.app-c-content-metadata {
+  @include responsive-bottom-margin;
+  @include core-16;
+
```
vanitabarrett: Don't think we need these spaces (see one below too)

app/views/components/_content-metadata.html.erb
```diff
@@ -0,0 +1,18 @@
+<%
+  published ||= false
+  last_updated ||= false
+  from ||= false
+  metadata = [published, last_updated, from]
+%>
+<% if metadata.any? %>
+  <div class="app-c-content-metadata">
+    <div class="grid-row">
```
vanitabarrett: Not sure how the metadata is used on pages, but at the moment it looks like it's all inside this grid/column? Does that mean we could take it out of the component and wrap the component inside it when needed in the layouts?

app/views/components/docs/content-metadata.yml
maxgds: I think I need to double check the expected behaviour here - I thought that this component was not going to expand the dates and that was a possibility elsewhere on the page - however the screenshot on the trello card shows otherwise.

app/views/content_items/detailed_guide.html.erb
maxgds: oops

app/views/components/docs/content-metadata.yml
vanitabarrett: I don't think this component need to expand the dates, but it should be able to contain a link to the #history anchor on the page (which will expand the dates).

test/components/content_metadata_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'component_test_helper'
+
+class ContentMetaDataTest < ComponentTestCase
+  def component_name
+    "content-metadata"
+  end
+
+  test "fails to render metadata when no data is given" do
```
fofr: This isn't a failure if it's expected. eg "renders nothing when no data given"

app/views/components/docs/content-metadata.yml
```diff
@@ -0,0 +1,14 @@
+name: Content Metadata
+description: Lists publication dates and publishers.
+shared_accessibility_criteria:
+- link
+
+examples:
+  default:
+    data:
+      from:  ["<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"]
```
fofr: Arrays in yml are usually presented like:
```
from:
  - "<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"
  - "<a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>"
```

app/views/components/docs/content-metadata.yml
```diff
@@ -0,0 +1,14 @@
...
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+  multiple_publishers:
+    data:
+      from: ["<a href='/government/organisations/department-for-education'>Department for Education</a>", "<a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>"]
```
fofr: File needs a new line. Your editor can be configured to add this automatically.

app/views/components/_content-metadata.html.erb
fofr: The grey borders need to run full width. So I think having the internal grid is ok.

app/assets/stylesheets/components/_content-metadata.scss
maxgds: I've un-nested it. I'd like to address this at some point but it means tweaking a presenter that doesn't currently support taking a class for the link_to magic so I'm leaving an a tag in the css for now with a comment.

app/assets/stylesheets/components/_content-metadata.scss
```diff
@@ -0,0 +1,16 @@
...
+  padding-bottom: $gutter-half;
+  border-top: 1px solid $border-colour;
+  border-bottom: 1px solid $border-colour;
+}
+
+.app-c-content-metadata__from {
+  margin-top: $gutter-half;
+}
+
+.app-c-content-metadata__from-link a {
```
fofr: As this wraps multiple links, I'd use `.app-c-content-metadata__from-links`, but I think what was there before is still effectively the same as this, just less verbose.

maxgds: My plan was to apply that class to the links until I found it was a whole chain of magic to produce the links. Have reverted it now but un-embedded the anchor class. Hopefully can find time to work out passing through the class and improve this at some point.

app/views/components/_content-metadata.html.erb
nickcolley: I don't think this is a blocker but I think this'd make more sense without and using `max-width` instead.

This grid is being used to make sure the [line length](https://en.wikipedia.org/wiki/Line_length) a good length to make the content legible.

app/views/components/docs/content-metadata.yml
```diff
@@ -0,0 +1,34 @@
...
+    data:
+      from:
+        - <a href='/government/organisations/ministry-of-defence'>Ministry of Defence</a>
+      published: "31 July 2017"
+      last_updated: "20 September 2017"
+      link_to_history: true
+  two_publishers:
+    data:
+      from:
+        - <a href='/government/organisations/department-for-education'>Department for Education</a>,
```
fofr: There's some trailing commas at the end of a few lines, these need to be removed. Otherwise commas show up i n the examples.

screen shot 2017-11-14 at 11 07 50

fofr: screen shot 2017-11-14 at 11 08 33

PR #536, opened 2017-11-13
Enable right-to-left text direction for Farsi content

emmabeynon:
We've received a [Zendesk ticket](https://govuk.zendesk.com/agent/tickets/2512759) informing us that our Farsi/Persian content is currently left-aligned, however it is a language that reads right-to-left and so should be right-aligned.  This is already in place for other "right-to-left" languages such as Hebrew (see [example](https://www.gov.uk/government/speeches/ambassador-quarreys-speech-for-international-holocaust-remembrance-day.he)).

This PR sets the direction of Farsi content to `rtl` in its locale YAML file to
enable correct alignment.

### Before

screen shot 2017-11-13 at 16 37 03

### After

screen shot 2017-11-13 at 16 36 46

### DIFF
--- a/config/locales/fa.yml
+++ b/config/locales/fa.yml
@@ -1,4 +1,6 @@
 fa:
+  i18n:
+    direction: rtl
   language_names:
     fa: "فارسی"
   content_item:


sihugh:
It feels like we might need to pass this Zendesk ticket onto a content team who might know about other RTL languages.

PR #537, opened 2017-11-14
Fix link to history behaviour

vanitabarrett:
Component was incorrectly applying the history id to the component each time, as the check for history should have been checking for history.any?

Updated to include this check, component now only contains history id if a history item has been passed to it.

Review app component guide:
https://government-frontend-pr-537.herokuapp.com/component-guide/published-dates

Visual regression results:
https://government-frontend-pr-537.surge.sh/gallery.html

### DIFF
--- a/test/components/published_dates_test.rb
+++ b/test/components/published_dates_test.rb
@@ -35,6 +35,18 @@ class PublishedDatesTest < ComponentTestCase
     assert_select ".app-c-published-dates .app-c-published-dates__change-date", text: "23 August 2013"
   end
 
+  test "only adds history id when passed page history" do
+    render_component(published: "1st November 2000")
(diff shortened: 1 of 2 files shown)


fofr:
Can you add a test for this too? Otherwise :+1:

test/components/published_dates_test.rb
```diff
@@ -35,6 +35,18 @@ def component_name
     assert_select ".app-c-published-dates .app-c-published-dates__change-date", text: "23 August 2013"
   end
 
+  test "only adds history id when passed page history" do
+    render_component(published: "1st November 2000")
+    assert_select ".app-c-published-dates#history", false, "should only render history id if passed history item"
```
fofr: Minor: Selector `.app-c-published-dates#history` only needs to be `#history`

PR #539, opened 2017-11-15
Add national applicability to universal navigation variant

fofr:
Allow content metadata to render any metadata types
* Use pattern from static metadata component:
https://govuk-static.herokuapp.com/component-guide/metadata
* Allow further metadata to be added easily as design is iterated
* This allows us to easily add national applicability

Add national applicability to content metadata
* Move content_metadata params into metadata presenter module
* Decorate with “Applies to” when National Applicability applies
* Add “Applies to” to the beginning of the hash to appear before “From”
* Use existing methods for generating “Applies to” from content item
* Improve rendering of component when no dates provided

Review app component guide:
https://government-frontend-pr-539.herokuapp.com/component-guide/content-metadata

screen shot 2017-11-15 at 09 52 29

### DIFF
--- a/app/views/components/docs/content-metadata.yml
+++ b/app/views/components/docs/content-metadata.yml
@@ -57,3 +57,7 @@ examples:
         Another: []
         "Empty thing": false
         "Also empty": {}
+  national_applicability:
+    data:
+      other:
+        Applies to: "England, Scotland, and Wales (see detailed guidance for <a href=\"http://www.dardni.gov.uk/news-dard-pa022-a-13-new-procedure-for\" rel=\"external\">Northern Ireland</a>)"
--- a/app/views/components/docs/content-metadata.yml
+++ b/app/views/components/docs/content-metadata.yml
@@ -10,17 +10,19 @@ shared_accessibility_criteria:
 examples:
   default:
     data:
-      from:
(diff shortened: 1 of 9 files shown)


app/views/components/docs/content-metadata.yml
```diff
@@ -31,16 +33,31 @@ examples:
...
-        - <a href='/government/organisations/department-for-work-pensions'>Department for Work and Pensions</a>
       published: 31 July 2017
       last_updated: 20 September 2017
+      other:
+        From:
+          - <a href='/government/organisations/department-for-education'>Department for Education</a>
+          - <a href='/government/organisations/education-funding-agency'>Education Funding Agency</a>
+          - <a href='/government/organisations/department-for-work-pensions'>Department for Work and Pensions</a>
+  nothing:
+    description: |
```
maxgds: Should this be an example or just a line in the intro to the component? I can see why you'd consider the fail state and test against it but specifically presenting an example seems counter intuitive to me.

maxgds:
Quite an overhaul! Makes sense, though. Just one query around the examples given in the guide but it otherwise looks good.

app/views/components/docs/content-metadata.yml
fofr: I think it's appropriate to keep to avoid the risk of future regressions.

PR #542, opened 2017-11-17
Create navigation sidebar component

vanitabarrett:
screen shot 2017-11-17 at 14 51 51

screen shot 2017-11-17 at 14 52 09

screen shot 2017-11-17 at 14 52 51

This PR only covers making the navigation sidebar component production-ready, it doesn't deal with any of the logic that's needed to get the data in the right format to be passed to the component within templates.

Review app component guide:
https://government-frontend-pr-542.herokuapp.com/component-guide/navigation-sidebar

Visual regression results:
https://government-frontend-pr-542.surge.sh/gallery.html

### DIFF
--- a/app/assets/stylesheets/components/_navigation-sidebar.scss
+++ b/app/assets/stylesheets/components/_navigation-sidebar.scss
@@ -3,7 +3,7 @@
 }
 
 .app-c-navigation-sidebar__main-heading {
-  margin-top: $gutter;
+  margin-top: $gutter-half;
   margin-bottom: $gutter-one-third;
   @include bold-19;
 }
--- a/app/assets/stylesheets/components/_navigation-sidebar.scss
+++ b/app/assets/stylesheets/components/_navigation-sidebar.scss
@@ -1,0 +1,33 @@
+.app-c-navigation-sidebar {
+  border-top: 2px solid $govuk-blue;
+}
+
+.app-c-navigation-sidebar__main-heading {
+  margin-top: $gutter;
+  margin-bottom: $gutter-one-third;
+  @include bold-19;
+}
+
+.app-c-navigation-sidebar__sub-heading {
+  border-top: 1px solid $border-colour;
+  padding-top: $gutter-half;
+}
+
+.app-c-navigation-sidebar__nav-section {
+  margin-top: $gutter-one-third;
+  margin-bottom: $gutter;
+}
+
+.app-c-navigation-sidebar__link {
+  list-style-type: none;
+  margin-bottom: $gutter-one-third;
+}
+
+.app-c-navigation-sidebar__related-link {
+  text-decoration: none;
+}
+
+.app-c-navigation-sidebar__section-link {
+  text-decoration: none;
+  @include bold-19;
+}
--- a/app/assets/stylesheets/components/_navigation-sidebar.scss
+++ b/app/assets/stylesheets/components/_navigation-sidebar.scss
@@ -13,6 +13,10 @@
   padding-top: $gutter-half;
 }
 
+.app-c-navigation-sidebar__external-heading {
+  @include bold-19;
+}
+
 .app-c-navigation-sidebar__nav-section {
   margin-top: $gutter-one-third;
   margin-bottom: $gutter;
@@ -23,11 +27,20 @@
   margin-bottom: $gutter-one-third;
 }
 
-.app-c-navigation-sidebar__related-link {
+.app-c-navigation-sidebar__related-link,
+.app-c-navigation-sidebar__external-related-link {
   text-decoration: none;
 }
 
 .app-c-navigation-sidebar__section-link {
   text-decoration: none;
   @include bold-19;
+}
+
+// reset the default browser styles
+.app-c-navigation-sidebar__link-list {
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  margin-bottom: 1.25em;
 }
--- a/app/helpers/navigation_sidebar_helper.rb
+++ b/app/helpers/navigation_sidebar_helper.rb
@@ -1,0 +1,11 @@
+module NavigationSidebarHelper
+  def sidebar_link_to(link, section)
+    if section === "related_items"
+      link_to link[:text], link[:href], class: "app-c-navigation-sidebar__related-link"
+    elsif section === "external_links"
+      link_to link[:text], link[:href], class: "app-c-navigation-sidebar__external-related-link", rel: "external"
+    else
+      link_to link[:text], link[:href], class: "app-c-navigation-sidebar__section-link"
+    end
+  end
+end
--- a/app/views/components/_navigation-sidebar.html.erb
+++ b/app/views/components/_navigation-sidebar.html.erb
@@ -4,38 +4,47 @@
   publishers ||= []
   collections ||= []
   policies ||= []
+  external_links ||= []
   related_content = [
     {"related_items" => related_items},
     {"topics" => topics},
     {"publishers" => publishers},
     {"collections" => collections},
-    {"policies" => policies}
+    {"policies" => policies},
+    {"external_links" => external_links}
   ]
 %>
 <% if related_content.map(&:values).flatten.any? %>
   <aside class="app-c-navigation-sidebar" role="complementary">
     <% related_content.each do |section| %>
-      <% section.select do |section_title, links| %>
+      <% section.each do |section_title, links| %>
 
         <% if section_title === "related_items" %>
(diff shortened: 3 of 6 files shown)


app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    {"publishers" => publishers},
+    {"collections" => collections},
+    {"policies" => policies}
+  ]
+%>
+<% if related_content.map(&:values).flatten.any? %>
+  <aside class="app-c-navigation-sidebar" role="complementary">
+    <% related_content.each do |section| %>
+      <% section.select do |section_title, links| %>
+
```
maxgds: Lines of whitespace look a little random.

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,48 @@
...
+              <% links.each do |link| %>
+              <li class="app-c-navigation-sidebar__link">
+                <%= link_to link[:text], link[:href], class: "app-c-navigation-sidebar__related-link" %>
+              </li>
+              <% end %>
+            </ul>
+          </nav>
+
+        <% elsif links.any? %>
+          <h3 id="nav-sidebar-<%= section_title %>" class="app-c-navigation-sidebar__sub-heading"><%= t('components.navigation_sidebar.' + section_title) %></h3>
```
maxgds: Should we give the different subheading/link blocks classes related to what they are instead of generic ones (__topic, __from etc)? I'm unsure - I feel compelled to do it but since they all look the same perhaps it's unnecessary clutter.

vanitabarrett: I did think about this, but ended up giving the same classes seeing as they're all styled the same (apart from the related content section at the top).

app/views/components/_navigation-sidebar.html.erb
vanitabarrett: Hmm, I put these in to break up the code into the 2 separate `if` statements, as otherwise I found it a bit difficult to read/follow the logic. Happy to remove if you think it's fine without the spacing though

maxgds: I think it's ok without. If you do leave it in then I'd make it more symmetrical so you have an opening and a closing space around the thing you're making more readable.

app/helpers/navigation_sidebar_helper.rb
```diff
@@ -0,0 +1,11 @@
+module NavigationSidebarHelper
+  def construct_sidebar_link(link, section)
```
steventux: It would be good to follow the naming pattern of the underlying Rails tag helper
Maybe something like `sidebar_link_to` ?
Alternatively as the link `:class` and `:rel` attrs are the only thing that varies you could use `link_to` and a helper method for the attributes.

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    {"publishers" => publishers},
+    {"collections" => collections},
+    {"policies" => policies},
+    {"external_links" => external_links}
+  ]
+%>
+<% if related_content.map(&:values).flatten.any? %>
+  <aside class="app-c-navigation-sidebar" role="complementary">
+    <% related_content.each do |section| %>
+      <% section.select do |section_title, links| %>
```
steventux: Is `section` a Hash? If so wouldn't `section.each` be more obvious? Nothing is being selected by the `select` method.

app/views/components/docs/navigation-sidebar.yml
```diff
@@ -0,0 +1,130 @@
...
+  - Should have a role of 'complementary' as it complements the main page content
+  - Should have a role of 'navigation' on any navigation elements inside the component
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      related_items:
+        - text: Find an apprenticeship
+          href: /apply-apprenticeship
```
andysellick: Genuine question - should we be using 'href' or 'path' for link href attributes? I was sure href was more appropriate but remember being told to use path at some point, so I've been using it since. What's the consensus?

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,50 @@
...
+              <% links.each do |link| %>
+              <li class="app-c-navigation-sidebar__link">
+                <%= sidebar_link_to(link, section_title) %>
+              </li>
+              <% end %>
+            </ul>
+          </nav>
+
+        <% elsif links.any? %>
+          <h3 id="nav-sidebar-<%= section_title %>" class="app-c-navigation-sidebar__sub-heading <%= 'app-c-navigation-sidebar__external-heading' if section_title == "external_links" %>"><%= t('components.navigation_sidebar.' + section_title) %></h3>
```
andysellick: Nitpick: this could be broken into more than one line to make it more readable. Same goes for the H2 above.

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,50 @@
...
+%>
+<% if related_content.map(&:values).flatten.any? %>
+  <aside class="app-c-navigation-sidebar" role="complementary">
+    <% related_content.each do |section| %>
+      <% section.each do |section_title, links| %>
+
+        <% if section_title === "related_items" %>
+          <h2 id="nav-sidebar-<%= section_title %>" class="app-c-navigation-sidebar__main-heading" ><%= t('components.navigation_sidebar.related_content') %></h2>
+          <nav role="navigation" class="app-c-navigation-sidebar__nav-section" aria-labelledby="nav-sidebar-<%= section_title %>" %>
+            <ul>
```
andysellick: Debatable gotcha: you've not styled the UL element. Had a problem with a component recently where it found itself on a page with no CSS reset, so the UL was automatically given a margin left by Chrome. Since your component is in government-frontend this shouldn't be a problem, but maybe worth considering for future proofing.

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,50 @@
...
+              <% links.each do |link| %>
+              <li class="app-c-navigation-sidebar__link">
```
andysellick: Nitpick: indentation.

app/assets/stylesheets/components/_navigation-sidebar.scss
```diff
@@ -0,0 +1,38 @@
...
+  border-top: 2px solid $govuk-blue;
+}
+
+.app-c-navigation-sidebar__main-heading {
+  margin-top: $gutter-half;
+  margin-bottom: $gutter-one-third;
+  @include bold-19;
+}
+
+.app-c-navigation-sidebar__sub-heading {
```
andysellick: Presumably you're just keeping the existing design here but is it odd that these headings aren't really styled at all? It's strange that they have a smaller font size than the links they're a heading for.

app/views/components/_navigation-sidebar.html.erb
```diff
@@ -0,0 +1,50 @@
...
+                <%= sidebar_link_to(link, section_title) %>
+              </li>
+              <% end %>
+            </ul>
+          </nav>
+
+        <% elsif links.any? %>
+          <h3 id="nav-sidebar-<%= section_title %>" class="app-c-navigation-sidebar__sub-heading <%= 'app-c-navigation-sidebar__external-heading' if section_title == "external_links" %>"><%= t('components.navigation_sidebar.' + section_title) %></h3>
+          <nav role="navigation" class="app-c-navigation-sidebar__nav-section" aria-labelledby="nav-sidebar-<%= section_title %>">
```
andysellick: This aria-labelledby is really great when using a screenreader. Might be good to have a test for this?

app/assets/stylesheets/components/_navigation-sidebar.scss
vanitabarrett: This is a change from the existing [related links component](https://govuk-static.herokuapp.com/component-guide/related_items) we have, but it follows @alextea 's design for this component.

app/views/components/_navigation-sidebar.html.erb
vanitabarrett: Good point - will add that in 👍

app/views/components/docs/navigation-sidebar.yml
vanitabarrett: I know there's been some debate on this, I thought the conclusion was to use 'href' as that's what we've used in the Back to Top link, Print link, and Download link components. Happy to change if we decide that 'path' is better though

cc @nickcolley

PR #544, opened 2017-11-20
Remove self assessment ab test

leenagupte:
Trello: https://trello.com/c/xdmgIGIO

The Self Assessment A/B test was switched off on 17/11/2017. This is the clean up to remove the code from the frontend.


leenagupte:
@emmabeynon @nickcolley The only thing I'm concerned about is that when I ran it locally, the styling for the buttons was missing. They looked like regular links.

nickcolley:
@leenagupte here? https://government-frontend-pr-544.herokuapp.com/log-in-file-self-assessment-tax-return

leenagupte:
@nickcolley No, just on my local machine. The buttons are displaying properly in the heroku app, so it looks like it's all good.

PR #547, opened 2017-11-22
Add a homepage for developers

tijmenb:
This adds a page at the root of the site that will be visible in development and on Heroku review apps.

It has the following things that would be useful in development:

- Links to the component guide
- The URLs of the content store, rummager, static, so it's clear which version we are using
- Link to the canonical example of a template (the one named after the schema)*
- Link to a random version of the template*
- List of links to pages rendered by government-frontend. I've reused the wraith config for this, it seemed like a nice representative list of different pages.

https://trello.com/c/IoHKyhoX

`*` will work after https://github.com/alphagov/govuk-content-schemas/pull/665 is merged and we point the Heroku review apps to the [example content store Heroku app](https://govuk-content-store-examples.herokuapp.com).

### DIFF
--- a/app/views/development/index.html.erb
+++ b/app/views/development/index.html.erb
@@ -1,0 +1,95 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>
+    government-frontend development page
+  </title>
+</head>
+<body>
+  <div id="wrapper">
+    <main class="govspeak">
+      <%= render 'govuk_component/title', title: 'government-frontend' %>
+
+      <% html = capture do %>
+        <p>
+          Hi, this is the root of the <%= link_to 'government-frontend', 'https://github.com/alphagov/government-frontend' %> application.
+          This page is intended to be shown in development and on Heroku review apps. It
+          won't be shown to users on GOV.UK because there are no routes to it.
+        </p>
+
+        <h2>Component guide</h2>
+
+        <p>
+          To see the available components, take a look at the <%= link_to 'component guide', '/component-guide' %>.
+        </p>
+
+        <h2>Configuration</h2>
+
+        <table>
+          <tr>
+            <td>Content store</td>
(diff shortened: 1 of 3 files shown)


nickcolley:
This is so awesome, great way to make this application more inclusive and easier to develop on :D

I'm getting not found errors for the content schema examples.

app/views/development/index.html.erb
```diff
@@ -0,0 +1,89 @@
...
+          <tr>
+            <td>Rummager (search)</td>
+            <td><%= link_to Plek.find('rummager'), Plek.find('rummager') %></td>
+          </tr>
+        </table>
+
+        <h2>Examples</h2>
+
+        <p>
+          The content schemas contain <%= link_to 'handmade examples', 'https://github.com/alphagov/govuk-content-schemas/tree/master/examples' %>
```
nickcolley: Could we also mention that we use these examples in our tests?

nickcolley: I missed this bit, wonder if we can put it in a call out or make it more obvious the links below will not work

tijmenb: I added a footnote to the links, hopefully clearer now.

h-lame:
This is a great idea.

> This adds a page at the root of the site that will be visible in development and on Heroku review apps.

I don't see anything that restricts this page - is that configured external to the app somewhere?

tijmenb:
@h-lame nope, it's just that nothing routes to it so it can't show up on GOV.UK (theoretically it would be available on https://government-frontend.publishing.service.gov.uk, but that's inside that frontend VDC).

FWIW smart answers also has a "homepage" that isn't visible:

https://github.com/alphagov/smart-answers/blob/598ec802d5848d6ef8085f929a42f16838aa9374/app/controllers/smart_answers_controller.rb#L21-L25

h-lame:
@tijmenb - that makes sense - I forgot about router.  Thanks for reminding me!

PR #548, opened 2017-11-22
[DO NOT MERGE] Fix preview layout for no whitelisted taxons

davidbasalla:
Use `GovukNavigationHelpers::ContentItem` to get the taxons for a content item. This enables us to check if the linked taxons belongs to the whitelisted root taxons from https://github.com/alphagov/govuk_navigation_helpers/pull/82.

Specifically this should fix an issue when Publisher would see a different layout between draft and live content, as well as different breadcrumbs.

### Live page:

screen shot 2017-11-22 at 16 58 34

### Page on Draft Stack - Before

screen shot 2017-11-22 at 16 58 45

### Page on Draft Stack - After

screen shot 2017-11-22 at 16 58 25

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -110,7 +110,7 @@ private
     if should_present_universal_navigation?
       request.variant = :universal_navigation
     else
-      @navigation = NavigationType.new(@content_item.content_item)
+      @navigation = NavigationType.new(@content_item)
       if @navigation.should_present_taxonomy_navigation?
         request.variant = :taxonomy_navigation
       end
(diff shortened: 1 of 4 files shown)


nickcolley:
Any updates on this?

cbaines:
> Any updates on this?

So I just had a look. The intent here was to improve the user experience on the draft stack, by getting Government Frontend on the draft stack to better reflect what the page would look like when published.

The two big differences are the page header, and the width of the content. The screenshots from David show the differences in page width, but not the difference in the header.

The page width issue I think has been helped by the layout changes in Government Frontend, I took some screenshots which are shown below. There is still a difference in the header, which I think this change would resolve. If I have time, I'll try rebasing these commits on master to resolve the merge conflicts.

# Draft

page-draft

# Live

page-live

cbaines:
The layout differences have now resolved themselves, probably due to either publishing all the taxons, or the "phase" related changes. Either way, I think we can abandon these changes now.

PR #549, opened 2017-11-23
[Do not merge] Add sticky contents link

maxgds:
@vanitabarrett @andysellick @nickcolley 
Could I get some comments on this please? This is a first pass at getting a "back to contents" link to work for the new layout. The actual styles are based on a glance at a stab Mia had, I am sure it needs tidying up but...

1) I'm struggling to understand why with js off the width is constrained differently to with js on.
2) I'm not sure of the correct conventions for applying this sort of css - at the moment I'm applying it only to specialist documents so that seemed the logical place but if I pull it out of there to apply to other templates where should it live? Similarly, should min-widths and colours be declared here or elsewhere?
3) I'm sure that hard coding 66.67 is wrong but what should I be doing?

Please constrain comments to the following files, the rest are Steve's changes (as are some of the specialist document view but you can see what I've applied to it):
app/assets/stylesheets/views/_specialist-document.scss
app/views/content_items/specialist_document.html.erb

Owing to Steve's changes a lot of tests are failing at the mo so there's no heroku app, sorry. 

screen shot 2017-11-23 at 15 25 42
Example test url:
https://government-frontend-pr-549.herokuapp.com//drug-device-alerts/intellivue-patient-monitors-used-with-12-lead-ecg-risk-of-ecg-trace-distortion-specific-models-and-software-versions-affected#contents-list


app/assets/stylesheets/views/_specialist-document.scss
```diff
@@ -1,3 +1,33 @@
+$contents-link-background: #dee0e2;
```
vanitabarrett: This colour is already defined in the colour styleguide as $panel-colour (http://govuk-elements.herokuapp.com/colour/)

app/views/content_items/specialist_document.html.erb
```diff
@@ -1,28 +1,38 @@
...
+        ) %>
+      <% end %>
+      <div class="responsive-bottom-margin">
+        <%= render 'components/published-dates', {
+              published: @content_item.published,
+              last_updated: @content_item.updated,
+              history: @content_item.history
+            } %>
+      </div>
+      <div class="contents-list-link-wrapper" data-sticky-element>
```
vanitabarrett: I feel like the styling applied to `contents-list-link-wrapper` means that it should be part of the back-to-top component.
Then we could just wrap the component in a div with the `data-sticky-element` on for the sticky behaviour?

app/assets/stylesheets/views/_specialist-document.scss
```diff
@@ -1,3 +1,33 @@
...
 .specialist-document {
   @include sidebar-with-body;
 }
+
+.govuk-sticky-element {
+  width: 100%;
+  margin: 0 $gutter-half;
+}
+
+.contents-list-link-wrapper {
```
vanitabarrett: I think we should be able to move this styling into the back-to-top component? Then it could live with the rest of the `app-c-back-to-top` styles.

app/assets/stylesheets/views/_specialist-document.scss
```diff
@@ -1,3 +1,33 @@
...
+  }
+}
+.app-c-back-to-top {
+  display: block;
+  margin: $gutter-half 0;
+}
+.app-c-back-to-top {
+    display: block;
+    margin: 15px 0 10px 0;
+}
```
vanitabarrett: You have 2 conflicting bit of css here - not sure which margin we want applying. But we can move this to `stylesheets/components/_back-to-top.scss`

app/views/content_items/specialist_document.html.erb
maxgds: I considered this but decided against it because it ties that particular use case into the component and makes it less flexible - at the moment you can use a back to top wherever you like and decorate it at will and I think that's worth preserving. I considered creating a second component that calls the existing one but it seems like a lot of overhead for essentially a single div wrapper. Also, I would imagine that a lot of the current templates will be collapsed into fewer - the 4 types that have contents lists may well end up in a single body partial, so we'd be looking at a component that only got used once. Happy to take further opinions on this though. @andysellick @nickcolley what are your thoughts?

app/assets/stylesheets/views/_specialist-document.scss
```diff
@@ -1,3 +1,33 @@
...
   @include sidebar-with-body;
 }
+
+.govuk-sticky-element {
+  width: 100%;
+  margin: 0 $gutter-half;
+}
+
+.contents-list-link-wrapper {
+  background-color: $contents-link-background;
```
andysellick: Need to look at restructuring these styles for when JS is disabled. At the moment it looks like this: 

screen shot 2017-11-24 at 08 59 48

I don't think it needs the background colour, width or padding in this situation.

nickcolley:
This feels off to me, but I don't really have much to say apart from there doesn't seem to be a compelling case to obstruct content for this.

Should this pattern follow the same as what modelling services are doing, where it's full width and has a border?

Should we be using `position: sticky`?

PR #553, opened 2017-11-28
Point to example content store

tijmenb:
This points Heroku apps to the example content store, so that we can use the examples and random content. For all other pages the app will keep working in the same way because it forwards to the real content store (https://github.com/alphagov/govuk-content-schemas/pull/665).

https://trello.com/c/IoHKyhoX

### DIFF
--- a/app.json
+++ b/app.json
@@ -12,7 +12,7 @@
       "value": "https://www.gov.uk"
     },
     "PLEK_SERVICE_CONTENT_STORE_URI": {
-      "value": "https://www.gov.uk/api"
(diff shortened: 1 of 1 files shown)


nickcolley:
We need to make sure that the heroku instance doesn't also have hardcoded environment variables (in the UI)

nickcolley:
@tijmenb looks like we do need to do that, but I think the heroku owner for that app is the heroku@ account?

tijmenb:
@nickcolley Done, works now: http://government-frontend.herokuapp.com/

PR #555, opened 2017-11-28
Document how to use the dummy content store locally

tijmenb:


tijmenb:
This is [already merged into master](https://github.com/alphagov/government-frontend/commit/77710d39df91273db9fd7e8202d9e9e5739c13ba), I think GitHub got confused.

PR #556, opened 2017-11-29
[WIP] Enhance nav sidebar with 'Other' section

vanitabarrett:
The navigation sidebar needs to be usable on contact pages, which display different data to other content pages (quick links and other contacts). This PR will expand the 'external_links' section to be more generic so that other data can be passed to the component.

Review app component guide:

Visual regression results:


app/views/components/_navigation-sidebar.html.erb
```diff
@@ -4,52 +4,70 @@
...
+      <% end %>
+    <% end %>
 
+    <% if other.any? %>
+      <% other.each do |other_sections| %>
+        <% other_sections.each do |section| %>
+          <h3 class="app-c-navigation-sidebar__sub-heading app-c-navigation-sidebar__sub-heading--bold" id="nav-sidebar-<%= section[:title].parameterize.underscore %>">
+            <%= section[:title] %>
+          </h3>
+          <nav role="navigation" class="app-c-navigation-sidebar__nav-section" aria-labelledby="nav-sidebar-<%= section[:title].parameterize.underscore %>">
```
steventux: It seems like the same set of elements are repeated for all 3 types of sections and their links.
Would it be possible to reduce this into a partial or helper method to reduce the repetition? I can see there's a bit of variation in css class names so maybe this would be tricky or overly-complex.
Also `other` links aren't the same data structure as the others. Would it be easier to write common code for all 3 types if the data was coerced into the same format before rendering?

app/assets/stylesheets/components/_navigation-sidebar.scss
```diff
@@ -11,10 +11,10 @@
 .app-c-navigation-sidebar__sub-heading {
   border-top: 1px solid $border-colour;
   padding-top: $gutter-half;
-}
 
-.app-c-navigation-sidebar__external-heading {
-  @include bold-19;
+  &--bold {
```
nickcolley: (non blocking) nit: gov.uk frontend is writing selectors out in full so they're easier to discover in the source code, for example when searching in github.

app/views/components/_navigation-sidebar.html.erb
vanitabarrett: @steventux This was my first stab at getting something as common as possible - I think the different styling definitely complicates things a bit but I'll have a look today and see if I can simplify it down further.

nickcolley:
Just a side note, does this navigation's name need to include 'sidebar' for example on small screens it wont be in a sidebar.

vanitabarrett:
@nickcolley Good point, no it doesn't. Maybe just `related-navigation` then?

app/views/components/_related-navigation-sidebar.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <h2 class="app-c-related-navigation-sidebar__main-heading" >
+      <%= t('components.navigation_sidebar.related_content') %>
+    </h2>
+
+    <% related_content.each do |section| %>
+      <% section.each do |section_title, links| %>
+        <% if links.any?%>
+          <% unless section_title === "related_items" %>
+            <h3 class="
+              app-c-related-navigation-sidebar__sub-heading
```
steventux: You could build this string of classes up in the erb block above and call a variable here.

app/views/components/_related-navigation-sidebar.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  related_content = [
+    {"related_items" => related_items ||= [] },
+    {"topics" => topics ||= [] },
+    {"publishers" => publishers ||= [] },
+    {"collections" => collections ||= [] },
+    {"policies" => policies ||= [] },
+  ]
+  other ||= []
+  format_data_for_related_navigation(related_content, other)
+%>
```
steventux: If you close ERB blocks with `-%>` they don't affect whitespace in the output.

nickcolley:
Nice work 👍

PR #557, opened 2017-11-29
Add logic to link up with govuk_navigation_helpers

vanitabarrett:
Wraps the govuk_navigation_helpers related_navigation_sidebar method in the content item presenter so that it is accessible within views. This will allow us to simply pass `@content_item.related_navigation_sidebar` to the sidebar component when we want to render it on content pages (full example below):

`<%= render 'components/related-navigation-sidebar', @content_item.related_navigation_sidebar %>`

**Note: the above code will only work once https://github.com/alphagov/government-frontend/pull/556 has been merged, as this renames the component and allows it to display external and contact links. Otherwise, use `render 'components/navigation-sidebar'`**

To-Do:

- [x]  Bump govuk_navigation_helpers gem once the logic has been merged and gem versioned.

Dependent on: https://github.com/alphagov/govuk_navigation_helpers/pull/85

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -58,6 +58,10 @@ class ContentItemPresenter
     @nav_helper.taxonomy_sidebar
   end
 
+  def related_navigation
+    @nav_helper.related_navigation_sidebar
+  end
+
   def related_items
     @nav_helper.related_items
   end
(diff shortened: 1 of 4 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -58,6 +58,10 @@ def taxonomy_sidebar
     @nav_helper.taxonomy_sidebar
   end
 
+  def related_navigation_sidebar
```
steventux: I think it's worth adding a test for this as it's a public method. It could be as simple as asserting that the relevant `@nav_helper` method is called as expected.

PR #559, opened 2017-11-30
Temporarily removes e2e tests from build as they are flakey

nickcolley:
These tests seem to frequently fail due to timeouts.

I don't think we should have flakey tests in master since people will start ignoring them entirely.

See https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/1519/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,5 +4,4 @@ REPOSITORY = 'government-frontend'
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject(publishingE2ETests: true)
 }


Jenkinsfile
```diff
@@ -4,5 +4,4 @@ REPOSITORY = 'government-frontend'
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject(publishingE2ETests: true)
```
kevindew: removing this will actually remove the whole build, what you'd want to do is: `govuk.buildProject()`

nickcolley:
Travis has a useful feature for this kind of thing

https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

nickcolley:
Sounds like this'll get resolved soon, will close.

PR #565, opened 2017-12-06
Add world locations to related nav component

vanitabarrett:
When testing the image formats universal layout, we noticed that speeches show related world locations  in the metadata [on the live page](https://www.gov.uk/government/speeches/2014-hillary-clinton-prize-for-women-peace-and-security), but [not on universal layout at all](https://government-frontend-pr-541.herokuapp.com/government/speeches/2014-hillary-clinton-prize-for-women-peace-and-security). We need to show these in the related navigation.

This PR uses the latest version of `govuk_navigation_helpers` which has been updated to return world locations https://github.com/alphagov/govuk_navigation_helpers/pull/88

screen shot 2017-12-06 at 13 52 30

Review app component guide:
https://government-frontend-pr-565.herokuapp.com/component-guide

Visual regression results:
https://government-frontend-pr-565.surge.sh/gallery.html

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,12 +28,13 @@ en:
     translations: "Translations"
   components:
     related_navigation:
-      collections: "Collections"
+      collections: "Collection"
       external_links: "Elsewhere on the web"
       policies: "Policy"
       publishers: "Detailed guidance from"
       related_content: "Related content"
       topics: "Explore the topic"
+      world_locations: 'Location'
     published_dates:
       full_page_history: "full page history"
       last_updated: "Last updated %{date}"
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -28,12 +28,13 @@ en:
     translations: "Translations"
   components:
     related_navigation:
-      collections: "Collections"
```
maxgds: Would it make sense to rename the key as well?

vanitabarrett: Hmm, I tend to have gone with the plural for the data within the component, to indicate that the variable can contain multiple things, and therefore also the translation so that it's clear how they're linked. I think this makes sense, but if people find it clearer the other way then I can change it. Don't think it should be a blocker for now though.

maxgds: Agree not a blocker. It just feels like the key should match the English translation as far as possible and the plural/non plural looks a bit confusing to me.

PR #579, opened 2017-12-08
Some tweaks to the type on the banners.

steventux:
Part of https://trello.com/c/q0PNOZkj/117-move-formats-with-images-to-universal-layout

Deleted the 1/3rd column for the aside. Made consultation date range
bold.

Review app component guide:

Visual regression results:

### DIFF
--- a/app/assets/stylesheets/components/_banner.scss
+++ b/app/assets/stylesheets/components/_banner.scss
@@ -20,26 +20,8 @@
   }
 }
 
-.app-c-banner__main {
-  @include grid-column( 2 / 3, $full-width: tablet, $float: right );
-
-  @include media(tablet) {
-    padding-right: 0;
-    border-left: 1px solid $white;
-  }
-}
-
-.app-c-banner__aside {
-  margin-top: $gutter-half;
-
-  @include media(tablet) {
-    margin-top: 0;
-    padding-right: $gutter / 2;
-  }
-}
-
 .app-c-banner__text {
-  @include bold-36;
+  @include core-24;
 }
 
 .app-c-banner__title {
@@ -48,7 +30,8 @@
 }
 
 .app-c-banner__desc {
-  @include core-24;
+  @include core-19;
   // Ensure the text has a line-length of around 60 characters
   max-width: 30em;
+  padding-top: $gutter-one-third;
 }
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -12,18 +12,8 @@
   </p>
 <% end %>
 <section class="app-c-banner<% if aside %> app-c-banner--grid<% end %>" aria-label="Notice">
+  <%= content_block %>
   <% if aside %>
-    <div class="grid-row">
-      <div class="app-c-banner__main">
-        <%= content_block %>
-      </div>
-      <div class="column-third">
-        <div class="app-c-banner__aside">
-          <p><%= aside %></p>
-        </div>
-      </div>
-    </div>
-  <% else %>
-    <%= content_block %>
+    <p class="app-c-banner__desc"><%= aside %></p>
   <% end %>
 </section>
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/components/_banner.scss
```diff
@@ -48,7 +39,11 @@
 }
 
 .app-c-banner__desc {
-  @include core-24;
+  @include core-19;
   // Ensure the text has a line-length of around 60 characters
   max-width: 30em;
 }
+
+.consultation-date {
```
vanitabarrett: When I created the banner component, it was suggested that the consultation date styling should still live with the consultation CSS scope as it's not technically a class within this component (it's not scoped to the banner CSS - app-c-banner etc). 
https://github.com/alphagov/government-frontend/pull/424#discussion_r130863515

app/assets/stylesheets/components/_banner.scss
```diff
@@ -20,17 +20,8 @@
...
-    border-left: 1px solid $white;
-  }
-}
-
 .app-c-banner__aside {
-  margin-top: $gutter-half;
+  padding-top: $gutter-one-third;
 
   @include media(tablet) {
     margin-top: 0;
```
vanitabarrett: I'm not sure these styles are needed now that the aside isn't located to the left of the other banner content. Probably worth removing and testing to make sure it still works on all screen sizes

app/views/components/_banner.html.erb
```diff
@@ -12,18 +12,10 @@
...
-    <div class="grid-row">
-      <div class="app-c-banner__main">
-        <%= content_block %>
-      </div>
-      <div class="column-third">
-        <div class="app-c-banner__aside">
-          <p><%= aside %></p>
-        </div>
-      </div>
+    <div class="app-c-banner__aside">
```
vanitabarrett: I think the `aside` and the `desc` classes can probably be consolidated into one class applied to both paragraphs as they have similar styling.

app/assets/stylesheets/components/_banner.scss
```diff
@@ -48,7 +39,11 @@
 }
 
 .app-c-banner__desc {
-  @include core-24;
+  @include core-19;
   // Ensure the text has a line-length of around 60 characters
   max-width: 30em;
```
vanitabarrett: Is this max-width figure still correct now there is no aside? It looks quite short considering the width of the component 
screen shot 2017-12-08 at 14 59 56

steventux:
Thanks @vanitabarrett - cc @alextea @miaallers

app/assets/stylesheets/components/_banner.scss
miaallers: I think let's leave it for now: max character limit works well for reading so I'm less concerned that it looks odd than that it is less readable. Also we may be changing the banners again soon and I don't want to lose this bit of code :)

app/assets/stylesheets/components/_banner.scss
miaallers: Sounds good!

app/assets/stylesheets/components/_banner.scss
miaallers: I'm not 100% sure what the technical impact of this change is, so we can drop it if its bad practice.

app/views/components/_banner.html.erb
miaallers: Sounds good to me

steventux: @miaallers / @alextea can you make this change and push back onto this branch please?

steventux:
@miaallers @alextea can you make the requested/agreed changes to this PR please?

app/assets/stylesheets/components/_banner.scss
alextea: @vanitabarrett should we move this code to consultations

vanitabarrett: @alextea Yeah, should be in `consultations.scss`. This PR doesn't actually delete it from that file, so it might just be a case of deleting in the component CSS and checking it still exists in `consultations.scss`

fofr:
I wonder if `aside` and `desc` can be combined in the banner component now.

PR #590, opened 2017-12-11
[Do not merge] Move formats with contents lists to universal layout

maxgds:
Review app component guide:
https://government-frontend-pr-590.herokuapp.com/component-guide

Example urls:
Specialist document:
https://government-frontend-pr-590.herokuapp.com/countryside-stewardship-grants/management-of-moorland-vegetation-supplement-up4

Detailed guide:
https://government-frontend-pr-590.herokuapp.com/guidance/hm-land-registry-registration-services-fees

Document collection:
https://government-frontend-pr-590.herokuapp.com/government/collections/hm-revenue-and-customs-service-availability-and-issues#contents-list

Detailed guide:
https://government-frontend-pr-590.herokuapp.com/guidance/rates-and-thresholds-for-employers-2017-to-2018

Working group:
https://government-frontend-pr-590.herokuapp.com/government/groups/nuclear-energy-skills-alliance

Statistical data set:
https://government-frontend-pr-590.herokuapp.com/government/statistical-data-sets/unclaimed-estates-list#history

Topical event about page:
https://government-frontend-pr-590.herokuapp.com/government/topical-events/eu-referendum/about


app/assets/stylesheets/helpers/_body-with-contents-list.scss
```diff
@@ -0,0 +1,41 @@
...
+  z-index: 0;
+}
+
+.contents-list-link-wrapper.govuk-sticky-element--stuck-to-window {
+  background-color: $panel-colour;
+  margin-left: 0;
+  padding: $gutter-half;
+}
+
+.contents-list-link-wrapper.govuk-sticky-element--stuck-to-window {
```
nickcolley: These percentages with negative margins look suspicious, what does this achieve?

app/assets/stylesheets/helpers/_body-with-contents-list.scss
```diff
@@ -0,0 +1,41 @@
+@mixin body-with-contents-list {
+  .body-with-contents-list {
+    position: relative;
+    margin-bottom: $gutter * 1.5;
+  }
+}
+
+.contents-list-link-wrapper .app-c-back-to-top {
```
nickcolley: Should we keep component selectors within the component itself?

app/assets/stylesheets/helpers/_body-with-contents-list.scss
```diff
@@ -0,0 +1,41 @@
...
+  margin: $gutter-half 0;
+}
+
+// sticky
+.body-with-contents-list .app-c-contents-list {
+  z-index: 0;
+}
+
+.contents-list-link-wrapper.govuk-sticky-element--stuck-to-window {
```
nickcolley: This feels specific to contents list but it's styling the back to top link, should this be a component?

nickcolley:
The application examples seem to be throwing errors, so I can't review this properly sorry

app/assets/stylesheets/helpers/_body-with-contents-list.scss
maxgds: It's hacky and there's probably a better way:

Content is centred. 
I want a band that is 100% viewport wide with a background colour.
Inside that, I want a link (back to contents link) that lines up exactly with the content. I don't know what that left margin is. If I over-size the band and pull it left by a specific amount, then I know that pushing the link inside back that far will line it up perfectly. 

But there is almost certainly a better way and I'd love to know what it is!

app/assets/stylesheets/helpers/_body-with-contents-list.scss
maxgds: The problem is that we are not moving over html documents, so it needs to preserve the old style of back to top link. I also didn't know if it was used elsewhere. Same cause for the specific-to-contents list comment below.

maxgds:
@nickcolley Sorry about the example urls, I mangled them when pasting. Fixed now.

nickcolley:
The important metadata component seems to output oddly.

screen shot 2017-12-11 at 16 32 19

You could add more examples that test the different variations, see the original metadata for a good example of this:

http://govuk-static.herokuapp.com/component-guide/metadata

nickcolley:
This variation does not stop before the contents list as expected.

screen shot 2017-12-11 at 16 33 58

nickcolley:
We're trying to avoid scoped styles that are based on individual views if possible.

We might be able to refactor the back to top link so it does not have left/right margins.

Ideally this would be a component and the only thing that is not in the component guide is the sticky behaviour.

If we want to modify a components output, we can consider passing an option as a modifier for example.

If that's not possible, the selectors should still be inside the component's source files, for example if we deleted `back-top.css` it should remove all traces, this makes sure people know the possible outputs for this.

nickcolley:
Caught up with Max in person.

The sticky behaviour needs some love, looks like there's CSS in this PR that overrides some behaviour.

We may need to change how the wrapping container elements works, so it stops in the right place.

maxgds:
The metadata has been fixed outside of this PR.

maxgds:
Experimenting with moving the sticky wrapper hasn't been successful so far.

maxgds:
This has been superceded by other work on these format.

PR #598, opened 2017-12-12
Route radio buttons on choose sign in page

leenagupte:
Trello card: https://trello.com/c/zsPE3IXV

Add routing to the radio buttons on the `choose sign in` page of the service_sign_in format.

The actual error handling, that displays an error message to the user, will be added in another PR. Reloading the `choose sign in` page if the user doesn't select an option just stops an exception from being raised.

### DIFF
--- a/app/presenters/service_sign_in/choose_sign_in_presenter.rb
+++ b/app/presenters/service_sign_in/choose_sign_in_presenter.rb
@@ -33,6 +33,12 @@ module ServiceSignIn
       content_item['links']['parent'].first['base_path']
     end
 
+    def selected_option(selected_value)
+      symbolized_options.each do |option|
+        return option if option[:text].parameterize == selected_value
+      end
+    end
+
   private
 
     def choose_sign_in
--- a/app/presenters/service_sign_in/choose_sign_in_presenter.rb
+++ b/app/presenters/service_sign_in/choose_sign_in_presenter.rb
@@ -13,19 +13,19 @@ module ServiceSignIn
     end
 
     def options
-      mapped_options = symbolized_options.map do |option|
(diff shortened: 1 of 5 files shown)


app/presenters/service_sign_in/choose_sign_in_presenter.rb
```diff
@@ -33,6 +33,12 @@ def back_link
       content_item['links']['parent'].first['base_path']
     end
 
+    def selected_option(selected_value)
+      symbolized_options.each do |option|
```
emmabeynon: Are we able to iterate through `options` and compare the `selected_value` against `value` as this already has the paramterized text value?

test/controllers/content_items_controller_test.rb
```diff
@@ -346,6 +346,22 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_equal response.headers["Access-Control-Allow-Origin"], "*"
   end
 
+  test "service_sign_in_options with option param set" do
+    content_item = content_store_has_schema_example("service_sign_in", "service_sign_in")
+    path = "#{path_for(content_item)}/#{content_item['details']['choose_sign_in']['slug']}"
+
+    option = content_item['details']['choose_sign_in']['options'][0]
```
emmabeynon: As we have to call `content_item['details']['choose_sign_in']` a couple of times in this test, could we extract it out and assign to a variable to make it a bit easier to read?

app/presenters/service_sign_in/choose_sign_in_presenter.rb
emmabeynon: Oh I've now seen the next commit where this is sorted out :)

PR #606, opened 2017-12-13
Add more tests for service_sign_in based on new examples, includes welsh and 'without or divider'

nickcolley:
We added more examples in https://github.com/alphagov/govuk-content-schemas/pull/689

This PR adds tests to match the new examples, so we're covering localisation and making sure that small amount of options do not have an 'or' between them.

- https://government-frontend-pr-606.herokuapp.com/examples/service_sign_in/service_sign_in/choose-sign-in
- https://government-frontend-pr-606.herokuapp.com/examples/service_sign_in/welsh/dewiswch-lofnodi
- http://government-frontend.dev.gov.uk/examples/service_sign_in/view_driving_licence/choose-sign-in

As part of https://trello.com/c/vsmhbjLt/245-build-the-choose-sign-in-options-template-2


test/integration/service_sign_in/choose_sign_in_test.rb
```diff
@@ -58,9 +58,25 @@ class ChooseSignInTest < ActionDispatch::IntegrationTest
...
+        assert page.has_css?(".app-c-radio__label__text", text: "Use GOV.UK Verify")
+        assert page.has_css?(".app-c-radio__label__hint", text: "You can use an existing identity account or create a new one. It usually takes about 5 minutes to create an account.")
+      end
+
+      refute page.has_css?(".app-c-radio__block-text", text: "or")
+    end
+
+    def setup_and_visit_choose_sign_in_page(example_name, path)
+      content_item = get_content_example(example_name)
+      path = content_item["base_path"] + path
```
leenagupte: I would rename one of the `path` variables.

nickcolley:
As far as I know Jenkins should be cloning master of the govuk-content-schemas repo, so I'm not sure why it can't find the example files.

leenagupte:
The create_new_account test fails locally too. I think this content needs to be updated in the tests as well.

nickcolley:
Found out that the examples need to be deployed before they can be used, re-building now.

PR #609, opened 2017-12-14
Add worldwide organisations to published by

vanitabarrett:
... in the related navigation component

The new related navigation is not displaying topical events or worldwide organisations. We want to display both of these.

**Worldwide organisations example (before and after):**
screen shot 2017-12-14 at 11 16 39
[Live example](https://government-frontend-pr-609.herokuapp.com/government/news/paralympic-legacy-and-disability-rights)

**Topical events example (before and after):**
screen shot 2017-12-14 at 11 29 32
[Live example](https://government-frontend-pr-609.herokuapp.com/government/news/uk-china-relations-flourish-at-the-2017-people-to-people-dialogue)

[Component Guide link](https://government-frontend-pr-609.herokuapp.com/component-guide/related-navigation)

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -59,7 +59,7 @@ class ContentItemPresenter
   end
 
   def related_navigation
-    @nav_helper.related_navigation_sidebar
+    merge_publishers_and_worldwide_organisations(@nav_helper.related_navigation_sidebar)
   end
 
   def related_items
@@ -68,10 +68,6 @@ class ContentItemPresenter
 
   def tagged_to_a_taxon?
     content_item.dig("links", "taxons").present?
-  end
-
-  def related_navigation_sidebar
-    @nav_helper.related_navigation_sidebar
   end
 
 private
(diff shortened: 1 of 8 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -103,4 +99,17 @@ def text_direction
   def native_language_name_for(locale)
     I18n.t("language_names.#{locale}", locale: locale)
   end
+
+  def merge_publishers_and_worldwide_organisations(related_navigation)
+    worldwide_orgs_and_publishers = related_navigation[:publishers]
```
steventux: Could this be simplified to
```ruby
related_navigation[:publishers] += related_navigation.delete(:worldwide_organisations)
```
I think this would be safe because `related_navigation[:worldwide_organisations]` returns an empty array if there are no worldwide organisations. It would also remove the `:worldwide_organisations` key.

vanitabarrett: @steventux Great, I knew there would be a simpler way of doing this 👍

PR #610, opened 2017-12-14
Feature: Service sign in

leenagupte:
Based off: https://github.com/alphagov/government-frontend/pull/591

Paired with: @emmabeynon & @nickcolley 

Trello cards:
https://trello.com/c/vsmhbjLt
https://trello.com/c/zsPE3IXV
https://trello.com/c/1i7oH6CO
https://trello.com/c/5rgT5gST
https://trello.com/c/ydt46bDm

## User need
As a user
I need to be able to clearly see the different options to sign-in to a service
So that I can make the right decision for me

## Detail
This PR renders the pages in the "service_sign_in" format.
The format allows for two different types of pages in the same content item:

* Choose sign in
* Create new account (optional)

[Example content item](https://github.com/alphagov/govuk-content-schemas/blob/master/examples/service_sign_in/frontend/service_sign_in.json)

The base path for the service_sign_in content item is assumed to be "{start page path}/sign-in", and the choose sign in and create new account pages follow this format:

* {start page path}/sign-in/{choose sign in slug}
* {start page path}/sign-in/{create new account slug}

### Heroku examples
- [Default: Choose sign in](https://goo.gl/Z1o8vX)

    - [Screenshot: Mobile](https://user-images.githubusercontent.com/2445413/33993421-5fb68f1c-e0ce-11e7-939e-c85590973722.png)

    - [Screenshot: Desktop](https://user-images.githubusercontent.com/2445413/33993427-67881b52-e0ce-11e7-872a-b6da35b78ce9.png)

- [Default: Create an account](https://goo.gl/jB3sbN)

    - [Screenshot: Mobile](https://user-images.githubusercontent.com/2445413/33993423-624899aa-e0ce-11e7-8efe-4be5f737ac8a.png)

    - [Screenshot: Desktop](https://user-images.githubusercontent.com/2445413/33993425-6651e93e-e0ce-11e7-85d7-d710c7a27090.png)

- [Welsh](https://goo.gl/ra6BLe)
- [VDL](https://goo.gl/bLnQrZ)

## Test plan
https://docs.google.com/a/digital.cabinet-office.gov.uk/spreadsheets/d/12jWWGW5xImDkVoFNr2ZNlIIKIxDyqVBHihPNaXtxXHY/edit?usp=sharing

PR #611, opened 2017-12-14
Add Welsh translations for service_sign_in pages

leenagupte:
Trello card: https://trello.com/c/T5zKxHRw
Depends on PR: https://github.com/alphagov/government-frontend/pull/610

Add the Welsh translation to the locales for:

* Back link
* Continue button
* Error messages

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -6,6 +6,8 @@ cy:
   components:
     radio:
       or: 'neu'
+    back_link:
+      back: "Yn ôl"
   content_item:
     schema_name:
       announcement:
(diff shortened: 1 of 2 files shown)


config/locales/cy.yml
```diff
@@ -260,6 +262,9 @@ cy:
         few:
         many:
         other: Nodiadau siarad
+      service_sign_in:
+        one: Gwasanaeth Mewngofnodi
```
emmabeynon: I had a conversation with Tom about this translation when he asked me to sense check this as a slug.  I think it's incorrect as it translates as "Sign in service".  I think it should be "Mewngofnodi i wasanaeth" which is literally "Sign in to service" and I think more correct.

leenagupte: Thanks! I'll update and fix up. There's also a failing test that needs to fixed.

leenagupte: Speaking of the slug. We haven't considered translating "sign-in" in the base_path!

I've agreed with Tom, that's it's not as important as the rest of the translated content, and that we can add it as a stretch task for next week.

PR #613, opened 2017-12-15
Turn off CSRF protection on service_sign_in POST route

leenagupte:
Related to: PR #610 

When we tried submitting a radio button on the choose-sign-in page the
following error is thrown:

`Can't verify CSRF token authenticity.`

As we are not submitting and sensitive data, and are just redirecting users,
we don't need this protection on the form.

We don't currently have a session store due to the layers of caching, so we
did not try to set the authenticity token, and have instead disabled it.

We did not reproduce this issue on Heroku and will investigate why later.


tijmenb:
See https://github.com/alphagov/email-alert-frontend/blob/master/docs/arch/adr-001-sessions.md for the same in email-alert-frontend.

PR #616, opened 2017-12-18
Make publisher metadata styling consistent

steventux:
https://trello.com/c/Q9io9cQU/132-move-publications-format-to-universal-design

When moving publications to the new layout we wanted
to combine the publisher metadata with an optional
logo and make the combination span all 3 columns.
This change needs to be reflected in formats which
won't have logos but will have a full width publisher
metadata border.
I've made a partial rendering the metadata component
in a 2/3 width column and the logo in the 1/3 column
because it seemed the most reasonable way to combine
both without compromising the component design.

screenshot from 2017-12-18 11-51-16

### Example case study

https://government-frontend-pr-616.herokuapp.com/government/case-studies/who-will-and-won-t-be-enrolled-into-a-workplace-pension-fictional-case-studies

### Example news article

https://government-frontend-pr-616.herokuapp.com/government/news/apprenticeships-will-contribute-34-billion-to-uk-economy-this-year

Review app component guide:

Visual regression results:

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -17,6 +17,7 @@
 @import 'helpers/add-title-margin';
 @import "helpers/task-list-title";
 @import "helpers/content-bottom-margin";
+@import "helpers/publisher-metadata-with-logo";
 
 // components
 @import 'components/*';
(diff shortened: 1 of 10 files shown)


fofr:
Should the logo be part of the important metadata component?

I was looking through the guide trying to find an example with a logo next to it. A logo could be passed in as an argument.

PR #622, opened 2017-12-19
Increase test coverage of tasklists

sihugh:
We weren't testing publications or answers here which missed a breaking change.

We also had a require for the govuk_publishing_components gem which we no longer needed as we always load it in the Gemfile now (since https://github.com/alphagov/government-frontend/pull/578)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -11,10 +11,6 @@ require "rails/test_unit/railtie"
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-if !Rails.env.production? || ENV['HEROKU_APP_NAME'].present?
-  require 'govuk_publishing_components'
-end
-
 module GovernmentFrontend
   class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -11,10 +11,6 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-if !Rails.env.production? || ENV['HEROKU_APP_NAME'].present?
```
nickcolley: This feels really dangerous

sihugh: Since we removed the `require: false` in https://github.com/alphagov/government-frontend/commit/4b0e39570e6401eefcc76c6b815017a396118256#diff-8b7db4d5cc4b8f6dc8feb7030baa2478L10 it's had this behaviour anyway.  I'll look into whether we can tune what's loaded in govuk_publishing_components.

sihugh: Since we removed the require: false in 4b0e395#diff-8b7db4d5cc4b8f6dc8feb7030baa2478L10 it's had this behaviour anyway. I'll look into whether we can tune what's loaded in govuk_publishing_components.

PR #623, opened 2017-12-19
Reduce border span on publisher metadata for fatality notices

steventux:
This format doesn't currently have a sidebar navigation
which means that the full-width border on the publisher
metadata + logo grid row extends out into an empty column.
This doesn't look very balanced, so constrain it to 2/3rds
width. We expect to iterate this format with better sidebar
navigation soon so this is a temporary design tweak.

screenshot from 2017-12-19 14-38-03

Review app component guide:
https://government-frontend-pr-623.herokuapp.com/component-guide

Example:
https://government-frontend-pr-623.herokuapp.com/government/fatalities/fusilier-gordon-gentle

### DIFF
--- a/app/assets/stylesheets/views/_fatality-notice.scss
+++ b/app/assets/stylesheets/views/_fatality-notice.scss
@@ -1,3 +1,0 @@
-.fatality-notice {
-  @include sidebar-with-body;
-}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -18,6 +18,7 @@
 @import "helpers/task-list-title";
 @import "helpers/content-bottom-margin";
 @import "helpers/publisher-metadata-with-logo";
+@import "helpers/horizontal-rule";
 
 // Components from this application
 @import 'components/*';
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -45,7 +45,6 @@
 @import 'views/detailed-guide';
 @import 'views/publication';
 @import 'views/document-collection';
-@import 'views/fatality-notice';
 @import 'views/statistical-data-set';
 @import 'views/consultation';
 @import 'views/speech';
--- a/app/assets/stylesheets/helpers/_horizontal-rule.scss
+++ b/app/assets/stylesheets/helpers/_horizontal-rule.scss
@@ -1,0 +1,6 @@
+.app-c-horizontal-rule {
+  background-color: $border-colour;
+  display: inline-block;
+  height: 1px;
+  width: 100%;
+}
--- a/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
+++ b/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
@@ -3,7 +3,6 @@
 // formats so they may ultimately become component styles.
 .metadata-logo-wrapper {
   @extend %contain-floats;
-  border-top: 1px solid $border-colour;
   margin-left: $gutter-half;
   margin-right: $gutter-half;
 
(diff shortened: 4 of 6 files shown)


app/assets/stylesheets/views/_fatality-notice.scss
```diff
@@ -1,3 +1,5 @@
 .fatality-notice {
-  @include sidebar-with-body;
+  .app-c-publisher-metadata {
```
vanitabarrett: Overriding component CSS isn't ideal, but not sure at the moment how we'd do it another way.

steventux: I'd expect this to be updated/removed once we've had a conversation with MoD to understand the needs for this format, we intend to provide sidebar links, but we need to build some additional functionality before adding them to this format, so this should be a temporary update.

nickcolley:
Can you link to an example of this rendered?

Agree with @vanitabarrett we should not be writing page specific CSS if it can be avoided.

It's not clear from this PR what problem is being solve so hard to give advice on an alternative.

nickcolley:
Had a look at this.

If you ensured unidirectional spacing and added a helper you could make this group of things work better together so it'd be easier to pull off what you're trying to do.

If we make sure the spacing between components is only in one direction then components don't need specific styling to adding spacing above

(Example uses a horizontal rule component would need to decide if this should be presentation (span) or not (hr) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr)



```html
<style>
  .app-c-horizontal-rule {
    display: block;
    border: 0;
    border-bottom: 1px solid #bfc1c3;
    margin: 0;
    margin-bottom: 15px;
  }

  .app-u-float-right {
    float: right;
  }

  .app-u-responsive-bottom-margin {
    margin-bottom: 15px;
  }
  @media (min-width: 641px) {
    .app-u-responsive-bottom-margin {
      margin-bottom: 45px;
    }
  }
  /* Remove padding top from metadata, allow above component to define spacing */
  .app-c-publisher-metadata {
    padding-top: 0;
  }
</style>
<!-- Component could have an option to set as presentational and swap between span and hr -->
<span class="app-c-horizontal-rule"></span>
<div class="grid-row">
    <div class="app-c-publisher-metadata">
      <div class="app-c-publisher-metadata__other">
        <dl data-module="track-click">
          <dt class="app-c-publisher-metadata__term">From: </dt>
          <dd class="app-c-publisher-metadata_definition">
            <a href="/government/organisations/department-for-environment-food-rural-affairs">Department for Environment, Food &amp; Rural Affairs</a>
          </dd>
        </dl>
      </div>
    </div>
  </div>
  <div class="column-one-third">
    <img
      alt="National Statistics"
      class="app-u-float-right app-u-responsive-bottom-margin"
      src="https://government-frontend-pr-623.herokuapp.com/government-frontend/national-statistics-420d17fb4625fb50ba50f0f1d03127f79168bd665633847e479073677e295dea.png"
    >
  </div>
</div>
```

then to have it work in only one column it'd be a case of moving the horizontal rule

```html
<div class="grid-row">
    <span class="app-c-horizontal-rule"></span>
    <div class="app-c-publisher-metadata">
      <div class="app-c-publisher-metadata__other">
        <dl data-module="track-click">
          <dt class="app-c-publisher-metadata__term">From: </dt>
          <dd class="app-c-publisher-metadata_definition">
            <a href="/government/organisations/department-for-environment-food-rural-affairs">Department for Environment, Food &amp; Rural Affairs</a>
          </dd>
        </dl>
      </div>
    </div>
  </div>
  <div class="column-one-third"></div>
</div>
```

fofr:
What's the status of this PR?

steventux:
@fofr this needs work, it's been low priority but I'll try and get this PR up to date (or closed) today.

steventux:
We're iterating the navigation for announcements, this PR won't be necessary.

PR #630, opened 2017-12-19
Update test to ensure random example always runs

nickcolley:
This could be done with seeding but then we wouldnt get a random result each time.

As part of https://trello.com/c/EzyEJjTk/296-update-create-new-account-test-so-it-does-not-randomly-skip


nickcolley:
@tijmenb this test is only really useful when it does exist

fofr:
@nickcolley Why is the test only useful when `create_new_account` exists?

nickcolley:
@fofr since for this new feature `create_new_account` is in the same content item as `choose_sign_in`, `create_new_account` is optional as for some services we do not need (or cannot) help users create an account.

PR #633, opened 2017-12-20
Reduced the spacing between on related nav

miaallers:
Done in order to visually tighten up the grouping.

Review app component guide:
https://government-frontend-pr-633.herokuapp.com/component-guide

Before:
screen shot 2017-12-20 at 11 31 07

After:
screen shot 2017-12-20 at 11 30 51

Link to check:
https://www.gov.uk/government/publications/russia-tax-treaties

### DIFF
--- a/app/assets/stylesheets/components/_related-navigation.scss
+++ b/app/assets/stylesheets/components/_related-navigation.scss
@@ -23,7 +23,6 @@
 }
 
 .app-c-related-navigation__nav-section {
-  margin-top: $gutter-one-third / 2;
   margin-bottom: $gutter;
 }
 


nickcolley:
Links from the review app:

- https://government-frontend-pr-633.herokuapp.com/component-guide/related-navigation
- https://government-frontend-pr-633.herokuapp.com/government/publications/russia-tax-treaties

fofr:
The spacing between "International tax" and "Living or working…" looks too big. Like the second link in the list doesn't belong to the "Explore the topic" group

screen shot 2018-01-02 at 12 04 04

miaallers:
I don't agree, I think the fact that it's a list will allow users to infer the relationship.

fofr:
I interpreted it with first impressions that the second part "Living or" wasn't a topic. I had to check the code to make sure it was a topic and that they were meant to be connected with each other. I believe others will make this interpretation too.

miaallers:
I wonder then if the space between the contextual title and the two links should be *increased* instead. I'll take a look just now

miaallers:
Here are examples with equidistant spacing. I think it begins to degrade the grouping relationship. Also, in cases where we have light and bold links, we are now dealing with a situation where the contextual headings look like broken links. 

It also just looks inelegant.

screen shot 2018-01-03 at 11 08 41
screen shot 2018-01-03 at 11 09 56

PR #636, opened 2017-12-20
Add tracking to related navigation component

vanitabarrett:
Tracking added to related navigation component:
https://government-frontend-pr-636.herokuapp.com/component-guide/related-navigation

### DIFF
--- a/app/views/components/_related-navigation.html.erb
+++ b/app/views/components/_related-navigation.html.erb
@@ -16,14 +16,16 @@
 %>
 <% if related_content.map(&:values).flatten.any? || other.flatten.any? %>
   <aside class="app-c-related-navigation" role="complementary">
-    <h2 id="related-nav-related_items" class="app-c-related-navigation__main-heading" >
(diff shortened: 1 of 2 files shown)


app/views/components/_related-navigation.html.erb
```diff
@@ -19,7 +19,7 @@
     <h2 id="related-nav-related_items" class="app-c-related-navigation__main-heading" >
       <%= t('components.related_navigation.related_content') %>
     </h2>
-    <% related_content.each do |section| %>
+    <% related_content.each_with_index do |section, section_index| %>
```
steventux: This could be `related_content.each.with_index(1) do |section, section_index|` which would start the index at 1 and simplify line 47. It's also consistent with how the links are iterated below.

vanitabarrett: @steventux Had to change the logic slightly anyway as the section_index was being incremented even for sections without any links to display.

PR #637, opened 2017-12-20
Move working group pages to universal layout

steventux:
https://trello.com/c/oLC1enhS/133-move-formats-that-have-a-content-list-to-universal-design

### Example working group page

https://government-frontend-pr-637.herokuapp.com/government/groups/joint-vat-consultative-committee-jvcc

screenshot from 2017-12-20 17-17-27

### Sticky link:

screenshot from 2017-12-20 17-17-52

Review app component guide:

Visual regression results:

### DIFF
--- a/app/views/components/docs/contents-list-with-body.yml
+++ b/app/views/components/docs/contents-list-with-body.yml
@@ -78,7 +78,6 @@ examples:
             <h3 id="contact">Contact</h3>
             <p>Please send written representations about any competition issues to:</p>
             <div class="address"><div class="adr org fn"><p>
-
             <br>Competition and Markets Authority 
             <br>Victoria House 
             <br>Southampton Row 
--- a/app/views/content_items/working_group.html.erb
+++ b/app/views/content_items/working_group.html.erb
@@ -1,6 +1,3 @@
-<%= render 'shared/title_and_translations', content_item: @content_item %>
-<%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
-
 <% @additional_body = capture do %>
   <% if @content_item.policies.any? %>
     <h2 id="policies">Policies</h2>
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/helpers/_contents-with-body.scss
```diff
@@ -0,0 +1,28 @@
...
+  @include media(tablet) {
+    max-width: 990px;
+  }
+
+  .app-c-back-to-top {
+    margin-left: 0;
+  }
+}
+
+.contents-list-link-wrapper.govuk-sticky-element--stuck-to-window {
```
fofr: This ought to be a component.

app/assets/stylesheets/helpers/_contents-with-body.scss
```diff
@@ -0,0 +1,28 @@
...
+  background-color: $panel-colour;
+  display: inline;
```
vanitabarrett: Are `display: inline` and `float: left` needed? I tried toggling them off and it didn't seem to make a difference.

maxgds:
@fofr re: The ios grey bar still showing when contents is on screen on mobile

- I can't make this happen at all when scrolling to the top on browserstack or on real devices
- I can reliably reproduce it on ios only when clicking the back to top link

However, it seems to be down to the way that ios renders the div that it jumps back to, it positions it VERY slightly off screen, the skip link is designed to show when the contents are moving off screen, so it's more of an ios positioning bug than anything else. Should we try to fix it considering that? The grey bar won't obscure anything the user wants to interact with so it feels to me like it would be ok to leave it as is?

app/assets/stylesheets/helpers/_contents-with-body.scss
steventux: @vanitabarrett you're right - they aren't necessary - will remove.

vanitabarrett:
It looks as though the page content is full width rather than 2/3 width? 
https://government-frontend-pr-637.herokuapp.com/government/groups/joint-vat-consultative-committee-jvcc

PR #638, opened 2017-12-20
Only run E2E tests which involve government-frontend

RyanMacG:
Hey :wave:,

We've tagged all the tests in the E2E repo which interact and depend on the government-frontend application in some way.

By only running the tests which interact with the government-frontend app we can reduce the running time of the E2E tests for commits against this app without reducing the confidence that the E2E tests bring. 🎉


RyanMacG:
New tag needs to be on test-against first

PR #643, opened 2017-12-21
[Don't merge] Prototype of contextual navigation

tijmenb:
This is a prototype where we generate lists of "related" content for pages.

There are a bunch of examples here:

- Popular/Recent/Similar pages published by the publishing org 
- Popular/Recent/Similar pages with the same document type 
- Popular/Recent/Similar pages with the same policy 
- Popular/Recent/Similar pages with the same person 
- Popular/Recent/Similar pages with the same taxon

"Popular" uses the search-api's popularity (number of pageviews in the last 45 days). "Similar" uses the search-api's "similar to" feature that looks in the text to find related pages.

The idea here is that we can try out different scopes (org, document type) with the different sorting mechanisms (popular/similar/recent). Click data can show us which perform better.

Other scopes that we could try:

- [x] Organisation
- [x] Document type
- [x] Policy
- [x] Person
- [x] Taxon
- [ ] Content purpose supertype
- [ ] Group
- [ ] Taxon and children
- [ ] Taxon and parents
- [ ] Legacy topic
- [ ] Mainstream browse
- [ ] Topical event
- [ ] Policy area
- [ ] Statistical dataset

### Examples

- https://government-frontend-pr-643.herokuapp.com/government/speeches/pm-statement-on-eu-negotiations-11-december-2017 (with a person)
- https://government-frontend-pr-643.herokuapp.com/1619-bursary-fund (taxon)

### DIFF
--- a/app/lib/services.rb
+++ b/app/lib/services.rb
@@ -1,7 +1,12 @@
 require 'gds_api/content_store'
+require 'gds_api/rummager'
 
 module Services
   def self.content_store
     @content_store ||= GdsApi::ContentStore.new(Plek.current.find("content-store"))
+  end
+
+  def self.rummager
+    @rummager ||= GdsApi::Rummager.new(Plek.current.find("rummager"))
   end
 end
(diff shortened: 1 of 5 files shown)


fofr:
@tijmenb Are there some screens of this prototype anywhere? The review app isn't available and won't always be available.

PR #649, opened 2018-01-02
Move answers and help to universal layout

vanitabarrett:
https://government-frontend-pr-649.herokuapp.com/software-tax-returns

**Answer Before:**
screen shot 2018-01-02 at 11 09 20

**Answer After:**
screen shot 2018-01-03 at 16 11 41

**Help Before:**
screen shot 2018-01-02 at 11 10 10

**Help After:**
screen shot 2018-01-03 at 16 11 31

**Links to Test:**

- [Answer Example](https://government-frontend-pr-649.herokuapp.com/national-minimum-wage-rates)
- [Help Example](https://government-frontend-pr-649.herokuapp.com/help/cookies)

Review app component guide:
https://government-frontend-pr-649.herokuapp.com/component-guide


fofr:
Looks good to me :+1:

Whats the state of the universal navigation variant? Have we removed a lot of it? Is it used anywhere?

vanitabarrett:
@fofr I think the AB testing side of it got removed, as we decided to ship these designs without AB testing, and the rest used as a basis for the universal layout designs

fofr:
When there's only one more related link it'd be better to show only the one link. ie The more link takes up the same about of space as that extra link would. So the user gets the same content with fewer clicks.

screen shot 2018-01-02 at 11 15 39

fofr:
Is the "published by" bit necessary on mainstream pages. Traditionally they haven't been associated with any organisation. This seems odd:

screen shot 2018-01-02 at 11 17 50

fofr:
For clarity on adding organisation links to mainstream pages:

We are adding them for:
* consistency
* to increase cross linking between content, especially to government content

They might be good for users because:
* they indicate who published the content

They may not be good for users because:
* they link to a generic top level page that needs improvement

They may be political because:
* GDS actually publishes the content
* Departments aren't responsible for the words, only the facts

alextea:
The sidebar is missing the mainstream browse topics and/or whitehall topics

app/views/components/_related-navigation.html.erb
```diff
@@ -33,8 +33,9 @@
           <nav role="navigation" class="app-c-related-navigation__nav-section" aria-labelledby="related-nav-<%= section_title %>" data-module="toggle">
             <ul class="app-c-related-navigation_link-list" data-module="track-click">
               <% constructed_link_array = [] %>
+              <% section_link_limit = calculate_section_link_limit(links, max_section_length) %>
               <% links.each.with_index(1) do |link, index| %>
```
fofr: Why does the index start with 1 rather than 0?

vanitabarrett: @fofr So that the Google Analytics tracking we're adding in https://github.com/alphagov/government-frontend/pull/636#discussion_r158231629 starts at 1 for the first section rather than 0 (as per the [Trello story](https://trello.com/c/5KR4jsTw/161-event-tracking-on-new-related-navigation-links#comment-5a3a683195c9804e58a551f0))

vanitabarrett: I think it was originally added for something else which has since been removed, but it will now be used in the Google Analytics tracking so that numbers start from x.1 rather than x.0 https://github.com/alphagov/government-frontend/pull/652

app/views/components/_related-navigation.html.erb
```diff
@@ -58,18 +59,18 @@
...
-                  data-controls="toggle_<%= section_title %>"
-                  data-expanded="false"
-                  data-toggled-text="<%= t("govuk_component.metadata.toggle_less", default: "Show fewer") %>">
-                    <%= t("govuk_component.metadata.toggle_more", number: remaining_links(links, max_section_length), default: "+ #{remaining_links(links, max_section_length)} more") %>
-                </a>
-              </li>
-              <li class="app-c-related-navigation__link">
-                <span id="toggle_<%= section_title %>" class="js-hidden"><%= constructed_link_array.to_sentence.html_safe %></span>
-              </li>
+              <% if links.length > section_link_limit %>
```
maxgds: Personally I would have just gone with 
<% if links.length > section_link_limit + 1 %>
possibly with a comment to explain why, rather than adding quite verbose extra logic, would be good to hear other opinions though.

test/integration/answer_test.rb
```diff
@@ -14,11 +14,15 @@ class AnswerTest < ActionDispatch::IntegrationTest
 
   test "related links are rendered" do
     setup_and_visit_content_item('answer')
-    within shared_component_selector("related_items") do
-      elsewhere = @content_item["details"]["external_related_links"].first
-      elsewhere_page = JSON.parse(page.text).fetch("sections").last["items"].first
-      assert_equal elsewhere["title"], elsewhere_page["title"]
-      assert_equal elsewhere["url"], elsewhere_page["url"]
+
+    related_links = @content_item["details"]["external_related_links"].first
```
maxgds: I assume you don't want to loop through all links because collapsed items complicate it, but you're just taking the first link, here, I'd say the var name should match, eg related_link or  first_related_link

app/views/components/_related-navigation.html.erb
vanitabarrett: Do you mean instead of having the `calculate_section_link_limit` method?

maxgds: Yes, it just felt like a lot of overhead for effectively a + 1 in the loop initialisation.

vanitabarrett: @maxgds I think this is what I initially tried, but it didn't work and the logic was actually more complicated than I thought. Totally happy to change if there is a simpler way though, perhaps I'm missing something.

maxgds: Oh, don't worry then if it doesn't work, it was just an assumption on my part.

fofr:
Is this blocked on the issue @alextea is looking at regarding topics in Mainstream and Whitehall having the same name?

vanitabarrett:
@fofr Yes, I think they had a discussion with Lucy from content support today - need to catch up with @alextea  on the outcome of that

steventux:
@maxgds @emmabeynon this could do with another review as I've amended the navigation.

PR #652, opened 2018-01-02
Cherrypick add analytics tracking

vanitabarrett:
Tracking added to related navigation component:
https://government-frontend-pr-636.herokuapp.com/component-guide/related-navigation

### DIFF
--- a/app/views/components/_related-navigation.html.erb
+++ b/app/views/components/_related-navigation.html.erb
@@ -16,14 +16,16 @@
 %>
 <% if related_content.map(&:values).flatten.any? || other.flatten.any? %>
   <aside class="app-c-related-navigation" role="complementary">
-    <h2 id="related-nav-related_items" class="app-c-related-navigation__main-heading" >
+    <h2 id="related-nav-related_items" class="app-c-related-navigation__main-heading" data-track-count="sidebarRelatedItemSection" >
       <%= t('components.related_navigation.related_content') %>
     </h2>
-    <% related_content.each_with_index do |section, section_index| %>
+    <% section_index = 0 %>
+    <% related_content.each do |section| %>
       <% section.each do |section_title, links| %>
         <% if links.any? %>
+          <% section_index += 1 %>
           <% unless section_title === "related_items" %>
-            <h3 id="related-nav-<%= section_title %>" class="
+            <h3 id="related-nav-<%= section_title %>" data-track-count="sidebarRelatedItemSection" class="
               app-c-related-navigation__sub-heading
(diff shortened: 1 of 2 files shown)


app/views/components/_related-navigation.html.erb
```diff
@@ -16,14 +16,16 @@
...
       <%= t('components.related_navigation.related_content') %>
     </h2>
+    <% section_index = 0 %>
     <% related_content.each do |section| %>
       <% section.each do |section_title, links| %>
         <% if links.any? %>
+          <% section_index += 1 %>
           <% unless section_title === "related_items" %>
-            <h3 id="related-nav-<%= section_title %>" class="
+            <h3 id="related-nav-<%= section_title %>" data-track-count="sidebarRelatedItemSection" class="
```
fofr: What uses the `data-track-count` property?

fofr:
(PR needs a more accurate title)

app/views/components/_related-navigation.html.erb
vanitabarrett: @fofr I'm not too sure about this, the current sidebar in static has this property which is why I thought it safer to keep here too. :woman_shrugging: 

When I dug into it, it seems to have been added to avoid a bug where things were being reported as 0: https://github.com/alphagov/static/blob/master/app/views/govuk_component/related_items.raw.html.erb#L19-L36

fofr: Can you confirm this is doing something and that it still works as intended?

vanitabarrett: @fofr I'm pretty sure this is still needed. `sidebarRelatedItemSection` is queried [in page-content.js](https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/page-content.js#L29) which is used [in custom-dimensions.js](https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/custom-dimensions.js#L92) to build custom dimensions (namely dimension26).

When I run the [Google Analytics Debugger tool](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en) against a page with the static sidebar and a page with the new related sidebar, we can see dimension26 is correctly set to the number of sections in the sidebar in both cases:

**For static page with sidebar with 2 sections:**
[Link to page](https://government-frontend-pr-652.herokuapp.com/national-minimum-wage-rates)

screen shot 2018-01-02 at 16 58 20

**For new sidebar with 3 sections:**
[Link to page](https://government-frontend-pr-652.herokuapp.com/government/case-studies/bis-criminal-prosecutions)
screen shot 2018-01-02 at 16 59 40

PR #667, opened 2018-01-04
Add an explicit dependency for phantomjs

tuzz:
poltergeist needs phantomjs but doesn't explicitly depend on it. The
reason it works now is because jasmine-rails is loading phantomjs. We
think it's clearer to explicitly call out this dependency as per the
readme of the poltergeist project.

Review app component guide:

Visual regression results:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ end
 group :test do
   gem 'capybara'
   gem 'mocha'
+  gem 'phantomjs', '~> 2.1'
   gem 'poltergeist', require: false
   gem 'webmock', '~> 2.3.0', require: false
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -334,6 +334,7 @@ DEPENDENCIES
   jasmine-rails!
   logstasher (= 0.6.1)
   mocha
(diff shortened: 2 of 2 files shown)


fofr:
Jasmine-rails shouldn't be:
https://github.com/alphagov/government-frontend/blob/master/spec/javascripts/support/jasmine.yml#L57

tuzz:
@fofr: I think that configuration line means jasmine-rails won't install the binary executable for phantomjs, but it will still depend on the phantomjs gem which provides bindings to this executable. I think we set that configuration line because we install the binary through puppet rather than on-demand when the phantomjs gem is installed- if that makes sense?

fofr:
It won't require the gem:
https://github.com/searls/jasmine-rails/blob/b087e3c145a9dedc4f7996beb8ad876ce0689894/lib/jasmine_rails/runner.rb#L10

tuzz:
Hmm ok, closing in that case.

fofr:
@tuzz Seems like it's required here though:
https://github.com/alphagov/government-frontend/blob/18cdd4dba4f28dab901b5dc805e413e01f407597/test/support/poltergeist.rb#L2

PR #670, opened 2018-01-05
Update components gem to 3.2.0

andysellick:
Mainly for the new print styles for task lists.

Trello card: https://trello.com/c/9CD3dY7r/409-add-task-list-print-styles-to-apps

Review app component guide:
https://government-frontend-pr-670.herokuapp.com/component-guide

Visual regression results:
https://government-frontend-pr-670.surge.sh/gallery.html


nickcolley:
Out of curiosity since the gem components are the same name as the app components, do they break?

I'm willing to do a PR once this is in to remove the old instances, but worth knowing.

nickcolley:
Looks like this is not broken, so should be fine: https://government-frontend-pr-670.herokuapp.com/examples/service_sign_in/service_sign_in/choose-sign-in

andysellick:
@nickcolley you mean if there's a component in the app called task-list and one in the gem with the same name? That's a good question. We were running with the task list component in the gem and still in static and it caused no problems that I was aware of, so I think it's okay.

nickcolley:
@andysellick we have duplicates now, for the form components.

fofr:
@nickcolley Can you confirm and document the behaviour? My guess is that one takes precedence over another, and that should be the app version.

nickcolley:
Will do.

## What happens if a component is moved upstream to the gem but still exists in the application?
When we want to share components we can move them into the gem, when we update the gem version in the application the application will continue to use the application's version of the components over the guide due to the precedence that the local files have.

In order to move them over to the new version, delete the application component versions and update any integration tests to update the namespace.

You can see an example of this in the following PRs: [insert original PR to add component to government-frontend] [PR to move components to gem] [PR to remove components from government-frontend]

PR #671, opened 2018-01-05
Use upstream form components

nickcolley:
We moved some [form components to the gem](https://github.com/alphagov/govuk_publishing_components/pull/116) this PR removes the application ones and use the gem versions instead.

See https://government-frontend-pr-671.herokuapp.com/examples/service_sign_in/service_sign_in/choose-sign-in for an example of a real page using the new components

See the component guide for moved components: https://government-frontend-pr-671.herokuapp.com/component-guide


config/locales/cy.yml
```diff
@@ -4,7 +4,7 @@ cy:
   common:
     last_updated: "Diweddarwyd diwethaf"
   components:
-    radio:
+    radio: # TODO: Push this upstream to the govuk_publishing_gem and remove from here
```
nickcolley: See this TODO, I made a mistake forgetting to move the Welsh translation of the Radio component when moving it into the gem.

I'll follow this up with another PR, but this is a good illustration of why co-locating file is so important, this would have been pretty hard to mess up if the files were right next to each other.

nickcolley:
I'm going to bump this with the fix and clean it up.

app/assets/stylesheets/application.scss
```diff
@@ -19,8 +19,13 @@
 @import "helpers/content-bottom-margin";
 @import "helpers/publisher-metadata-with-logo";
 
-// components
+// Components from this application
 @import 'components/*';
+
+// Components from govuk_publishing_components gem
+@import 'components/fieldset';
```
fofr: Using the same namespace is confusing isn't it.
I would have expected `govuk_publishing_components/fieldset`

nickcolley: Yeah very, we could consider moving the components, for now this comment should help somewhat.

cc @andysellick

andysellick: Yeah, govuk_publishing_components/component-name would be good. Firebreak? :)

fofr: Namespace problem being looked at in https://github.com/alphagov/govuk_publishing_components/pull/136

PR #672, opened 2018-01-05
Extract the back_link component to govuk_publishing_components

tuzz:
Review app component guide:

Visual regression results:


app/views/content_items/service_sign_in/_choose_sign_in.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render "components/back-link", href: @content_item.back_link %>
+<%= render "components/back_link", href: @content_item.back_link %>
```
fofr: Gems in the component _ought_ to be using hyphens too. It's on my todo list to update the existing ones. They incorrectly copied the older Static component conventions.

fofr:
Examples:
http://government-frontend-pr-672.herokuapp.com/component-guide/back_link
http://government-frontend-pr-672.herokuapp.com/log-in-file-self-assessment-tax-return/sign-in/prove-identity

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
fofr: Though `govuk_publishing_components/components/back-link` would be an awkward mix of hyphens and underscores. 🤔

tuzz:
@fofr: Good spot, I was missing this:
https://github.com/alphagov/government-frontend/pull/672/files#diff-f859e2848b07324f808ce1a5ccd9fcbaR24

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
tuzz: We're already using hyphens when importing the `scss` files, so I think you're right, we ought to use hyphens everywhere. (This has already tripped me up a couple of times)

I don't think we include the name of the gem in any `import` or `render` lines, so hopefully there won't be many places where there's a mish-mash of underscores and hyphens.

Do you want me to update the gem now to use hyphens for all components? I suppose that'd need to be a major version bump.

PR #659, opened 2018-01-03
Move specialist documents to universal layout

steventux:
https://trello.com/c/OvoXIWYp/184-move-specialist-documents-to-universal-design

Moves specialist documents to universal layout. This format contains extensive metadata specific to the various document types and often long content lists.

- Adds a `contents-list-with-body` component which decorates content with a `contents-list` and `back-to-top` link

screenshot from 2018-01-03 18-36-32

#### Sticky _back-to-top_ contents link:

screenshot from 2018-01-03 18-50-01

#### CMA Case

- https://government-frontend-pr-659.herokuapp.com/cma-cases/euro-car-parts-andrew-page-merger-inquiry

#### AAIB Report

- https://government-frontend-pr-659.herokuapp.com/aaib-reports/aaib-investigation-to-agusta-aw139-g-cipw

#### Business Finance Support

- https://government-frontend-pr-659.herokuapp.com/business-finance-support/business-development-grant-scheme-scarborough-borough

#### Drug Safety Update

- https://government-frontend-pr-659.herokuapp.com/drug-safety-update/paraffin-based-treatments-risk-of-fire-hazard

#### Countryside Stewardship Grants

- https://government-frontend-pr-659.herokuapp.com/countryside-stewardship-grants/creation-of-inter-tidal-and-saline-habitat-on-intensive-grassland-ct7

Review app component guide:
https://government-frontend-pr-659.herokuapp.com/component-guide

Visual regression results:
https://government-frontend-pr-XXX.surge.sh/gallery.html

  
 Component guide for the new `content-list-with-body` component needs this https://github.com/alphagov/govuk_publishing_components/pull/117 to correctly render the component added in this PR.

PR #678, opened 2018-01-10
Revert "Revert "Move specialist documents to universal layout""

steventux:
Reverts alphagov/government-frontend#677


steventux:
@fofr @vanitabarrett would you mind re-reviewing this as it's a different PR from the original. [The componentry has already been merged](https://github.com/alphagov/government-frontend/pull/685) and the template work here is unchanged from https://github.com/alphagov/government-frontend/pull/659

PR #690, opened 2018-01-11
Add Paas manifest file.

stephenharker:
A first pass at a PaaS CF manifest file so that we can "cf push" this app to the PaaS.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,0 +1,19 @@
+---
+applications:
+- name: govuk-dev-government-frontend
+  memory: 256M
+  buildpack: ruby_buildpack
+  instances: 1
+  env:  
+    GOVUK_APP_DOMAIN: www.gov.uk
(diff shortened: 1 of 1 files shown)


manifest.yml
```diff
@@ -0,0 +1,19 @@
...
+    GOVUK_APP_DOMAIN: www.gov.uk
+    GOVUK_ASSET_ROOT: https://assets.digital.cabinet-office.gov.uk
+    GOVUK_WEBSITE_ROOT: https://www.gov.uk
+    # Public Production Content Store
+    PLEK_SERVICE_CONTENT_STORE_URI: https://www.gov.uk/api
+    
+    # Content Store running in PaaS
+    #PLEK_SERVICE_CONTENT_STORE_URI: https://srh-content-store.cloudapps.digital
+
+    PLEK_SERVICE_STATIC_URI: assets.digital.cabinet-office.gov.uk
```
issyl0: Isn't this `assets.publishing.service.gov.uk` now? (It redirects, but still.)

PR #695, opened 2018-01-12
Move statistical data sets to univeral layout

fmlharrison:
Trello https://trello.com/c/yreeM1zm/199-move-statistical-data-sets-to-universal-layout

---
Example of new layout:
screen shot 2018-01-12 at 16 44 32

Component guide for this PR:
https://government-frontend-pr-695.herokuapp.com/component-guide

Example on heroku:
https://government-frontend-pr-695.herokuapp.com/government/statistical-data-sets/tra04-pedal-cycle-traffic

### DIFF
--- a/app/views/content_items/statistical_data_set.html.erb
+++ b/app/views/content_items/statistical_data_set.html.erb
@@ -1,7 +1,38 @@
-<%= render 'shared/title_and_translations', content_item: @content_item %>
-<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
-<%= render 'shared/metadata', content_item: @content_item %>
-<%= render 'shared/history_notice', content_item: @content_item %>
-<%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
-<%= render 'shared/sidebar_contents_with_body', content_item: @content_item %>
-<%= render 'shared/footer', @content_item.document_footer %>
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/title',
+               context: t("content_item.schema_name.#{@content_item.document_type}", count: 1),
+               title: @content_item.title,
+               average_title_length: "long" %>
+  </div>
+  <%= render 'shared/translations', content_item: @content_item %>
+  <div class="column-two-thirds">
+    <%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
+    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
+    <%= render 'shared/history_notice', content_item: @content_item %>
+  </div>
+</div>
+
+<%= render 'shared/publisher_metadata_with_logo' %>
+ 
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= render 'components/important-metadata', items: @content_item.metadata[:other] %>
+
+    <%= render "components/contents-list-with-body", contents: @content_item.contents do %>
+    <div class="responsive-bottom-margin">
+      <%= render 'govuk_component/govspeak',
+                 content: @content_item.body,
+                 direction: page_text_direction %>
+    </div>
+    <div class="responsive-bottom-margin">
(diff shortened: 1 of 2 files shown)


app/views/content_items/statistical_data_set.html.erb
```diff
@@ -1,7 +1,41 @@
-<%= render 'shared/title_and_translations', content_item: @content_item %>
-<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
-<%= render 'shared/metadata', content_item: @content_item %>
-<%= render 'shared/history_notice', content_item: @content_item %>
-<%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
-<%= render 'shared/sidebar_contents_with_body', content_item: @content_item %>
-<%= render 'shared/footer', @content_item.document_footer %>
+<%= content_for :simple_header, @navigation.should_present_taxonomy_navigation? %>
```
steventux: I'm not sure we should be applying taxonomy navigation conditions to this format. It doesn't look like there was a previous taxonomy nav variant for this format so we should either add this to all formats or maintain the previous behaviour i.e. don't set `:simple_header` based on taxonomy navigation applicability.

app/views/content_items/statistical_data_set.html.erb
```diff
@@ -1,7 +1,41 @@
...
+
+<div class="grid-row">
+    <div class="column-two-thirds">
```
steventux: I think indentation should be set to 2 spaces for templates, looks to be the case in others.

app/views/content_items/statistical_data_set.html.erb
```diff
@@ -1,7 +1,41 @@
...
+        <%= render 'govuk_component/title',
+            context: t("content_item.schema_name.#{@content_item.document_type}", count: 1),
```
steventux: [If there are no different `document_type`s for statistical data sets](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/statistical_data_set/frontend/schema.json#L41) then we don't need to provide this `context` arg.

steventux:
@fmlharrison Can you add a heroku app example URL to the PR description?

app/views/content_items/statistical_data_set.html.erb
fofr: Another case for ERB linting.

app/views/content_items/statistical_data_set.html.erb
fofr: @steventux Not sure I agree. Doesn't it need to still say "Statistical data set" above the title?

app/views/content_items/statistical_data_set.html.erb
fofr: As @steventux says, this will always be false for statistical data sets:
https://github.com/alphagov/government-frontend/blob/d42d4056c8325c90b06ea89e56557addbb457730/app/models/navigation_type.rb#L9-L13

app/views/content_items/statistical_data_set.html.erb
steventux: ignore me, didn't understand the purpose of this correctly.

fofr:
Taking this opportunity to share a statistical data set with the longest change history:
https://government-frontend-pr-695.herokuapp.com/government/statistical-data-sets/unclaimed-estates-list

(click full page history 🙀 )

app/views/content_items/statistical_data_set.html.erb
fmlharrison: Shall I remove this completely then? 
Does it need replacing with anything?

steventux: Yes remove completely - the navigation is always present in the current live version, it should behave the same way.

steventux:
https://www.gov.uk/government/statistical-data-sets/unclaimed-estates-list the current live implementation does much the same. ie the user is left with no _contents_ link.

PR #697, opened 2018-01-15
Move corporate information page to universal layout

steventux:
Trello: https://trello.com/c/ZWoMdsNS/201-move-corporate-information-page-to-universal-layout

#### Example
https://government-frontend-pr-697.herokuapp.com/government/organisations/public-health-england/about/equality-and-diversity

#### Before
screenshot from 2018-01-15 16-32-23

#### After
screenshot from 2018-01-15 16-30-11

---

Component guide for this PR:
https://government-frontend-pr-697.herokuapp.com/component-guide

### DIFF
--- a/app/views/content_items/corporate_information_page.html.erb
+++ b/app/views/content_items/corporate_information_page.html.erb
@@ -1,36 +1,51 @@
+<% @additional_body = capture do %>
+  <% if @content_item.corporate_information? %>
+    <%= @content_item.corporate_information_heading_tag %>
+    <% @content_item.corporate_information.each do |group| %>
+      <%= group[:heading] %>
+      <ul>
+        <% group[:links].each do |link| %>
+          <li>
+            <%= link %>
+          </li>
+        <% end %>
+      </ul>
+    <% end %>
+
+    <% if @content_item.further_information.present? %>
+      <p>
+        <%= @content_item.further_information %>
+      </p>
+    <% end %>
+  <% end %>
+<% end %>
+
 <div class="<%= @content_item.organisation_brand_class %>">
   <div class="grid-row">
-    <div class="column-quarter">
+    <div class="column-two-thirds">
       <%= render 'govuk_component/organisation_logo', @content_item.organisation_logo %>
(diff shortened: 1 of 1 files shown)


app/views/content_items/corporate_information_page.html.erb
```diff
@@ -1,36 +1,56 @@
...
-          <%= @content_item.further_information %>
-        </p>
-      <% end %>
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <%= render 'govuk_component/title',
+         title: @content_item.title,
+         average_title_length: "long" %>
+    </div>
+    <% if @content_item.available_translations.length > 1 %>
```
vanitabarrett: Can we use the shared/translation partial here?

steventux:
Thanks @vanitabarrett I've fixed the link colour and used the translations partial. Here's an example with the translations nav https://government-frontend-pr-697.herokuapp.com/government/organisations/companies-house/about/procurement.cy

fmlharrison:
Looks good to me...

fofr:
Looks like the title has accidentally got smaller?

fofr:
Will all the link colours pass accessibility contrast tests when tested against grey background?

screen shot 2018-01-16 at 12 27 51

This one doesn't meet our standard:
http://leaverou.github.io/contrast-ratio/#%2339836e-on-%23dee0e2

For our banner we say:
> have a text contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA

We should:
* add this as an accessibility acceptance criteria for the contents link when on grey too
* show examples of the contents link component with different link colours so automated tests can warn us of these accessibility fails

vanitabarrett:
@fofr @steventux looks like the title is smaller because for some reason we're passing 'average_title_length: "long"' to the title component.

alextea:
Could we override the colour of the link in the contents list so it is always the standard?

steventux:
@fofr @vanitabarrett @alextea will resolve these issues first, have reverted the merge.

PR #698, opened 2018-01-16
Improve published dates component

maxgds:
- Add a top border to separate from rest of page
- Reword open/closed states to be more explicit

---

Component guide for this PR:
https://government-frontend-pr-698.herokuapp.com/component-guide

Before:
screen shot 2018-01-16 at 09 32 47

After:
screen shot 2018-01-16 at 09 35 49

### DIFF
--- a/app/assets/stylesheets/components/_published-dates.scss
+++ b/app/assets/stylesheets/components/_published-dates.scss
@@ -15,3 +15,8 @@
   display: block;
   @include bold-16;
 }
+
+.app-c-published-dates--history {
+  padding-top: $gutter-one-third;
+  border-top: 1px solid $border-colour;
+}
--- a/app/views/components/_published-dates.html.erb
+++ b/app/views/components/_published-dates.html.erb
@@ -4,9 +4,10 @@
   history = Array(history)
   last_updated ||= false
   link_to_history ||= false
+  history_class = "app-c-published-dates--history" if history.any?
 %>
 <% if published || last_updated %>
-<div class="app-c-published-dates" <% if history.any? %>id="history" data-module="toggle"<% end %>>
+<div class="app-c-published-dates <%= history_class %>" <% if history.any? %>id="history" data-module="toggle"<% end %>>
   <% if published %>
     <%= t('components.published_dates.published', date: published) %>
   <% end %>
@@ -15,7 +16,11 @@
     <% if link_to_history && history.empty? %>
       &mdash; <a href="#history"><%= t('components.published_dates.see_all_updates') %></a>
     <% elsif history.any? %>
-      <a href="#full-history" data-controls="full-history" data-expanded="false">+ <%= t('components.published_dates.full_page_history') %></a>
+      <a href="#full-history"
+      data-controls="full-history"
+      data-expanded="false"
+      data-toggled-text="&#45;&nbsp;<%= t('components.published_dates.hide_all_updates') %>">&#43;&nbsp;<%= t('components.published_dates.show_all_updates')
+       %></a>
       <div class="app-c-published-dates__change-history js-hidden" id="full-history">
         <ol>
           <% history.each do |change| %>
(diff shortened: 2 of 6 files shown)


app/views/components/_published-dates.html.erb
```diff
@@ -15,7 +15,11 @@
     <% if link_to_history && history.empty? %>
       &mdash; <a href="#history"><%= t('components.published_dates.see_all_updates') %></a>
     <% elsif history.any? %>
-      <a href="#full-history" data-controls="full-history" data-expanded="false">+ <%= t('components.published_dates.full_page_history') %></a>
+      <a href="#full-history"
+      data-controls="full-history"
+      data-expanded="false"
+      data-toggled-text="&#43;&nbsp;<%= t('components.published_dates.hide_full_page_history') %>">&#43;&nbsp;<%= t('components.published_dates.show_full_page_history')
```
fofr: `"+ hide full page history"` doesn't make sense with the + and the hide.

fofr:
Is "full page history" something that users understand? "show history of changes"? Copy seems a little technical.

app/assets/stylesheets/components/_published-dates.scss
```diff
@@ -15,3 +15,8 @@
   display: block;
   @include bold-16;
 }
+
+.app-c-published-dates__history {
+  padding-top: 10px;
```
fofr: Should this be a variable?

app/views/components/_published-dates.html.erb
```diff
@@ -6,7 +6,7 @@
   link_to_history ||= false
 %>
 <% if published || last_updated %>
-<div class="app-c-published-dates" <% if history.any? %>id="history" data-module="toggle"<% end %>>
+<div <% if history.any? %>id="history" data-module="toggle" class="app-c-published-dates__history" <% else %> class="app-c-published-dates" <% end %>>
```
fofr: This isn't quite how BEM works:

`app-c-published-dates__history` should represent a history block within `app-c-published-dates`.

`app-c-published-dates` should always be present

Then if the "history" version is different it can be modified:

`app-c-published-dates app-c-published-dates--history`

app/assets/stylesheets/components/_published-dates.scss
maxgds: Maybe? There was a bare 10px for padding in the file, I just copied that. What would the variable be called if it were one?

app/views/components/_published-dates.html.erb
maxgds: Ok, I can review and fix this.

app/views/components/_published-dates.html.erb
```diff
@@ -6,7 +6,7 @@
   link_to_history ||= false
 %>
 <% if published || last_updated %>
-<div class="app-c-published-dates" <% if history.any? %>id="history" data-module="toggle"<% end %>>
+<div <% if history.any? %>id="history" data-module="toggle" class="app-c-published-dates app-c-published-dates--history" <% else %> class="app-c-published-dates" <% end %>>
```
vanitabarrett: I feel like there's a bit of repetition here with the classes, it could perhaps be simplified something like this:

`history_class = "app-c-published-dates--history" if history.any?`
`<div <% if history.any? %>id="history" data-module="toggle"<% end %> class="app-c-published-dates <%= history_class %>">`

maxgds:
@vanitabarrett @fofr I think I've incorporated all suggested changes and the updated text strings were supplied by the content team. Could you take another look over this?

vanitabarrett:
The colour of the border looks a lot paler than the design in the Trello ticket and compared to other borders on the page, e.g: the metadata at the top of the page. Is that intentional?

Example: https://government-frontend-pr-698.herokuapp.com/government/consultations/capacity-market-consultation-improving-the-framework-detailed-proposals#history

maxgds:
@vanitabarrett border colour was a mistake, now fixed.

fofr:
Probably worth indicating here that a border displays when there is a history toggle:
https://github.com/alphagov/government-frontend/blob/master/app/views/components/docs/published-dates.yml#L29

PR #707, opened 2018-01-18
Fix universal layout component ordering

emmabeynon:
For https://trello.com/c/LgveTTm5/197-ensure-components-are-consistently-ordered-in-all-formats

- In this PR we review all formats that are now using the universal layout to ensure that the ordering of components is correct.
- There were some common occurrences of components being in the wrong order, so below are some example screenshots of how the ordering has changed.  
- There were also a few occurrences where the important metadata component was present for formats that would never have that data populated, and there was a withdrawal notice present for a format that can't be withdrawn, so they have been removed in this instance.
- Finally there was one instance of translations code being used unnecessarily in a view, which has now been replaced by the shared translations partial.

### Important metadata should appear above an image

*Before*
fatality_notice_important_metadata_1

*After*
fatality_notice_important_metadata_2

### Withdrawn notices should appear below publisher metadata

*Before*
fatality_notice_withdrawn_1

*After*
fatality_notice_withdrawn_2

### History notices should appear below publisher metadata

*Before*
news_article_history_1

*After*
news_article_history_2

### When a history notice and withdrawal notice are present, the history notice should be above the withdrawal notice

*Before*
news_article_withdrawn_history_1

*After*
news_article_withdrawn_history_2

### Statistics notices should appear below important metadata

*Before*
statistics_announcement_cancelled_1

*After*
statistics_announcement_cancelled_2

---

Component guide for this PR:
https://government-frontend-pr-707.herokuapp.com/component-guide

### DIFF
--- a/app/views/content_items/case_study.html.erb
+++ b/app/views/content_items/case_study.html.erb
@@ -1,7 +1,6 @@
 <div class="grid-row">
   <div class="column-two-thirds responsive-top-margin">
     <%= render 'govuk_component/title', @content_item.title_and_context %>
-    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
   </div>
   <%= render 'shared/translations' %>
   <div class="column-two-thirds">
@@ -10,6 +9,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
+<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
 
 <div class="grid-row">
   <div class="column-two-thirds content-bottom-margin">
--- a/app/views/content_items/case_study.html.erb
+++ b/app/views/content_items/case_study.html.erb
@@ -19,12 +19,9 @@
           alt: @content_item.image["alt_text"],
           caption: @content_item.image["caption"] if @content_item.image %>
 
-      <%= render 'components/important-metadata',
-          items: @content_item.metadata[:other] %>
-
       <%= render 'govuk_component/govspeak',
-        content: @content_item.body,
-        direction: page_text_direction %>
+          content: @content_item.body,
+          direction: page_text_direction %>
     </div>
 
     <%= render 'components/published-dates', {
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -6,13 +6,8 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo', content_item: @content_item %>
-
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
-    <%= render 'shared/history_notice', content_item: @content_item %>
-  </div>
-</div>
+<%= render 'shared/history_notice', content_item: @content_item %>
+<%= render 'components/notice', @content_item.withdrawal_notice_component %>
 
 <div class="grid-row">
   <div class="column-two-thirds">
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -3,13 +3,16 @@
     <%= render 'govuk_component/title', @content_item.title_and_context %>
   </div>
   <%= render 'shared/translations' %>
+</div>
+
+<%= render 'shared/publisher_metadata_with_logo', content_item: @content_item %>
+
+<div class="grid-row">
   <div class="column-two-thirds">
     <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
     <%= render 'shared/history_notice', content_item: @content_item %>
   </div>
 </div>
-
-<%= render 'shared/publisher_metadata_with_logo', content_item: @content_item %>
 
 <div class="grid-row">
   <div class="column-two-thirds">
--- a/app/views/content_items/fatality_notice.html.erb
+++ b/app/views/content_items/fatality_notice.html.erb
@@ -1,7 +1,6 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <%= render 'govuk_component/title', @content_item.title_and_context %>
-    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
   </div>
   <%= render 'shared/translations' %>
   <div class="column-two-thirds">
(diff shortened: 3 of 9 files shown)


app/views/content_items/statistics_announcement.html.erb
```diff
@@ -9,15 +9,21 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
+    <%= render "components/important-metadata",
```
vanitabarrett: Have you chatted to @miaallers about this change? I know when we moved things over to the universal layout, she wanted the metadata and notice tied together if they both exist on the same page. This change introduces a space between the two.

emmabeynon: Ok, from the ordering diagram it looked like there was a space between the two, but I'll check with her 👍

app/views/content_items/publication.html.erb
```diff
@@ -8,12 +8,8 @@
        title: @content_item.title,
        average_title_length: "long" %>
   </div>
-  <% if @content_item.available_translations.length > 1 %>
-    <div class="column-third">
-      <%= render 'components/translation-nav',
-          translations: @content_item.available_translations %>
-    </div>
-  <% end %>
+  <%= render 'shared/translations' %>
```
steventux: :+1:

app/views/content_items/case_study.html.erb
```diff
@@ -19,12 +19,9 @@
           alt: @content_item.image["alt_text"],
           caption: @content_item.image["caption"] if @content_item.image %>
 
-      <%= render 'components/important-metadata',
-          items: @content_item.metadata[:other] %>
-
       <%= render 'govuk_component/govspeak',
-        content: @content_item.body,
-        direction: page_text_direction %>
+          content: @content_item.body,
```
steventux: Is this additional indentation intentional?

app/views/content_items/case_study.html.erb
```diff
@@ -19,12 +19,9 @@
           alt: @content_item.image["alt_text"],
           caption: @content_item.image["caption"] if @content_item.image %>
 
-      <%= render 'components/important-metadata',
```
steventux: :+1:

app/views/content_items/fatality_notice.html.erb
```diff
@@ -10,6 +9,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
+<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
```
steventux: @miaallers Is the notice component meant to be full width or 2/3rds? I think this will render it at full width.

app/views/content_items/fatality_notice.html.erb
```diff
@@ -14,15 +14,15 @@
...
       <%= render 'components/figure',
           src: @content_item.image["url"],
           alt: @content_item.image["alt_text"],
           caption: @content_item.image["caption"] if @content_item.image %>
-      <%= render 'components/important-metadata',
-          items: @content_item.metadata[:other] %>
       <%= render 'govuk_component/govspeak',
-        content: @content_item.body,
-        direction: page_text_direction %>
+          content: @content_item.body,
```
steventux: Indentation looks different to the rest of this template, I don't mind which way it goes, so long as linting is happy and it's consistent.

app/views/content_items/news_article.html.erb
```diff
@@ -11,6 +10,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
+<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
```
steventux: @miaallers similarly this looks like it will render notices at full width not 2/3rds width. Is this right?

app/views/content_items/publication.html.erb
```diff
@@ -12,12 +12,12 @@
 
   <div class="column-two-thirds">
     <%= render 'govuk_component/lead_paragraph', text: @content_item.description %>
-    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
     <%= render 'shared/history_notice', content_item: @content_item %>
   </div>
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
+<%= render 'components/notice', @content_item.withdrawal_notice_component  %>
```
steventux: We need to check with @miaallers about full width vs. 2/3rds width for these notices.

app/views/content_items/specialist_document.html.erb
```diff
@@ -3,7 +3,6 @@
 <div class="grid-row">
   <div class="column-two-thirds responsive-top-margin">
     <%= render 'govuk_component/title', @content_item.title_and_context %>
-    <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
```
steventux: :+1:

app/views/content_items/case_study.html.erb
emmabeynon: Yeah, just to match it up with the indentation when rendering the figure component on line 17.  I think it looks neater like this.

app/views/content_items/fatality_notice.html.erb
emmabeynon: It does render full width, which is how it looks in Alex's ordering diagram.  Will run it past Mia.

emmabeynon:
@steventux Mia's happy to keep the notices at full width, as they are likely to be redesigned in the near future 👍

app/views/content_items/statistics_announcement.html.erb
emmabeynon: Mia confirmed there should be no space between them, so I'll change that now 🚀

app/views/content_items/consultation.html.erb
```diff
@@ -3,14 +3,17 @@
     <%= render 'govuk_component/title', @content_item.title_and_context %>
   </div>
   <%= render 'shared/translations' %>
+</div>
+
+<%= render 'shared/publisher_metadata_with_logo', content_item: @content_item %>
+
+<div class="grid-row">
   <div class="column-two-thirds">
     <%= render 'components/notice', @content_item.withdrawal_notice_component  %>
```
steventux: These notices will be 2/3rds width which doesn't seem consistent? Is this a special case?

emmabeynon: Ah that's one I missed - will fix now 👍

app/views/content_items/fatality_notice.html.erb
miaallers: Yep!

miaallers:
The links all look correct to me! We will definitely need to update the styling on the withdrawn / history notices and banners, as well as the important metadata to get everything look good.

PR #708, opened 2018-01-18
Adds test for new specialist doc type

elenatanasoiu:
## What this does 
Updates tests on government-frontend to check for a new document type for specialist documents: `residential_property_tribunal_decision`

## Caveat

Depends on: https://github.com/alphagov/govuk-content-schemas/pull/705 being merged and deployed.

For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

---

Component guide for this PR:
https://government-frontend-pr-708.herokuapp.com/component-guide

### DIFF
--- a/test/integration/specialist_document_test.rb
+++ b/test/integration/specialist_document_test.rb
@@ -21,6 +21,7 @@ class SpecialistDocumentTest < ActionDispatch::IntegrationTest
     setup_and_visit_content_item('international-development-funding')
     setup_and_visit_content_item('maib-reports')
     setup_and_visit_content_item('raib-reports')
+    setup_and_visit_content_item('residential-property-tribunal-decision')
     setup_and_visit_content_item('service-standard-report')
     setup_and_visit_content_item('tax-tribunal-decision')
     setup_and_visit_content_item('utaac-decision')
@@ -28,6 +29,14 @@ class SpecialistDocumentTest < ActionDispatch::IntegrationTest
 
   test "renders title, description and body" do
     setup_and_visit_content_item('aaib-reports')
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+    assert_has_component_govspeak(@content_item["details"]["body"])
+  end
+
+  test "returns example for residential tribunal decision" do
+    setup_and_visit_content_item('residential-property-tribunal-decision')
 
     assert_has_component_title(@content_item["title"])
     assert page.has_text?(@content_item["description"])


elenatanasoiu:
Tests will fail until alphagov/govuk-content-schemas#705 is merged

test/integration/specialist_document_test.rb
```diff
@@ -3,9 +3,23 @@
...
+    setup_and_visit_random_content_item(document_type: 'esi_fund')
+    setup_and_visit_random_content_item(document_type: 'international_development_fund')
+    setup_and_visit_random_content_item(document_type: 'maib_report')
+    setup_and_visit_random_content_item(document_type: 'medical_safety_alert')
     setup_and_visit_random_content_item(document_type: 'raib_report')
+    setup_and_visit_random_content_item(document_type: 'residential_property_tribunal_decision')
+    setup_and_visit_random_content_item(document_type: 'service_standard_report')
     setup_and_visit_random_content_item(document_type: 'tax_tribunal_decision')
-    setup_and_visit_random_content_item(document_type: 'cma_case')
+    setup_and_visit_random_content_item(document_type: 'utaac_decision')
```
h-lame: Turns out, this `setup_and_visit_random_content_item` thing doesn't do what we want or expect in this case.  It calls `GovukSchemas::RandomExample.for_schema()` asking for the `frontend_schema` of the `document_type` we've requested.  Or, that's what we want it to do, but the implementation of `setup_and_visit_random_content_item` is broken.

```ruby
  def setup_and_visit_random_content_item(document_type: schema_type)
    payload = GovukSchemas::RandomExample.for_schema(frontend_schema: schema_type)
    # rest of method
```
should be:

```ruby
  def setup_and_visit_random_content_item(document_type: schema_type)
    payload = GovukSchemas::RandomExample.for_schema(frontend_schema: document_type)
    # rest of method
```

The method expects a keyword argument called `document_type` and defaults it to the value of `schema_type` (defined https://github.com/alphagov/government-frontend/blob/9483d27e7e96eec21e4fc4c5348d394c0c9b9d5e/test/test_helper.rb#L212-L214 and will be `"specialist_document"` for this test).  Inside the method we should use the variable `document_type` if we want the value passed in as the keyword argument, but we're using `schema_type` instead, so in this test, when we call the method we always get a random `specialist_document` when we want a random `aaib-report`, or `utaac_decision` or whatever.

The problem is, that none of the values we're passing in as `document_type` are actually valid schemas.  They're all ... subtypes? (I don't know what the nomenclature is here) ... of the schema `specialist_document`.  There's no way in govuk_schema (as far as I can see) for us to generate a random example of a subtype of `specialist_document`.  The best thing to do here might be to change all these `setup_and_visit_random_content_item` into `setup_and_visit_content_item` because these will read the examples from govuk-content-schemas and we aim to provide an example of each possible subtype of `specialist_document`.

Maybe we also want to look into making it possible to generate random examples of `specialist_document` subtypes, but I'm a bit wary of introducing randomness into our tests like this, because it's almost a guarantee of a intermittent test failure.  The downside to using the examples is that we rely on those examples being complete examples, and we might end up only correctly rendering those, not any valid, but edgecasey real examples.

elenatanasoiu: I've split out fixing these tests in a different PR, which you can see here: https://github.com/alphagov/government-frontend/pull/712

elenatanasoiu: I will leave this PR with the minimal change that is adding tests for `residential-property-tribunal-decision`. Tests are green when using Murray's alphagov/govuk-content-schemas#705 branch of govuk-content-schemas

h-lame:
Assume we'll rebase this on master once #712 is merged so that the diff is just adding `residential-property-tribunal-decision` to the `specialist document subtypes do not error` test.

PR #709, opened 2018-01-18
Move topical event about page to universal layout

emmabeynon:
For https://trello.com/c/8MY4DtDQ/203-move-topical-event-about-page-to-universal-layout

#### Example
https://government-frontend-pr-709.herokuapp.com/government/topical-events/anti-corruption-summit-london-2016/about

In this PR we move "Topical Event About Pages" to use the universal layout.

**Before**
topical_event_about_page_before

**After**
topical_event_about_page_after

---

Component guide for this PR:
https://government-frontend-pr-709.herokuapp.com/component-guide


fofr:
Aside: Screenshot here strongly suggests we do some sort of content check on whether to show a list of contents. Seems that it should be inferred from content itself rather than format. cc @miaallers

steventux:
@fofr good point, I'm interested to know how many items in the content store have `contents.size < 2`

miaallers:
Yeah I agree that if we have a content list of less than three items it looks odd. Can we hide the contents in cases like that?

RyanMacG:
:wave: just wanted to quickly pop in and let you know why the e2e tests are failing; the "full page history" link was removed earlier today in another PR and the tests were updated to match, should be solved with a rebase.

fofr:
@miaallers @steventux Probably needs to be a little smarter than content list item count, consider a page with two headings but a long preamble. Some sort of content length heuristic needed.

PR #712, opened 2018-01-19
Fixes and expands tests for specialist doc types

elenatanasoiu:
For: https://trello.com/c/qzSMAFHZ/6-3-moj-residential-tribunals-finder

As a consequence of adding a test for a new doc type: `residential-property-tribunal-decision`, we discovered that the tests were not doing what was expected (generate a random example of a certain document type).
This has been fixed in this PR.

We've also added more tests to include all of the specialist doc types available.

---

Component guide for this PR:
https://government-frontend-pr-712.herokuapp.com/component-guide

### DIFF
--- a/test/integration/specialist_document_test.rb
+++ b/test/integration/specialist_document_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 
 class SpecialistDocumentTest < ActionDispatch::IntegrationTest
-  test "random but valid items do not error" do
+  test "random but valid specialist documents do not error" do
     setup_and_visit_random_content_item(document_type: 'aaib_report')
     setup_and_visit_random_content_item(document_type: 'raib_report')
     setup_and_visit_random_content_item(document_type: 'tax_tribunal_decision')
--- a/test/integration/specialist_document_test.rb
+++ b/test/integration/specialist_document_test.rb
@@ -8,6 +8,24 @@ class SpecialistDocumentTest < ActionDispatch::IntegrationTest
     setup_and_visit_random_content_item(document_type: 'cma_case')
   end
 
+  test "specialist document subtypes do not error" do
+    setup_and_visit_content_item('aaib-reports')
+    setup_and_visit_content_item('asylum-support-decision')
+    setup_and_visit_content_item('business-finance-support-scheme')
+    setup_and_visit_content_item('cma-cases')
+    setup_and_visit_content_item('countryside-stewardship-grants')
+    setup_and_visit_content_item('drug-safety-update')
+    setup_and_visit_content_item('employment-appeal-tribunal-decision')
+    setup_and_visit_content_item('employment-tribunal-decision')
+    setup_and_visit_content_item('european-structural-investment-funds')
+    setup_and_visit_content_item('international-development-funding')
+    setup_and_visit_content_item('maib-reports')
+    setup_and_visit_content_item('raib-reports')
+    setup_and_visit_content_item('service-standard-report')
+    setup_and_visit_content_item('tax-tribunal-decision')
+    setup_and_visit_content_item('utaac-decision')
+  end
+
   test "renders title, description and body" do
     setup_and_visit_content_item('aaib-reports')
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -220,13 +220,15 @@ class ActionDispatch::IntegrationTest
     end
   end
 
-  def setup_and_visit_random_content_item(document_type: schema_type)
-    payload = GovukSchemas::RandomExample.for_schema(frontend_schema: schema_type)
-    path = payload["base_path"]
+  def setup_and_visit_random_content_item(document_type: nil)
+    content_item = GovukSchemas::RandomExample.for_schema(frontend_schema: schema_type) do |payload|
+      payload.merge('document_type' => document_type) unless document_type.nil?
+      payload
+    end
+    path = content_item["base_path"]
 
     stub_request(:get, %r{#{path}})
-      .to_return(status: 200, body: payload.to_json, headers: {})
-
+      .to_return(status: 200, body: content_item.to_json, headers: {})
     visit path
   end
 


test/integration/specialist_document_test.rb
```diff
@@ -2,10 +2,22 @@
 
 class SpecialistDocumentTest < ActionDispatch::IntegrationTest
   test "random but valid items do not error" do
```
h-lame: Minor: these are no longer random, so we could rename this test to: `example items do not error`

test/integration/specialist_document_test.rb
```diff
@@ -2,10 +2,22 @@
...
+    setup_and_visit_content_item('cma-cases')
+    setup_and_visit_content_item('countryside-stewardship-grants')
+    setup_and_visit_content_item('drug-safety-update')
+    setup_and_visit_content_item('employment-appeal-tribunal-decision')
+    setup_and_visit_content_item('employment-tribunal-decision')
+    setup_and_visit_content_item('european-structural-investment-funds')
+    setup_and_visit_content_item('international-development-funding')
+    setup_and_visit_content_item('maib-reports')
+    setup_and_visit_content_item('raib-reports')
+    setup_and_visit_content_item('residential-property-tribunal-decision')
```
h-lame: This line should probably be introduced by the other PR #708 which we'll rebase on top of this one I guess?

test/integration/specialist_document_test.rb
fofr: Shouldn't they still be random?

test/integration/specialist_document_test.rb
elenatanasoiu: Sure

test/integration/specialist_document_test.rb
fofr: The aim of the random test is to ensure that the app doesn't error on anything regarded as acceptable by the schema.

fofr:
Can we keep the random tests in please? https://github.com/alphagov/government-frontend/pull/712#discussion_r162606834

test/integration/specialist_document_test.rb
h-lame: But the `setup_and_visit_random_content_item` doesn't work for specialist document sub types, it only works for top level schemas.  So perhaps we replace this test implementation with just `setup_and_visit_random_content_item(document_type: 'specialist_document')` and keep the random in the name, and add a separate new test with this new implementation and we get both bits of coverage.

h-lame: However, I think the random test is dangerous because if it does fail, chances are someone will simply re-run the test and see it pass and move on with their life, not fixing the randomly generated error.  We'd need the failure to output some very clear instructions on how to re-run it with the same data so that we can debug it and fix it.  Perhaps it could output the content_item it used so we can copy paste it or something?

elenatanasoiu: As per @fofr's request, I've left the random examples in as a generic `specialist_document`. If we chose specific examples, these would fail because there aren't any schemas, as Murray explained above. The reason these random example have passed so far was because the `setup_and_visit_random_content_item` method doesn't fail when it can't find the schema you want. It will instead default to fetching a random `specialist_document` example. 

So might as well make it do that in the first place.

elenatanasoiu: Also, the `setup_and_visit_random_content_item` was doubly broken, because it was using `schema_type` instead of the parameter `document_type`, which has been fixed here: https://github.com/alphagov/government-frontend/pull/712/files#diff-f14ca9959ff06f3ca2ee6b9564866954L194. 

So it was failing because 
1. mistake in the method which I've fixed now 
2. once it actually attempts to look for any of the original 4 schema types, it won't find them because they are subtypes of `specialist_document` and will never get their own schema. 
Original 4 random schema types used:

```ruby
setup_and_visit_random_content_item(document_type: 'aaib_report')
setup_and_visit_random_content_item(document_type: 'raib_report')
setup_and_visit_random_content_item(document_type: 'tax_tribunal_decision')
setup_and_visit_random_content_item(document_type: 'cma_case')
```

elenatanasoiu: any thoughts @fofr ?

fofr: I think this might have broken in: https://github.com/alphagov/government-frontend/commit/bbc66a44a9b99ebbb251df93d46ccf1ff84455d9#diff-f14ca9959ff06f3ca2ee6b9564866954R194 

It looks like the ability to specify a custom document_type was removed there.

h-lame: Looks like it, I wonder if it's because it would fail occasionally.  The old version would create a random `specialist_document` and then change the `document_type` to the one you supplied.  I think it's very likely that this could produce a document that wasn't valid.  The metadata you need for an `aaib-report` is not the same as the metadata you need for a `utaac-decision` so if `RandomExample` gave you an `aaib-report` and you simply set the `document_type` on it to `utaac-decision`, it would expect it would no longer be valid.

test/integration/publication_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class PublicationTest < ActionDispatch::IntegrationTest
   test "random but valid items do not error" do
-    setup_and_visit_random_content_item(document_type: 'statutory_guidance')
+    setup_and_visit_random_content_item(frontend_schema: 'statutory_guidance')
```
elenatanasoiu: @fofr Following our discussion, I've renamed the `document_type` param to `frontend_schema` since that's what it's actually used for. This makes it clearer what we're requesting as a random example.

test/test_helper.rb
```diff
@@ -189,9 +189,9 @@ def setup_and_visit_content_item(name, parameter_string = '')
       visit("#{item['base_path']}#{parameter_string}")
     end
   end
-
-  def setup_and_visit_random_content_item(document_type: schema_type)
-    payload = GovukSchemas::RandomExample.for_schema(frontend_schema: schema_type)
+  
+  def setup_and_visit_random_content_item(frontend_schema: schema_type)
+    payload = GovukSchemas::RandomExample.for_schema(frontend_schema: frontend_schema)
```
elenatanasoiu: @fofr Following our discussion, I've renamed the document_type param to frontend_schema since that's what it's actually used for. This makes it clearer what we're requesting as a random example.

test/integration/specialist_document_test.rb
h-lame: Turns out I'm wrong.  The json schema doesn't validate that the `document_type` and `metadata` match up in the `specialist_document` schema.  This feels wrong as a "utaac_decision" with the metadata of an "aaib-report" is not correct even if it is valid, but also, I looked at how to make our json schema do this and it seemed very painful.  I withdraw my objection here.

PR #713, opened 2018-01-19
Move guides to universal layout

maxgds:
- pull in new nav sidebar to replace old
- align nav with the body of the content (below parts list
  where applicable)
- remove unused variant on guides

---
Before:
screen shot 2018-01-19 at 14 36 17
After:
screen shot 2018-01-19 at 14 36 43

Before:
screen shot 2018-01-19 at 14 36 58

After:
screen shot 2018-01-19 at 14 37 11


maxgds:
Note, to view the task list variant of pages, you will need to manipulate headers, I use a chrome plugin for this:
screen shot 2018-01-19 at 15 10 38

app/assets/stylesheets/views/_guide.scss
```diff
@@ -21,4 +21,12 @@
       @include bold-24;
     }
   }
+
+  .task-list-sidebar {
+    margin-top: 0;
```
vanitabarrett: Should this follow the pattern we've used for other components when it comes to top/bottom margins? Example: https://github.com/alphagov/government-frontend/blob/master/app/views/components/_notice.html.erb#L5

This means we don't have to override component styles in specific cases.

app/views/content_items/guide.html.erb
```diff
@@ -2,11 +2,35 @@
...
+          disable_youtube_expansions: true,
+          rich_govspeak: true %>
+      <%= render 'govuk_component/previous_and_next_navigation', @content_item.previous_and_next_navigation %>
+
+      <% if @content_item.multi_page_guide? %>
+        <%= render 'components/print-link', href: @content_item.print_link, link_text: t("multi_page.print_entire_guide") %>
+      <% end %>
+    <% end %>
   </div>
   <% if current_tasklist_ab_test.show_tasklist_sidebar? %>
```
vanitabarrett: `shared/sidebar_navigation` already contains this check, so I think you should be able to delete this block and just import `shared/sidebar_navigation`?

maxgds:
Thanks @vanitabarrett - making the change to just call the sidebar negated the need for me to call it with no margin, as that was already built in and being used by the sidebar. Could you take another quick look over it?

vanitabarrett:
@maxgds changes look good! 👍

I noticed this format seem to suffer from the same issue we had with answers (showing links to whitehall and mainstream topics with the same names). Not sure if a decision has been made around this? Example: http://government-frontend.dev.gov.uk/log-in-file-self-assessment-tax-return

maxgds:
It's going to be put into blocked pending content team conversations and then adjusted or released as appropriate, so I'm not looking to merge it just yet, just have it in a theoretical "could be merged" state.

PR #706, opened 2018-01-17
Move detailed guides to new layout

steventux:
https://trello.com/c/uScvU8q2/189-add-detailedguide-to-universal-layout

Depends on https://github.com/alphagov/govuk-content-schemas/pull/706
(PR number jinx!)

### Example

https://government-frontend-pr-706.herokuapp.com/guidance/classifying-vehicles

#### Before

screenshot from 2018-01-17 17-58-14

#### After

screenshot from 2018-01-17 17-59-38

---

Component guide for this PR:
https://government-frontend-pr-706.herokuapp.com/component-guide

PR #719, opened 2018-01-23
Display contents lists under specific conditions

emmabeynon:
For https://trello.com/c/2UIR4Sen/210-investigate-how-to-deal-with-short-content-lists

We have identified that there are occasions where it's not appropriate to display a contents list, for example when there is only one item on the list.  Having spiked an approach for displaying a contents list dependent on certain conditions, this PR implements the approach.  We've reached the below criteria for displaying a contents list in conjunction with Si Stephens-Manassiev in the content team:

1) We won't show a contents list that has one item.
2) We will show a contents list with 2 items if the first item's character count is over 415.
3) We will show a contents list with 2 items if the first item contains a table with more than 13 rows.
4) We will show a contents list with 2 items if the first item contains an image and character count is over 224.
5) We will show a contents list with 2 items if the first item contains an image and a table with more than 6 rows.
6) We will show a contents list with 3 items or more.

**To Do**
- [x] Update detailed guides after https://github.com/alphagov/government-frontend/pull/706 has been merged.

Component guide for this PR:
https://government-frontend-pr-719.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -4,6 +4,7 @@ module ContentItem
     include TypographyHelper
 
     CHARACTER_LIMIT = 100
+    CHARACTER_LIMIT_WITH_IMAGE = 50
     TABLE_ROW_LIMIT = 13
 
     def contents
@@ -23,7 +24,8 @@ module ContentItem
       return false if contents_items.count < 2
       return true if contents_items.count > 2
       first_item_has_long_content? ||
-        first_item_has_long_table?
+        first_item_has_long_table? ||
+        first_item_has_image_and_long_content?
     end
 
   private
@@ -37,7 +39,7 @@ module ContentItem
     end
 
     def first_item_has_long_content?
-      first_item_content > CHARACTER_LIMIT
+      first_item_character_count > CHARACTER_LIMIT
     end
 
     def first_item_content
@@ -48,7 +50,11 @@ module ContentItem
         first_item_text += element.text if element.name == 'p'
         element = element.next_element
       end
-      first_item_text.length
+      first_item_text
+    end
+
+    def first_item_character_count
+      @first_item_character_count ||= first_item_content.length
     end
 
     def first_item_has_long_table?
@@ -66,6 +72,19 @@ module ContentItem
         return element if element.name == 'table'
         element = element.next_element
       end
+    end
+
+    def first_item_has_image?
+      element = first_item.next_element
+
+      until element.name == 'h2'
+        return true if element.name == 'div' && element['class'] == 'img'
+        element = element.next_element
+      end
+    end
+
+    def first_item_has_image_and_long_content?
+      first_item_has_image? && first_item_character_count > CHARACTER_LIMIT_WITH_IMAGE
     end
 
     def parsed_body
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -24,6 +24,8 @@ module ContentItem
     def show_contents_list?
       return false if contents_items.count < 2
       return true if contents_items.count > 2
+      return false if no_first_item?
+
       first_item_has_long_content? ||
         first_item_has_long_table? ||
         first_item_has_image_and_long_content? ||
@@ -100,6 +102,10 @@ module ContentItem
 
     def first_item
       parsed_body.css('h2').first
+    end
+
+    def no_first_item?
+      first_item.nil?
     end
   end
 end
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -3,24 +3,58 @@ module ContentItem
     include ActionView::Helpers::UrlHelper
     include TypographyHelper
 
+    CHARACTER_LIMIT = 100
+
     def contents
-      @contents ||= contents_items.each do |item|
-        item[:href] = "##{item[:id]}"
-      end
+      @contents ||=
+        if show_contents_list?
+          contents_items.each { |item| item[:href] = "##{item[:id]}" }
+        else
+          []
+        end
     end
 
     def contents_items
-      extract_headings_with_ids(body)
+      extract_headings_with_ids
+    end
+
+    def show_contents_list?
+      return false if contents_items.count < 2
+      return true if contents_items.count > 2
+      first_item_has_long_content? if first_item.next_element.name == 'p'
     end
 
   private
 
-    def extract_headings_with_ids(html)
-      headings = Nokogiri::HTML(html).css('h2').map do |heading|
+    def extract_headings_with_ids
+      headings = parsed_body.css('h2').map do |heading|
         id = heading.attribute('id')
         { text: strip_trailing_colons(heading.text), id: id.value } if id
       end
       headings.compact
+    end
+
+    def first_item_has_long_content?
+      first_item_content > CHARACTER_LIMIT
+    end
+
+    def first_item_content
+      element = first_item.next_element
+      first_item_text = ''
+
+      until element.name == 'h2'
+        first_item_text += element.text if element.name == 'p'
+        element = element.next_element
+      end
+      first_item_text.length
+    end
+
+    def parsed_body
+      @parsed_body ||= Nokogiri::HTML(body)
+    end
+
+    def first_item
+      parsed_body.css('h2').first
     end
   end
 end
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -3,8 +3,8 @@ module ContentItem
     include ActionView::Helpers::UrlHelper
     include TypographyHelper
 
-    CHARACTER_LIMIT = 100
-    CHARACTER_LIMIT_WITH_IMAGE = 50
+    CHARACTER_LIMIT = 415
+    CHARACTER_LIMIT_WITH_IMAGE = 224
     TABLE_ROW_LIMIT = 13
     TABLE_ROW_LIMIT_WITH_IMAGE = 6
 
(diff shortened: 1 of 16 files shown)


app/presenters/content_item/contents_list.rb
```diff
@@ -22,5 +28,19 @@ def extract_headings_with_ids(html)
       end
       headings.compact
     end
+
+    def first_item_has_long_content?
+      first_item_content.count > WORD_COUNT_LIMIT
+    end
+
+    def first_item_content
+      element = Nokogiri::HTML(body).css('p').first
```
fofr: Some questions about this part:

* what happens if the content begins with something that's not a paragraph, eg a table or image?
* is this relatively slow at the moment? how performant could we make it?

app/presenters/content_item/contents_list.rb
```diff
@@ -3,6 +3,7 @@ module ContentsList
     include ActionView::Helpers::UrlHelper
     include TypographyHelper
 
+    WORD_COUNT_LIMIT = 100
```
fofr: I wonder if total word count is a useful heuristic.

fofr:
Love this as a starting point.

Are there some example URLs showing the effect?

app/presenters/content_item/contents_list.rb
steventux: I think we can optimise this a bit by only parsing the body once with Nokogiri, building the contents list as the existing implementation does, but also populating the first _n_ chapters so we can check their length.
This https://stackoverflow.com/questions/3484874/how-to-split-a-html-document-using-nokogiri does something similar by building a hash of chapters.

test/integration/topical_event_about_page_test.rb
```diff
@@ -49,9 +49,54 @@ class TopicalEventAboutPageTest < ActionDispatch::IntegrationTest
...
+
 private
 
   def topical_event_end_date
     Date.parse(@content_item['links']['parent'][0]['details']['end_date'])
   end
+
+  def long_first_item_body
+    "<div class='govspeak'><h2 id='response-in-the-uk'>Item 1</h2>
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean facilisis
```
steventux: The [Faker gem](https://github.com/stympy/faker/blob/master/doc/lorem.md) is great for generating this sort of thing.

emmabeynon:
> Are there some example URLs showing the effect?

We've got this as an example of the contents list not showing if there's only one item: https://government-frontend-pr-719.herokuapp.com/government/topical-events/ukraine-reform-conference/about
Unfortunately there aren't any live examples of Topical Event About Pages with two items - I've fudged some data locally to check it works - but I know that Detailed Guides has some, so I can extend this to the view for this format and will update the PR with an example 👍

test/integration/topical_event_about_page_test.rb
emmabeynon: Nice one!  Will keep that in mind for production tests 👍

app/presenters/content_item/contents_list.rb
emmabeynon: Good point about the paragraphs - I'll try and find some Detailed Guides examples with other content and work something out.
Steve's got a suggestion for optimising the Nokogiri stuff below 👇

app/views/content_items/topical_event_about_page.html.erb
```diff
@@ -8,9 +8,15 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
-    <%= render "components/contents-list-with-body", {
-      contents: @content_item.contents
-    } do %>
+    <% if @content_item.show_content_list? %>
```
steventux: This might be better implemented in the `contents-list-with-body` component as there's already a check for `@content_item.contents.any?`

emmabeynon: Yeah definitely - I've noted it's not an ideal implementation in the PR description, and would need to investigate other formats using this component before making changes to it in prod.

emmabeynon:
@fofr @steventux I've added a condition for showing a contents list when a table has a certain number of rows, and a condition for when an image is present (I imagine we could use the same idea for videos).  Regarding images, I had a quick chat with Steve about this last night and we agreed that since it seems that images are always the same size, the simplest way to deal with them is to just show contents lists when they're present.

app/presenters/content_item/contents_list.rb
```diff
@@ -13,14 +16,76 @@ def contents_items
...
+      end
+      first_item_text.split
+    end
+
+    def first_item_has_long_table?
+      table = find_table
+      return unless table.present?
+
+      row_count = table.css('tr').count
+      row_count > 10
```
steventux: I know this PR is still quite experimental, so a reminder to use the relevant constant `TABLE_ROW_LIMIT` here.

emmabeynon: Yep, that'll be me messing around with row numbers and not replacing the constant :woman_facepalming:

app/presenters/content_item/contents_list.rb
```diff
@@ -13,14 +16,76 @@ def contents_items
...
+    end
+
+    def find_table
```
steventux: This looks very similar to the following method, wonder if the logic is similar enough for one method.

test/integration/topical_event_about_page_test.rb
fofr: We can also use the govuk_schema gem to generate random examples to test this with

emmabeynon:
@steventux @fofr I've updated this PR to be less spikey and more production-friendly, and added some new conditions based on conversations with Si who has been testing various content lengths with a focus on common mobile displays.

app/presenters/content_item/contents_list.rb
```diff
@@ -3,24 +3,112 @@ module ContentsList
...
+      first_item_table_rows > TABLE_ROW_LIMIT
+    end
+
+    def find_table
+      element = first_item.next_element
+
+      until element.name == 'h2' do
+        if element.name == 'table'
+          return element
+        else
```
steventux: Are these `else` conditions necessary given we `return` on the `if`?

app/presenters/content_item/contents_list.rb
```diff
@@ -3,24 +3,112 @@ module ContentsList
...
+    def first_item_has_image_and_long_content?
+      first_item_has_image? && first_item_word_count > WORD_COUNT_LIMIT_WITH_IMAGE
+    end
+
+    def first_item_has_image_and_long_table?
+      first_item_has_image? && first_item_table_rows > TABLE_ROW_LIMIT_WITH_IMAGE
+    end
+
+    def parsed_body
+      @parsed_body ||= Nokogiri::HTML(body)
```
steventux: :+1:

app/presenters/content_item/contents_list.rb
steventux: What @fofr said. Some formats will contain medical terminology with superlongtechnicalwords, some will contain more legalese. I think a char count might be a more honest metric.

app/presenters/content_item/contents_list.rb
```diff
@@ -3,24 +3,58 @@ module ContentsList
...
     end
 
     def contents_items
-      extract_headings_with_ids(body)
+      extract_headings_with_ids
+    end
+
+    def show_contents_list?
+      return false if contents_items.count < 2
+      return true if contents_items.count > 2
```
maxgds: what if contents_items.count == 2?

steventux: Line 24 would catch this case but only if it's a paragraph - @emmabeynon is it ever likely to be a 2 item list which doesn't contain paragraphs?

test/presenters/content_item/contents_list_test.rb
```diff
@@ -78,4 +78,45 @@ def body
       { text: "Four", id: "four" }
     ], @contents_list.contents_items
   end
+
+  test "#show_contents_list? returns true if number of contents items is less than 3 and the first item's character count is above 100" do
+    class << @contents_list
+      def body
+        "<h2 id='one'>One</h2>
```
maxgds: I'm not sure how we can avoid this, but it feels a little fragile. If somebody updates the heading structure and all the items in govspeak become h3s then this won't happen on the live site, but the tests will continue to pass.

app/presenters/content_item/contents_list.rb
```diff
@@ -49,6 +51,23 @@ def first_item_content
...
 
+    def first_item_has_long_table?
+      table = find_table
+      return unless table.present?
+
+      row_count = table.css('tr').count
+      row_count > TABLE_ROW_LIMIT
+    end
+
+    def find_table
```
maxgds: I think for clarity this should be called find_first_table - I went to check whether it was scoped when I saw it called in the previous method.

test/presenters/content_item/contents_list_test.rb
steventux: The [existing implementation relies on the presence of `h2`s](https://github.com/alphagov/government-frontend/blob/master/app/presenters/content_item/contents_list.rb#L19), it's not a problem we're solving here. It's also a pattern we use in Govspeak frequently for anchored headings.

app/presenters/content_item/contents_list.rb
emmabeynon: Subsequent commits deal with checking for content with a lot of table rows, and also content with a mix of images, long content and long tables.

app/presenters/content_item/contents_list.rb
steventux: @maxgds it's a common convention in ruby to suffix methods which return a boolean with `?`
It's also conventional (at least with ActiveRecord) to return an object of some sort with methods prefixed with `find_`.
Given this method returns a boolean and not a table element I think the method name is OK.

app/presenters/content_item/contents_list.rb
maxgds: Makes sense, I hadn't looked ahead enough, sorry.

maxgds:
Looks good overall - happy to pass this, although I would prefer that one method name be updated.

PR #722, opened 2018-01-25
Improve accessibility for metadata links

maxgds:
This adds the ability to optionally pass a "labelledby" string
to our link_to methods.

The important metadata and publisher metadata components
implement this as an "aria-labelledby" attribute which means
these links are contextualised for assistive tech.

---

Component guide for this PR:
https://government-frontend-pr-722.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/content_item/linkable.rb
+++ b/app/presenters/content_item/linkable.rb
@@ -3,7 +3,7 @@ module ContentItem
     include ActionView::Helpers::UrlHelper
 
     def from
-      organisations_ordered_by_importance + links_group(%w{worldwide_organisations ministers speaker})
+      organisations_ordered_by_importance(labelledby: 'metadata-from') + links_group(%w{worldwide_organisations ministers speaker}, labelledby: 'metadata-from')
     end
 
     def part_of
@@ -20,9 +20,9 @@ module ContentItem
 
   private
 
-    def links(type)
+    def links(type, labelledby: nil)
       expanded_links_from_content_item(type).map do |link|
-        link_for_type(type, link)
+        link_for_type(type, link, (labelledby.nil? ? {} : { labelledby: labelledby }))
       end
     end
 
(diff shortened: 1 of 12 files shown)


fofr:
@maxgds Can you provide more detail to clarify why each link needs a labelledby attribute to improve accessibility?

app/presenters/content_item/linkable.rb
```diff
@@ -3,7 +3,7 @@ module Linkable
     include ActionView::Helpers::UrlHelper
 
     def from
-      organisations_ordered_by_importance + links_group(%w{worldwide_organisations ministers speaker})
+      organisations_ordered_by_importance(labelledby: 'metadata-from') + links_group(%w{worldwide_organisations ministers speaker}, labelledby: 'metadata-from')
```
fofr: Rather than put the `labelledby` parameter down through all these methods, would it make more sense for the original methods to return an array of link titles/paths that we can iterate over at a higher level?

The hardcoded HTML ID in this helper is a long way from where that ID is defined

app/views/components/_important-metadata.html.erb
```diff
@@ -12,7 +12,8 @@
     <% end %>
     <dl data-module="track-click">
     <% items.each do |title, definition| %>
-      <dt class="app-c-important-metadata__term"><%= title %>: </dt>
+      <% metadata_id = definition.to_s =~ /href/ ?  "id=metadata-" + title.to_s.parameterize : "" %>
```
steventux: As @fofr mentions, the coupling of defining this id attribute and defining the aria attribute need to be much closer otherwise the code will be brittle and prone to failure in future changes.
Possibly a helper to prepare the definition list would do the job or moving the links back to being hashes of text and path so that the attributes could be applied at the component template level.

app/views/components/_publisher-metadata.html.erb
```diff
@@ -18,7 +18,8 @@
             values = Array(values)
           %>
           <% if values.any? %>
-            <dt class="app-c-publisher-metadata__term"><%= title %>: </dt>
+            <% metadata_id = values.to_s =~ /href/ ?  "id=metadata-" + title.to_s.parameterize : "" %>
```
steventux: I think the need to do this reinforces the earlier point that we need to bring the attribute-adding-code nearer to the component ERB template.

Issue #437, opened 2017-08-10
Rescue from GdsApi::HTTPGone


kevindew:
There are rescue blocks for a bunch of the status codes that content store can return in [content_items_controller.rb](https://github.com/alphagov/government-frontend/blob/master/app/controllers/content_items_controller.rb#L4-L6) however there isn't one to handle a 410 response.

With the [Publishing E2E Tests](https://github.com/alphagov/publishing-e2e-tests) we've seen an issue occur which causes government frontend to 500:

- Content is unpublished, but hasn't reached content-store yet
- A user requests a piece of content
- Router sends user to government-frontend
- The content-store is updated with the unpublishing
- government-frontend accesses content-store and receives a [410 response](https://github.com/alphagov/content-store/blob/master/app/controllers/content_items_controller.rb#L135) 
- government frontend doesn't know how to rescue a `GdsApi::HTTPGone` error, so raises a 500

I haven't been able to find any evidence of this happening in production environment.

fofr:
This may also be true for redirects:
https://errbit.publishing.service.gov.uk/apps/546493f90da1150e4e001cfc/problems/598d8f8c65786378a687c400

kevindew:
I've got https://github.com/alphagov/government-frontend/pull/725 open which should resolve the immediate issue in this card.

The related issue with redirects seems a thornier issue to resolve since Content Store doesn't return a list of redirects with a response and we would need some logic to determine which one to use for a redirect.

PR #725, opened 2018-01-29
Return 410 when Content Store returns 410

kevindew:
Addresses: https://github.com/alphagov/government-frontend/issues/437

Rather than returning a 500 exception for a gone piece of content this
will return a 410 result.

The scenario this can occur in is when the following sequence occurs

- A content item is updated in the content-store to be gone
- A request is made for that piece of content, the router proxies to
  government-frontend
- The content-store then updates the router that the content is gone
- Government-frontend requests the content from the content-store
- The content-store responds with a 410 response
- Government-frontend throws a 500 exception

A similar scenario also can occur with redirects however these are less
trivial to resolve.

---

Component guide for this PR:
https://government-frontend-pr-725.herokuapp.com/component-guide


kevindew:
Thanks @steventux 👍

PR #726, opened 2018-01-30
Protect against missing org

maxgds:
Fixes an issue where 
/government/organisations/counter-fraud-and-security-management-service/about
is throwing errors due to having no organisation set.

eg

https://sentry.io/govuk/app-government-frontend/issues/436288507/

Pages render on live and error silently, but can be seen in a broken/fixed state on dev, eg:
http://government-frontend.dev.gov.uk/government/organisations/counter-fraud-and-security-management-service/about was failing before this fix.

### DIFF
--- a/app/presenters/content_item/organisation_branding.rb
+++ b/app/presenters/content_item/organisation_branding.rb
@@ -38,12 +38,14 @@ module ContentItem
     # Remove this in the future after migrating organisations to the content store API,
     # and updating them with the correct brand in the actual store.
     def organisation_brand(organisation)
+      return unless organisation
       brand = organisation["details"]["brand"]
       brand = "executive-office" if executive_order_crest?(organisation)
       brand
     end
 
     def executive_order_crest?(organisation)
+      return unless organisation
       organisation["details"]["logo"]["crest"] == "eo"
     end
   end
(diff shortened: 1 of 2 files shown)


fofr:
Can you include a test to indicate what will happen when no org is set?

steventux:
https://www.gov.uk/api/content/government/organisations/counter-fraud-and-security-management-service/about there appears to be an org `content_id` in the `details` hash, looking at the `updated_at` value this item was recently updated in the content store. Other corporate information pages seem to have organisation links, I wonder if we can reproduce this on integration by updating a CIP?

app/presenters/content_item/organisation_branding.rb
```diff
@@ -38,13 +38,13 @@ def default_organisation
     # Remove this in the future after migrating organisations to the content store API,
     # and updating them with the correct brand in the actual store.
     def organisation_brand(organisation)
-      brand = organisation["details"]["brand"]
-      brand = "executive-office" if executive_order_crest?(organisation)
+      brand = organisation["details"]["brand"] if organisation
+      brand = "executive-office" if executive_order_crest?(organisation) if organisation
```
steventux: Chaining `if` conditionals like this isn't valid syntax, it looks different from the change in https://github.com/alphagov/government-frontend/pull/726/commits/3015f55bde04e5c76d97bc6dff996e92ec90d4cd, I think adding:
```ruby
return unless organisation
```
might be cleaner than multiple if conditions throughout the method.

PR #740, opened 2018-02-06
Exclude all specialist date facets from finder link

steventux:
https://trello.com/c/nrAQ1eXm/198-iterate-related-navigation-to-show-other-similar-formats-from-the-same-organisation

Following up on https://github.com/alphagov/government-frontend/pull/739 we don't need to use any dates from metadata to filter the finders.
Date fields defined here: https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/_specialist_document.jsonnet 

---

Component guide for this PR:
https://government-frontend-pr-740.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/navigation/specialist_documents.rb
+++ b/app/presenters/navigation/specialist_documents.rb
@@ -2,13 +2,6 @@ module Navigation
   module SpecialistDocuments
     include Finders
 
-    PARAMS_TO_IGNORE = %w(
-      bulk_published
-      opened_date
-      closed_date
-      date_of_occurrence
-    ).freeze
-
     def finder_path_and_params
       "#{finder_path}?#{facet_params}"
     end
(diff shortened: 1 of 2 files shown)


app/presenters/navigation/specialist_documents.rb
```diff
@@ -3,10 +3,13 @@ module SpecialistDocuments
     include Finders
 
     PARAMS_TO_IGNORE = %w(
+      assessment_date
```
fofr: What happens if a finder adds a new type of date?
Don't facets have a type that could be checked?

steventux:
@fofr good point https://github.com/alphagov/government-frontend/pull/740/commits/5e3dbc7ab0db83cc4ef524614fe81e71818a6aab

PR #741, opened 2018-02-07
Update step by step navigation

sihugh:
This PR:
- removes task list AB tests
- renames task lists to step by step navs
- Updates references to components in govuk_publishing_components and govuk_navigation_helpers that have driven the first two points

govuk_navigation_helpers no longer contains AB testing logic for enabling/
disabling step by step navigation.  It also has a change to the format of the
data to pass to the task_list component.

govuk_publishing_components has the corresponding change to the format of the
data that task_list components expect, but also has renamed task_list elements
to step_by_step_nav.

We need to update this code to work with that.  It makes for an unwieldy review
but nothing will work if we do piecemeal commits. :-/

https://trello.com/c/3pQPJDmj/440-allow-collections-to-render-task-lists-from-the-publishing-api


app/controllers/application_controller.rb
```diff
@@ -5,19 +5,17 @@ class ApplicationController < ActionController::Base
...
-      request: request
-    )
-  end
-  helper_method :current_tasklist_ab_test
-
 private
 
-  def current_tasklist
-    GovukNavigationHelpers::TasklistContent.current_tasklist(request.path)
+  def current_step_nav
```
Davidslv: If we decide to keep using this we should introduce both this methods in the gem so all we have to do around all the frontend apps is to include the class. 
(Not something to do now)

sihugh: Yep, when we move to the published info, we can do a bit of a clean up again 👍

PR #746, opened 2018-02-09
Render European structural investment funds logo

steventux:
Responding to urgent ZD ticket https://govuk.zendesk.com/agent/tickets/2578725

screenshot from 2018-02-09 11-30-19

### Example

https://government-frontend-pr-746.herokuapp.com/guidance/england-2014-to-2020-european-structural-and-investment-funds

Publisher metadata should render the relevant logo next to it,
previously this worked for National Statistics but the European
structural investment funds logo was not tested and therefore
not included in the move to a new layout.
This commit adds a presenter unit test and a format specific
integration test to cover the rendering of this logo.

---

Component guide for this PR:
https://government-frontend-pr-746.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/detailed_guide_presenter.rb
+++ b/app/presenters/detailed_guide_presenter.rb
@@ -23,6 +23,12 @@ class DetailedGuidePresenter < ContentItemPresenter
     nav
   end
 
+  def logo
+    image = content_item.dig("details", "image")
+    return unless image
+    { path: image["url"], alt_text: "European structural investment funds" }
+  end
+
 private
 
   def related_links(key)
--- a/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
+++ b/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
@@ -12,6 +12,8 @@
   }
 
   .metadata-logo {
+    max-height: 90px;
+    max-width: 100%;
     padding-bottom: $gutter-half;
     padding-top: $gutter-half;
 
--- a/app/presenters/content_item/national_statistics_logo.rb
+++ b/app/presenters/content_item/national_statistics_logo.rb
@@ -1,0 +1,8 @@
+module ContentItem
+  module NationalStatisticsLogo
+    def logo
+      return unless national_statistics?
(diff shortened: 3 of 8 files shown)


app/presenters/detailed_guide_presenter.rb
```diff
@@ -23,6 +23,12 @@ def related_navigation
     nav
   end
 
+  def logo
+    image = content_item.dig("details", "image")
+    return unless image
+    { path: image["url"], alt_text: "European structural investment funds" }
```
fofr: Is there a way that this alt text can be put into the content item?

steventux: @fofr this bothered me too, I don't understand why we return a url in the content item data and not some alt text. It's not currently presented to the Publishing API https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/detailed_guide_presenter.rb#L64

fofr: Is it difficult to add? ("Image" is a little ambiguous too)

steventux: There are 6 detailed guides in the content store using one of the two variations of this logo. There are no other detailed guides using other images so the problem is relatively limited.
I will open a PR in Whitehall to add the alt text, in the mean time, to satisfy the fairly urgent ZD issue, hardwired alt text will be consistent with the previous version of this format.

steventux: Actually looking back at the previous version https://github.com/alphagov/government-frontend/blob/061c742a74af77c6fc55b80ac7ed42e0426dac13/app/views/content_items/detailed_guide.html.erb#L8 I don't see mention of alt text but the national archives version implies it was there: http://webarchive.nationalarchives.gov.uk/20171012232401/https://www.gov.uk/guidance/england-2014-to-2020-european-structural-and-investment-funds
very confusing.

steventux: Ah - https://apidock.com/rails/ActionView/Helpers/AssetTagHelper/image_tag the name part of the source is used. I'll add a PR to do this until we get this data from Whitehall.

PR #747, opened 2018-02-09
Fix contents list bugs

emmabeynon:
For https://trello.com/c/2UIR4Sen/210-investigate-how-to-deal-with-short-content-lists

After deploying https://github.com/alphagov/government-frontend/pull/719 we noticed some bugs for Document Collections and Working Groups, so this PR aims to fix thes

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -47,7 +47,7 @@ module ContentItem
     end
 
     def first_item_content
-      element = first_item.next_element
+      element = first_item
       first_item_text = ''
 
       until element.name == 'h2'
@@ -66,7 +66,7 @@ module ContentItem
     end
 
     def find_first_table
-      element = first_item.next_element
+      element = first_item
 
       until element.name == 'h2' do
         return element if element.name == 'table'
@@ -80,7 +80,7 @@ module ContentItem
     end
 
     def first_item_has_image?
-      element = first_item.next_element
+      element = first_item
 
       until element.name == 'h2'
         return true if element.name == 'div' && element['class'] == 'img'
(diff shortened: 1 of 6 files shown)


app/presenters/working_group_presenter.rb
```diff
@@ -33,4 +33,13 @@ def extra_headings
     extra_headings << { id: "contact-details", text: "Contact details" } if email.present?
     extra_headings
   end
+
+  def first_item
+    return if body.nil?
+    if parsed_body.css("h2").empty?
+      parsed_body.css("div").first.first_element_child
```
steventux: Similar to my question/comment above - is there a chance `parsed_body.css('div')` is ever empty or has no `first_element_child`?

app/presenters/content_item/contents_list.rb
```diff
@@ -101,7 +104,7 @@ def parsed_body
     end
 
     def first_item
-      parsed_body.css('h2').first
+      parsed_body.css('h2').first.next_element
```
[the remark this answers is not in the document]
emmabeynon: I think it would be an edge case for nothing to follow an h2, but I think we should add a guard to catch this in case it does happen 👍

app/presenters/working_group_presenter.rb
emmabeynon: My thinking here is that there would always be at least a govspeak div in the body, but as above we should probably guard against it just in case.

emmabeynon:
@steventux I've addressed these element edge cases; I'm happy for nil to be returned if the expected elements aren't present as this will mean that the contents list is not shown, which I think is correct because the absence of h2s or indeed any content suggests there shouldn't be a contents list anyway.

PR #753, opened 2018-02-12
Add support for blank hrefs

maxgds:
Guide chapter headings are changing to match the format of Content
Lists. Rather than creating a new component it makes sense to
reuse this one. The main difference is that the guides have contents
that span different urls and it isn't desirable to link a page
directly to itself. This adds support for a list item without a link
in a contents list.

Additionally, tracking has been added, as per https://trello.com/c/prpbSir0/486-create-or-adapt-a-frontend-component-for-mainstream-guides-chapters-with-tracking - all links are labelled as content_item or nested_content_item for tracking purposes. This was a requirement for Guide chapter headings so it made sense to apply it across the board.

screen shot 2018-02-12 at 11 23 10
---

Component guide for this PR:
https://government-frontend-pr-753.herokuapp.com/component-guide

### DIFF
--- a/app/views/components/_contents-list.html.erb
+++ b/app/views/components/_contents-list.html.erb
@@ -16,20 +16,49 @@
       <%= t("content_item.contents") %>
     </h2>
     <ol class="app-c-contents-list__list">
-      <% contents.each do |contents_item| %>
+      <% contents.each.with_index(1) do |contents_item, position| %>
+        <%
+            active_class = "app-c-contents-list__list-item--active" if contents_item[:active]
+            aria_current = "aria-current = true" if contents_item[:active]
+        %>
         <li
-          class="app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %>"
+          class="app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %> <%= active_class %>"
+          <%= aria_current %>
         >
-          <%= link_to contents_item[:href], class: 'app-c-contents-list__link' do %>
-            <%= format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
-          <% end %>
+          <% link_text = format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
+          <%= link_to_if !contents_item[:active], link_text, contents_item[:href],
+               class: 'app-c-contents-list__link',
+               data: {
+                 track_category: 'contentsClicked',
+                 track_action: "content_item #{position}",
+                 track_label: contents_item[:href],
+                 track_options: {
+                   dimension29: contents_item[:text]
+                  }
+                }
+          %>
           <% if contents_item[:items] && contents_item[:items].any? %>
             <ol class="app-c-contents-list__nested-list">
-              <% contents_item[:items].each do |nested_contents_item| %>
-                <li class="app-c-contents-list__list-item app-c-contents-list__list-item--dashed">
-                  <%= link_to nested_contents_item[:href], class: 'app-c-contents-list__link' do %>
-                    <%= nested_contents_item[:text] %>
-                  <% end %>
+              <% contents_item[:items].each.with_index(1) do |nested_contents_item, nested_position| %>
+              <%
+                active_class = "app-c-contents-list__list-item--active" if nested_contents_item[:active]
+                aria_current = "aria-current = true" if nested_contents_item[:active]
+              %>
+                <li
+                  class="app-c-contents-list__list-item app-c-contents-list__list-item--dashed <%= active_class %>"
+                  <%= aria_current %>
+                >
+                  <%= link_to_if !nested_contents_item[:active], nested_contents_item[:text], nested_contents_item[:href],
+                    class: 'app-c-contents-list__link',
+                    data: {
+                      track_category: 'contentsClicked',
+                      track_action: "nested_content_item #{position}:#{nested_position}",
+                      track_label: nested_contents_item[:href],
+                      track_options: {
+                        dimension29: nested_contents_item[:text]
+                      }
+                    }
+                  %>
                 </li>
               <% end %>
             </ol>
--- a/app/views/components/docs/contents-list.yml
+++ b/app/views/components/docs/contents-list.yml
@@ -15,6 +15,7 @@ accessibility_criteria: |
 
   - inform the user how many items are in the list
   - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
 
   Links with formatted numbers must separate the number and text with a space for correct screen reader pronunciation. This changes pronunciation from "1 dot Item" to "1 Item".
 shared_accessibility_criteria:
@@ -36,6 +37,17 @@ examples:
           text: Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
         - href: "#second-thing"
           text: Another pretty long contents list entry, not as long as the first, but still a little.
+        - href: "#third-thing"
+          text: Third thing
+  active content link:
+    description: 'An active state allows for "you are here" items in a contents list that spans different pages to avoid linking back to the current page.'
+    data:
+      contents:
+        - href: "#first-thing"
+          text: First
+        - href: "#two"
+          text: Second
+          active: true
         - href: "#third-thing"
           text: Third thing
   nested_contents_lists:
@@ -60,7 +72,8 @@ examples:
       contents:
         - href: "#first-thing"
           text: 1. First thing
-        - href: "#second-thing"
+        - href: "#two"
+          active: true
           text: 2. Second thing
         - href: "#third-thing"
           text: 3. Third thing
@@ -107,6 +120,7 @@ examples:
           - href: "#second-thing"
             text: No numbers here
           - href: "#third-thing"
+            active: true
             text: None here either
   right_to_left:
     data:
--- a/test/components/contents_list_component_test.rb
+++ b/test/components/contents_list_component_test.rb
@@ -12,6 +12,14 @@ class ContentsListComponentTest < ComponentTestCase
     ]
   end
 
+  def contents_list_with_active_item
+    [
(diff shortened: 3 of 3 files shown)


app/views/components/_contents-list.html.erb
```diff
@@ -20,8 +20,12 @@
         <li
           class="app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %>"
         >
-          <%= link_to contents_item[:href], class: 'app-c-contents-list__link' do %>
-            <%= format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
+          <% if contents_item[:href] == nil %>
```
steventux: There's probably no need to compare to nil here, you can use the truthiness of the return value of  `Hash#[]` to determine the condition.
What happens with numbers in the text only case? I see the links treat them differently. If the same treatment is needed it might be better to do this before deciding whether to render a link or not...

maxgds: Thanks for comments. Currently there isn't a case where you'd have numbers and no link, so I'd overlooked it, but yes, I'll go back and make sure that works as well (it likely doesn't right now).

app/views/components/_contents-list.html.erb
```diff
@@ -20,8 +20,12 @@
...
+            <%= contents_item[:text] %>
+          <% else %>
+            <%= link_to contents_item[:href], class: 'app-c-contents-list__link' do %>
```
tijmenb: I think you can also use [link_to_if](https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to_if):

```rb
<%= link_to_if contents_item[:href], contents_item[:href], class: 'app-c-contents-list__link' do %>
  <%= format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
<% end %>
```

maxgds:
@steventux @tijmenb Thanks for comments, could you take another look?

app/views/components/_contents-list.html.erb
```diff
@@ -16,20 +16,37 @@
       <%= t("content_item.contents") %>
     </h2>
     <ol class="app-c-contents-list__list">
-      <% contents.each do |contents_item| %>
+      <% contents.each_with_index do |contents_item, position| %>
```
vanitabarrett: If you use .each.with_index(1) the index will start from 1 and you won't need to +1 to position later on in the logic. (same goes for nested position)

app/views/components/docs/contents-list.yml
```diff
@@ -60,7 +71,7 @@ examples:
       contents:
         - href: "#first-thing"
           text: 1. First thing
-        - href: "#second-thing"
+        - href:
```
vanitabarrett: I'm probably lacking some context around exactly how this component will be used on guides and the kind of data we're getting back to feed into the component, but would it make more sense to support no href being passed in at all, rather than having to pass a blank/nil value?

test/components/contents_list_component_test.rb
```diff
@@ -35,6 +47,23 @@ def nested_contents_list
     assert_select ".app-c-contents-list__link[href='/two']", text: "2. Two"
   end
 
+  test "renders text only when link path is nil" do
+    render_component(contents: contents_list_with_blank_path)
+    assert_select ".app-c-contents-list"
+    assert_select ".app-c-contents-list__link[href='/one']", text: "1. One"
+    assert_select ".app-c-contents-list__link[href='/two']", count: 0
```
vanitabarrett: I feel like this line may be too specific and not actually testing what we want to test, as we're not passing a '/two' href anywhere it seems weird to check that it's not there. It may make more sense just to check that there is no link element?

app/views/components/docs/contents-list.yml
```diff
@@ -106,7 +117,7 @@ examples:
           items:
           - href: "#second-thing"
             text: No numbers here
-          - href: "#third-thing"
+          - href:
```
fofr: Could the API omit the `href` altogether?
```
- text: 'Thing'
- text: 'Another thing'
```

app/views/components/docs/contents-list.yml
```diff
@@ -38,6 +38,17 @@ examples:
           text: Another pretty long contents list entry, not as long as the first, but still a little.
         - href: "#third-thing"
           text: Third thing
+  blank_href:
```
fofr: I'd call this example 'Contents items without links' or 'Text only content items'
"Blank href" is not clear.

app/views/components/_contents-list.html.erb
```diff
@@ -16,20 +16,37 @@
...
           class="app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %>"
         >
-          <%= link_to contents_item[:href], class: 'app-c-contents-list__link' do %>
-            <%= format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
-          <% end %>
+          <% link_text = format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
+          <%= link_to_if contents_item[:href], link_text, contents_item[:href],
+               class: 'app-c-contents-list__link',
+               data: {
+                 track_category: 'contentsClicked',
```
fofr: Does this match the events used to track contents lists back in Q1 2017? It'd be good to keep the analytics data consistent.

fofr:
An alternative approach would be to pass in a flag, eg `active: true` or `current: true`. This would make the expected behaviour more obvious.

app/views/components/_contents-list.html.erb
maxgds: I've spoken to Tim who supplied the analytics tags and he says this is based on that, but deliberately not identical and he's happy to proceed with this tagging, so I'll leave this as is.

app/views/components/docs/contents-list.yml
maxgds: Yep, amending.

maxgds:
@vanitabarrett @fofr Thanks for your comments. I've refactored and pushed, would appreciate another look from you both.

app/views/components/_contents-list.html.erb
```diff
@@ -16,20 +16,47 @@
...
+          aria_current = contents_item[:active]? "aria-current = true" : ""
+        %>
         <li
-          class="app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %>"
-        >
-          <%= link_to contents_item[:href], class: 'app-c-contents-list__link' do %>
-            <%= format_numbers ? wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text] %>
-          <% end %>
+           class = "app-c-contents-list__list-item app-c-contents-list__list-item--<%= parent_list_item_modifier %> <%= active_class %>"
+           <%= aria_current %> >
```
vanitabarrett: I'm not sure the aria-current attribute is being read out where it is currently positioned on the `li` element. When I tested, Voiceover didn't read anything when the element was 'current'. It may need to be moved to the text itself?

steventux:
@maxgds I think the commit history could do with a bit of a cleanup.
https://github.com/alphagov/government-frontend/pull/753/commits/bd80cddde1a616ce257649b07b2ca9b8a31b4358 can be fixed up for clarity.

app/views/components/docs/contents-list.yml
```diff
@@ -38,6 +39,17 @@ examples:
...
         - href: "#third-thing"
           text: Third thing
+  active content link:
+    description: 'An active state allows for "you are here" items in a contents list that spans different pages to avoid linking back to the current page.'
+    data:
+      contents:
+        - href: "#first-thing"
+          text: First
+        - href: "#two"
+          text: Second
```
vanitabarrett: Using Voiceover, I've noticed the active text is read out differently to the actual links when it comes to separating out the dash before each list item. The actual links separate out the two: "emdash" "First". Whereas the active text reads like: 'comma Second'

http://government-frontend.dev.gov.uk/component-guide/contents-list/active%20content%20link/preview

app/views/components/_contents-list.html.erb
```diff
@@ -16,20 +16,47 @@
       <%= t("content_item.contents") %>
     </h2>
     <ol class="app-c-contents-list__list">
-      <% contents.each do |contents_item| %>
+      <% contents.each.with_index(1) do |contents_item, position| %>
+        <%
+          active_class = contents_item[:active]? "app-c-contents-list__list-item--active" : ""
+          aria_current = contents_item[:active]? "aria-current = true" : ""
```
vanitabarrett: We don't really need a value for aria_current is active is false, so could we rewrite this as:

` aria_current = "aria-current = true" if contents_item[:active]`

Same goes for active class

app/views/components/_contents-list.html.erb
maxgds: I've tested here and with a scratch html file locally but I can't get it to work either - I've double checked the implementation, but it looks like support is coming in slowly - according to the table on this page http://design-patterns.tink.uk/aria-current/ it should work in the latest MacOS so perhaps that's the problem.

app/views/components/docs/contents-list.yml
maxgds: Have replicated locally, but I'm reluctant to change anything as it would seem to be a case of trying to "fix" things specifically for the screenreader, which is generally advised against. In this case it would be awkward to add a span because of the use of link_to_if

vanitabarrett: @maxgds Having looked at the original PR for the contents list component, it looks like reading out the emdash perhaps wasn't ideal behaviour anyway, so makes sense to leave it as is for that reason too. 👍

app/views/components/_contents-list.html.erb
```diff
@@ -16,20 +16,49 @@
       <%= t("content_item.contents") %>
     </h2>
     <ol class="app-c-contents-list__list">
-      <% contents.each do |contents_item| %>
+      <% contents.each.with_index(1) do |contents_item, position| %>
+      <%
+          active_class = "app-c-contents-list__list-item--active" if contents_item[:active]
```
steventux: This and the next conditional line would arguably be better expressed in one conditional `if...end` block. My thinking is that it's more obvious if slightly longer.

app/views/components/_contents-list.html.erb
maxgds: Done.

maxgds:
Thanks all! @vanitabarrett I checked the heroku deploy on my home machine last night with the latests OS and was able to see voiceover recognise the "aria-current".

app/views/components/docs/contents-list.yml
```diff
@@ -15,6 +15,7 @@ accessibility_criteria: |
 
   - inform the user how many items are in the list
   - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
```
fofr: 👍 💯

PR #760, opened 2018-02-13
[DO NOT MERGE] Parts nav example

maxgds:
This branch is a *sample* implementation of using the contents list to provide chapter headings in government-frontend and should not be considered production-ready. The plan is to introduce this in an A/B test but this branch shows a straight replacement. 

Knowing that the A/B test would require the old CSS in place, and that the css clashes with the desired style, I have changed the class of the nav from part-navigation to parts-navigation. This then requires tests to be updated, so handling the tests in an a/b situation might mean a fair bit of repetition.

The change made here has affected both guides and travel advice, a workaround would be needed if this isn’t desirable. 

Known issues:
- travel advice is affected and probably shouldn't be
- this is currently a nav element wrapped inside a nav element which is far from ideal
- the styling needs tweaking as the component currently hangs slightly left of the body

---

Example pages:
https://government-frontend-pr-760.herokuapp.com/theory-test/what-to-take
https://government-frontend-pr-760.herokuapp.com/foreign-travel-advice/belize


sihugh:
This is great, thanks!  The AB test will be specifically limited to guides and will omit those within the education taxonomy (to limit our impact on other testing going on).

maxgds:
No longer needed.

PR #765, opened 2018-02-15
Take lists into account when checking content length

maxgds:
Our contents lists appear according to whether the length
of content warrants it. This fix addresses a problem whereby
sections with content in list format were excluding the list
from the calculation of length.

Before:
screen shot 2018-02-15 at 10 41 29
After:
screen shot 2018-02-15 at 10 41 10

---

Examples:
https://government-frontend-pr-765.herokuapp.com/government/collections/rare-and-imported-pathogens-laboratory-ripl
https://government-frontend-pr-765.herokuapp.com/government/collections/fraud-error-debt-and-grants-function

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -49,9 +49,10 @@ module ContentItem
     def first_item_content
       element = first_item
       first_item_text = ''
+      allowed_elements = %w(p ul ol)
 
       until element.name == 'h2'
-        first_item_text += element.text if element.name == 'p'
+        first_item_text += element.text if element.name.in?(allowed_elements)
         element = element.next_element
         break if element.nil?
       end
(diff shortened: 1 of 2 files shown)


app/presenters/content_item/contents_list.rb
```diff
@@ -51,7 +51,7 @@ def first_item_content
       first_item_text = ''
 
       until element.name == 'h2'
-        first_item_text += element.text if element.name == 'p'
+        first_item_text += element.text if element.name.in? %w(p ul ol)
```
steventux: There's a bit of a performance hit in constructing an array within a loop like this.
Ideally we'd use a constant or at least a variable initialised outside the loop.
Also it might be worth amending a test to cover content in a list.

maxgds:
@steventux Could you take another look?

test/presenters/content_item/contents_list_test.rb
```diff
@@ -92,6 +92,24 @@ def body
     assert @contents_list.show_contents_list?
   end
 
+  test "#show_contents_list? returns true if number of contents items is 2 and the first item's character count is above 415 including a list" do
```
steventux: :+1: Great to have proper coverage for this.

PR #769, opened 2018-02-16
Support aria-label in contents lists

maxgds:
This allows an aria-label to be added in order to contextualise
the contents list. It was requested by modelling services since
they will be using it for chapter headings on guides that were
previously contextualised for assistive technology.

---

Component guide for this PR:
https://government-frontend-pr-769.herokuapp.com/component-guide

### DIFF
--- a/app/views/components/_contents-list.html.erb
+++ b/app/views/components/_contents-list.html.erb
@@ -1,6 +1,7 @@
 <%
   format_numbers ||= false
   contents ||= []
+  aria_label ||= ''
   nested = !!contents.find { |c| c[:items] && c[:items].any? }
   parent_list_item_modifier = if nested
                                 'parent'
@@ -11,7 +12,13 @@
                               end
 %>
 <% if contents.any? %>
-  <nav role="navigation" class="app-c-contents-list">
+  <nav
+    role="navigation"
+    class="app-c-contents-list"
+    <% if aria_label.present? %>
+      aria-label="<%= aria_label %>"
+    <% end %>
+  >
     <h2>
       <%= t("content_item.contents") %>
     </h2>
--- a/app/views/components/docs/contents-list.yml
+++ b/app/views/components/docs/contents-list.yml
@@ -17,6 +17,9 @@ accessibility_criteria: |
   - convey the content structure
   - indicate the current page when contents span different pages, and not link to itself
 
+  The contents list may:
+  - include an aria-label to contextualise the list if helpful
+
   Links with formatted numbers must separate the number and text with a space for correct screen reader pronunciation. This changes pronunciation from "1 dot Item" to "1 Item".
 shared_accessibility_criteria:
   - link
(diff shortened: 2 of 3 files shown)


app/views/components/_contents-list.html.erb
```diff
@@ -11,7 +12,13 @@
                               end
 %>
 <% if contents.any? %>
-  <nav role="navigation" class="app-c-contents-list">
+  <nav
+    role="navigation"
+    class="app-c-contents-list"
+    <% unless aria_label.blank? %>
+      aria-label="<%= aria_label %>"
```
fofr: I wonder if the component should have a default, eg "Page contents list"

app/views/components/_contents-list.html.erb
```diff
@@ -11,7 +12,13 @@
                               end
 %>
 <% if contents.any? %>
-  <nav role="navigation" class="app-c-contents-list">
+  <nav
+    role="navigation"
+    class="app-c-contents-list"
+    <% unless aria_label.blank? %>
```
fofr: Can we use `if aria_label.present?`?

maxgds: Sure.

app/views/components/_contents-list.html.erb
maxgds: I decided against putting a default in since we've not been adding anything previously and it makes sense to be able to decide your context at the time of use. If somebody using it for cross-page contents lists forgets to supply a label then that particular default text seems misleading. Are you really keen for this to be added?

app/views/components/_contents-list.html.erb
```diff
@@ -11,7 +12,13 @@
                               end
 %>
 <% if contents.any? %>
-  <nav role="navigation" class="app-c-contents-list">
+  <nav
+    role="navigation"
+    class="app-c-contents-list"
+    <% if aria_label.present? ? %>
```
vanitabarrett: Stray `?`

andysellick: This could be written as 

aria_label ||= false

<%= "aria-label=#{aria_label}" if aria_label %>

maxgds: Fixed. (Copy paste error :(  )

app/views/components/_contents-list.html.erb
```diff
@@ -11,7 +12,13 @@
                               end
 %>
 <% if contents.any? %>
-  <nav role="navigation" class="app-c-contents-list">
+  <nav
+    role="navigation"
+    class="app-c-contents-list"
+    <% if aria_label.present? %>
```
andysellick: This could be written as

aria_label ||= false

<%= "aria-label=#{aria_label}" if aria_label %>

maxgds: I gave it a try but your version builds 
<... aria-label="firstword" every other word ...>
I tried quoting it to avoid but I couldn't get it working and it was starting to look messy.

PR #770, opened 2018-02-16
Revert dependabot prs

dwhenry:
Reverting the PR as we are now seeing a compile failure in production as I need to deploy this in order to remove a webchat instance.

This PR should be reapplied one at a time to determine which change is causing the issue

---

Component guide for this PR:


cbaines:
It seems unnecessary to me to revert the capybara and pry-byebug changes, unless you really think that these gems could be responsible for issues in production? They are in groups in the Gemfile, which as I understand it should mean that they are not even loaded by bundler.

dwhenry:
@cbaines I don't but the deploy is time critical so I have gone the nuclear approach to maximise my chances of success. As I said these should be reapplied later.

Issue #786, opened 2018-02-21
Sticky back to top/contents link causes footer to jump on transition


andysellick:
On pages such as [this one](https://www.gov.uk/drug-device-alerts/drug-alert-amoxicillin-sodium-500mg-powder-for-solution-for-injection-additional-reports-of-injection-site-reactions-in-paediatrics-and-adults?cachebust=1519124601) when the page is scrolled near the bottom and the 'Contents' link switches between sticky and non-sticky states, this causes the content beneath the footer to 'jump' upwards.

![contents-moving](https://user-images.githubusercontent.com/861310/36493263-1a612e9c-1727-11e8-8bd4-70cd74c55ddf.gif)

(thanks to @fofr for the video)

The cause of this could be that when the Contents link is changed to sticky, it leaves a gap in the page so the rest of the content has to move up to fill it. This was an issue in the prototype of a sticky header (not implemented), the solution was to wrap the element in a fixed height container that [wouldn't collapse](https://github.com/alphagov/govuk-services-prototype/pull/95/files#diff-3e8db389240e7dc0be0d61611d613510).

PR #787, opened 2018-02-21
Add a min-height to stop the back to top link collapsing

steventux:
Addresses https://github.com/alphagov/government-frontend/issues/786

A good test page is https://government-frontend-pr-787.herokuapp.com/drug-device-alerts/drug-alert-amoxicillin-sodium-500mg-powder-for-solution-for-injection-additional-reports-of-injection-site-reactions-in-paediatrics-and-adults?cachebust=1519124601

Here's a video of it misbehaving:

contents-moving

The elements below the back to top link can briefly
appear above the link as it transitions from an in-body
element to a fixed position, adding a minimum height stops the
element collapsing and causing this flickering element
reordering.

Thanks to @andysellick and @fofr for reporting and explaining this.

---

Component guide for this PR:
https://government-frontend-pr-787.herokuapp.com/component-guide

### DIFF
--- a/app/assets/stylesheets/components/_contents-list-with-body.scss
+++ b/app/assets/stylesheets/components/_contents-list-with-body.scss
@@ -16,6 +16,9 @@
   padding-bottom: $gutter-one-third;
 
   .app-c-back-to-top {
+    // This ensures the link doesn't 'collapse' allowing elements
+    // below the link to briefly appear above the link
+    min-height: 1px;
     padding-bottom: $gutter-one-third;
   }
 }


fofr:
Looks like this isn't currently working as expected.

steventux:
@fofr can you elaborate? I've tested this in several browsers now, the original commit wasn't successful as it was a bit simplistic but since https://github.com/alphagov/government-frontend/pull/787/commits/64b5e944d435c3f8a6f21b82e1d9390364a6f7ae I can't replicate the same jumpy behaviour. I've now tested on Safari 8, Chrome 64 and IE9. I've been able to replicate it (intermittently) on Firefox 58 if I use arrow keys but otherwise scrolling no longer produces a noticeable jump. Perhaps I'm used to working around the issue? Any info on how to replicate would be handy.

Would you say this latest commit is an improvement at all? If so I'd be inclined to accept this short term fix until we have time to dive a bit deeper into the problem.

vanitabarrett:
@steventux @fofr Possibly a caching issue - I thought I could replicate on https://government-frontend-pr-787.herokuapp.com/drug-device-alerts/drug-alert-amoxicillin-sodium-500mg-powder-for-solution-for-injection-additional-reports-of-injection-site-reactions-in-paediatrics-and-adults?cachebust=1519124601 on Chrome too, but then refreshed and the problem seems to be resolved with the latest commit.

steventux:
@vanitabarrett would you say this improves the behaviour? I think if we've improved the behaviour this is a harmless if slightly simplistic way of improving user experience.

vanitabarrett:
@steventux It definitely seems to improve the problem - I've just quickly checked on latest versions of Safari, Firefox and Chrome, and IE 9 on browserstack, and I can't replicate the issue any more

steventux:
@fofr @vanitabarrett this PR could do with approval if it's deemed to at least improve the behaviour.

vanitabarrett:
@steventux I can't seem to replicate this issue any more on the latest master build of government-frontend so I'm wondering if this fix is still needed? Testing in Chrome 64

steventux:
Very strange, I know some small stylistic tweaks were applied to the feedback component, perhaps the additional margin before the feedback component/after the content has negated the space where the two elements appeared to switch place. I'll close this then, thanks :+1:

PR #816, opened 2018-03-09
Add clarification message for tax credit webchat

dwhenry:
Message to try and avoid people incorrectly using the webchat when they want a
different department.

This only appears for the specific webchat instance:

screenshot 2018-03-09 14 11 27

https://trello.com/c/6Gt4ON2h/100-2-text-change-to-webchat

### DIFF
--- a/app/views/shared/_webchat.html.erb
+++ b/app/views/shared/_webchat.html.erb
@@ -1,6 +1,12 @@
 <span class="js-webchat"
   data-availability-url="<%= @content_item.webchat_availability_url %>"
   data-open-url="<%= @content_item.webchat_open_url %>">
+  <% if @content_item.base_path == '/government/organisations/hm-revenue-customs/contact/tax-credits-enquiries' %>
+    Advisers can only talk to you about Child Tax Credit and Working Tax Credit.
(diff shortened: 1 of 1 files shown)


app/views/shared/_webchat.html.erb
```diff
@@ -1,6 +1,11 @@
 <span class="js-webchat"
   data-availability-url="<%= @content_item.webchat_availability_url %>"
   data-open-url="<%= @content_item.webchat_open_url %>">
+  <% if @content_item.base_path == '/government/organisations/hm-revenue-customs/contact/tax-credits-enquiries' %>
+    Advisers can only talk to you about Child Tax Credit and Working Tax Credit.
+  <% end %>
+  <br/>
```
fofr: Should the line breaks be in the `if` statement?

dwhenry:
This is waiting on content agreement with the initial requester https://govuk.zendesk.com/agent/tickets/2581335

app/views/shared/_webchat.html.erb
suzannehamilton: Fixed. 👍

PR #823, opened 2018-03-13
Remove unused stubbing

tijmenb:
The "whitelisted taxons" thing has been removed in https://github.com/alphagov/govuk_navigation_helpers/pull/120.


tijmenb:
Turns out, this wasn't unused! This application hasn't yet been upgraded to the new navigation helpers gem which removes the feature.

PR #829, opened 2018-03-14
Stop using request variants to determine layout

tijmenb:
This removes the code for the taxonomy_navigation requests variants and uses `if` statements in the views instead.

This makes it clearer what's going on. It shouldn't make any change in the pages.

I suspect that most of these `if` statements aren't actually meaningful and can be removed. That's for a follow up PR.

https://trello.com/c/WJlqy3VD


app/views/shared/_related_items.html+taxonomy_navigation.erb
```diff
@@ -1,15 +0,0 @@
-<div class="column-third add-title-margin">
-  <% if @navigation.should_present_taxonomy_navigation? %>
```
tijmenb: The observant reader will notice that because this partial variant will only be shown when `should_present_taxonomy_navigation?` is true, this statement will also always be true. The `else` could never have been executed.

PR #837, opened 2018-03-19
Use contextual navigation components

tijmenb:
This PR uses the contextual navigation sidebar and breadcrumbs from the publishing components gem (https://github.com/alphagov/govuk_publishing_components/pull/228). It removes all the logic for what to display where and replaces it with calls to the components.

The intention is that this PR doesn't change the layout, but there are a few changes:

* [Contacts have the new sidebar](https://government-frontend-pr-837.herokuapp.com/government/organisations/hm-revenue-customs/contact/income-tax-enquiries-for-individuals-pensioners-and-employees), but the data has stayed the same. I'm unsure why universal layout hadn't been implemented for contacts in the first place, and Ganesh wasn't sure either. This might have been an oversight.
* ~The contacts breadcrumb has changed a little bit from "Contact HM Revenue & Customs" to "HM Revenue & Customs Contacts". This is because all contacts have the `parent` link to the contacts finder. I think this is a fine change. If we want to change the title we should change the finder title.~ The [finder has been renamed](https://www.gov.uk/government/organisations/hm-revenue-customs/contact) so there are no more changes now (https://github.com/alphagov/contacts-admin/pull/372).
* On topical event pages about pages, there's no more "(archived)" in the title when the topical event is archived. I think this is an acceptable trade-off given the complexity it would mean moving this special functionality to the components.  
* We no longer override the header navigation type for things tagged tagged to the taxonomy. I don't think there was very much reason behind showing it and not showing it, so I'm okay with this change (this is the `content_for :simple_header` stuff that's removed)
* Even though there's a `related_items` method in the travel advice presenter, it was unused and the travel advice pages are unchanged after this PR.

Can be reviewed now, but don't merge until it's using a released version of https://github.com/alphagov/govuk_publishing_components/pull/228.

https://trello.com/c/1LEofaHC


tijmenb:
@vanitabarrett thank you! updated.

PR #842, opened 2018-03-21
Check content item is a sign in page before attempting to sign in

thomasleese:
This works by checking that the path contains the string 'sign-in'. I was hoping to use a route constraint, but couldn't get it to work on a glob/catch all parameter.

---

Component guide for this PR:
https://government-frontend-pr-842.herokuapp.com/component-guide

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -23,6 +23,8 @@ class ContentItemsController < ApplicationController
   end
 
   def service_sign_in_options
+    return head :not_found unless is_sign_in_content_item_path?
+
     if params[:option].blank?
       @error = true
       show
@@ -34,6 +36,10 @@ class ContentItemsController < ApplicationController
   end
 
 private
+
+  def is_sign_in_content_item_path?
+    content_item_path.include?("sign-in")
+  end
 
   # Allow guides to pass access token to each part to allow
   # fact checking of all content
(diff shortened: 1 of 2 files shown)


h-lame:
I was curious about the constraint not working so did a bit of digging and found out that:

```ruby
post '*path' => 'content_items#service_sign_in_options',
      constraints: {
        path: /\/sign\-in\//
      }
```

didn't work (as @thomasleese suggested) which seemed weird to me.  Turns out it's because the regex is anchored to the start/end so turns into something like `/\A\/sign\-in\/\Z/` which would only match on an exact path like `/sign-in/` rather than `/personal-tax-account/sign-in/prove-identity` which is what we want to match.  If we use something like:

```ruby
post '*path' => 'content_items#service_sign_in_options',
      constraints: {
        path: /.+\/sign\-in\/.*/
      }
```

then it does work because this allows for stuff before `/sign-in/` in the path and optionally stuff after it (which may still be too permissive, not sure if we should match `/personal-tax-account/sign-in/` or just `/personal-tax-account/sign-in/prove-identity`.

I'm not 100% sure that a constraint in the routes is clearer than a check in the action though.  And, of course, the check in the action is merged so it wins by default 😉 I just wanted to satisfy my own curiosity really.

leenagupte:
@h-lame `/personal-tax-account/sign-in/` already redirects users to the first sign-in page, so `/personal-tax-account/sign-in/prove-identity`.

PR #853, opened 2018-03-28
Always use the non-numbered contents list

tijmenb:
This removes the `GuideChapterNav` A/B test. We've tested this and we'll keep the B version.

https://trello.com/c/ndUiAA1y


app/views/shared/_parts_navigation.html.erb
```diff
@@ -1,20 +1 @@
-<% if show_alternate_guide_chapters? %>
-  <%= render "components/contents-list", aria_label: navigation_label, contents: content_item.part_link_elements %>
-<% else %>
-  <nav role="navigation" class="grid-row part-navigation" aria-label="<%= navigation_label %>" data-module="track-click">
-    <% content_item.parts_navigation.each_with_index do |part_group, i| %>
-      <ol
-        class="column-half"
-        <% if i == 1 %>
-          start="<%= content_item.parts_navigation_second_list_start %>"
```
andysellick: This can also be removed!

app/views/shared/_parts_navigation.html.erb
```diff
@@ -1,20 +1 @@
-<% if show_alternate_guide_chapters? %>
-  <%= render "components/contents-list", aria_label: navigation_label, contents: content_item.part_link_elements %>
-<% else %>
-  <nav role="navigation" class="grid-row part-navigation" aria-label="<%= navigation_label %>" data-module="track-click">
```
andysellick: Are there any styles that can be removed?

tijmenb: Not anymore, since we're keeping the old style nav for travel advice.

PR #896, opened 2018-05-09
Implement schema placeholders

tijmenb:
This PR sets us up to be able to add more structured data to all the formats. It:

- Adds placeholders for all formats
- Adds a ArticleSchema supertype
- Eases the current restriction on when we add the schema to the page. We now also add it when there's no publishing organisation or image (which is "required" by Google, [but isn't really](https://stackoverflow.com/questions/41414163/google-requiring-certain-properties-for-certain-types-when-adding-schema-org-mar)).

https://trello.com/c/RLQI1IUV


app/presenters/schema_org/news_article_schema.rb
```diff
@@ -0,0 +1,66 @@
+module SchemaOrg
```
sihugh: Github doesn't seem to have recognised this as a move :-/

app/presenters/gone_presenter.rb
```diff
@@ -10,4 +10,9 @@ def initialize(content_item, requested_content_item_path = nil)
   def page_title
     "No longer available"
   end
+
+  # Gone pages do not need a schem because they're not useful to search engines
```
sihugh: `schem` --> `schema`

app/presenters/unpublishing_presenter.rb
```diff
@@ -10,4 +10,9 @@ def initialize(content_item, requested_content_item_path = nil)
   def page_title
     "No longer available"
   end
+
+  # Unpublishing pages do not need a schem because they're not useful to search engines
```
sihugh: `schem` --> `schema`

tijmenb:
Closing this, I'm going for a different approach!

PR #909, opened 2018-05-18
HUB-143: Small style change to description text on service sign in pages

rachelthecodesmith:
An option to display a text paragraph description under the heading on the service sign in pages has existed for some time but was not being used. We are about to start a base line test to see if having a description here improves through put by increasing understanding. However when testing what this would look like we found that the text was a little small and close to the error text (when shown). This commit contains small changes to change the text size and element spacing.

Note to reviewer: I am less familiar with the Gov.UK code and you seem to use different elements for styling than the ones we use on Verify so please let me know if this is not the right way to set the text size/if there is a better way of doing this that avoids adding a file. I am also not sure if I need to update the wraith tests or if they are just for running locally against the live pages. Please just comment and I can amend accordingly.

solo: @rachelthecodesmith

---

Visual regression results:

Component guide for this PR:

### DIFF
--- a/app/views/content_items/service_sign_in/_choose_sign_in.html.erb
+++ b/app/views/content_items/service_sign_in/_choose_sign_in.html.erb
@@ -21,7 +21,8 @@
   <%= render "govuk_publishing_components/components/fieldset", legend_text: legend_text do %>
     <div class="grid-row">
       <div class="column-two-thirds">
-        <%= render 'govuk_component/govspeak', content: @content_item.description %>
+        <div class="description"><%= render 'govuk_component/govspeak', content: @content_item.description %></div>
+        <br/>
         <% if @error %>
           <%= render "components/error-message", text: t('service_sign_in.error.option') %>
         <% end %>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -48,4 +48,5 @@
 @import 'views/specialist-document';
 @import 'views/answer';
 @import 'views/help-page';
-@import "views/guide";
+@import 'views/guide';
+@import 'views/service_sign_in';
--- a/app/assets/stylesheets/views/_service_sign_in.scss
+++ b/app/assets/stylesheets/views/_service_sign_in.scss
@@ -1,0 +1,5 @@
+.service-sign-in {
+  .description {
+    @include core-19;
+  }
+}


app/views/content_items/service_sign_in/_choose_sign_in.html.erb
```diff
@@ -21,7 +21,8 @@
   <%= render "govuk_publishing_components/components/fieldset", legend_text: legend_text do %>
     <div class="grid-row">
       <div class="column-two-thirds">
-        <%= render 'govuk_component/govspeak', content: @content_item.description %>
+        <p class="description"><%= render 'govuk_component/govspeak', content: @content_item.description %></p>
```
andysellick: Can you provide a link to an example page where this PR can be viewed? It should be somewhere on https://government-frontend-pr-909.herokuapp.com/ - the reason I ask is I think when you call the govspeak component it creates a div and then inserts the content you've specified inside it, which would mean you've got a paragraph containing a div here, which is invalid.

andysellick: I think [this an example page](https://government-frontend-pr-909.herokuapp.com/examples/service_sign_in/service_sign_in/choose-sign-in)? It looks like it hasn't put the govspeak div inside the p, it's put it afterwards, but that means you've got an empty p tag for no reason. You can probably just remove it?

rachelthecodesmith: Thanks, yes that's the one. The create account page is pretty much all govspeak but I thought from the schema the description was just a string, might have got that wrong though. My only aim with wrapping it was to adjust the font size and put a break in between the text and the error (if the error is showing) so I was hoping there might be a component I could utilise already, like the lead paragraph in govuk elements? Then I wouldn't have to add a style sheet. Thanks for your help with this.

rachelthecodesmith: Sorry, wont let me delete the above comment for some reason. Just looked at the heroku example and it's gov-speak, though it looks different when I run it locally probably just something I don't understand about how it processes govspeak. Basically we just want to make sure the text comes up at 19px and not too close the the error message. How it is on that heroku page is perfect so I'm not sure if this PR is required at all.

PR #920, opened 2018-06-01
Use the Title component from the gem

tijmenb:
https://trello.com/c/EruCrkqO/24-move-the-title-component-to-the-gem


tijmenb:
This keeps failing because it turns out we have a number of pages with duplicate `h1`s.

PR #928, opened 2018-06-07
Use title component from gem

tijmenb:
Supersedes https://github.com/alphagov/government-frontend/pull/920.

https://trello.com/c/EruCrkqO

### DIFF
--- a/test/integration/fatality_notice_test.rb
+++ b/test/integration/fatality_notice_test.rb
@@ -15,7 +15,7 @@ class FatalityNoticeTest < ActionDispatch::IntegrationTest
       visible: false
     )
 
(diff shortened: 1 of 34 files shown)


test/integration/fatality_notice_test.rb
```diff
@@ -15,7 +15,6 @@ class FatalityNoticeTest < ActionDispatch::IntegrationTest
       visible: false
     )
 
-    assert_component_parameter("title", "context", "Operations in Zululand")
```
andysellick: Should this deleted line be replaced with an `assert_has_component_title` test, as added in `choose_sign_in_test.rb`?

tijmenb: Updated with fixed test.

PR #944, opened 2018-06-19
HUB-167: Recognise and report Verify users on GOV.UK

rachelthecodesmith:
https://govukverify.atlassian.net/browse/HUB-167 : Test - verify user behaviour on interstitial page GOV.UK

We'd like to see what choices existing Verify users make on the
service sign-in page. Therefore we're loading a JS snippet from Verify to make
an ajax call and based on the user cookie to decide whether it's a
Verify user and report accordingly. This is a baseline test to learn about the users' behaviour.

The js that is referenced is in verify-frontend -> https://github.com/alphagov/verify-frontend/blob/master/public/get_hint.js

It receives a boolean answer to if the user has the journey hint cookie present and triggers a Google Analytics event with the presence or absence of the hint together with the option the user selected.

The cookie that determines if a user has visited verify before is issued from the verify domain and is covered under the verify cookie policy.

Co-Authored-by: Jakub Miarka <jakub.miarka@digital.cabinet-office.gov.uk>

---

Visual regression results:
https://government-frontend-pr-944.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-944.herokuapp.com/component-guide

### DIFF
--- a/app/assets/javascripts/modules/track-radio-group.js
+++ b/app/assets/javascripts/modules/track-radio-group.js
@@ -8,21 +8,45 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GOVUK.Modules.TrackRadioGroup = function () {
     this.start = function (element) {
-      element.on('submit', function (event) {
+      track(element)
+
+      checkVerifyUser(element)
+    }
+
+    function track (element, withHint) {
+        element.on('submit', function (event) {
+        
         var options = { transport: 'beacon' }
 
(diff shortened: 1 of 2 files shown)


sihugh:
Do we need to have this cookie listed on https://www.gov.uk/help/cookies?

rachelthecodesmith:
Changes made as per email exchange. I can add any documentation you feel is needed if you let me know where best to put it.

PR #969, opened 2018-07-17
Add initial content pages nav AB testing code

sihugh:
At present we don't want to show any difference.  This commit is to add the
AB testing code so we can hook into it wih subsequent work.  We will be showing
topic taxonomy based navigation on variant B.

https://trello.com/c/sbDytyen/75-set-up-ab-test-in-government-frontend

https://trello.com/c/wIy1dXTj/82-epic-get-data-for-taxonomy-navigation

### DIFF
--- a/app/controllers/concerns/content_pages_nav_ab_testable.rb
+++ b/app/controllers/concerns/content_pages_nav_ab_testable.rb
@@ -1,0 +1,38 @@
+module ContentPagesNavAbTestable
+  CUSTOM_DIMENSION = 65
+
+  def self.included(base)
+    base.helper_method(
+      :content_pages_nav_test_variant,
+      :show_new_navigation?,
+      :page_in_scope?
+    )
+    base.after_action :set_test_response_header
+  end
+
+  def content_pages_nav_test
+    @content_pages_nav_test ||= GovukAbTesting::AbTest.new(
+      "ContentPagesNav",
+      dimension: CUSTOM_DIMENSION,
+      allowed_variants: %w(A B),
+      control_variant: "A"
+    )
+  end
+
+  def content_pages_nav_test_variant
+    @content_pages_nav_test_variant ||= content_pages_nav_test.requested_variant(request.headers)
+  end
+
+  def set_test_response_header
+    content_pages_nav_test_variant.configure_response(response) if page_in_scope?
+  end
+
+  def show_new_navigation?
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/content_pages_nav_ab_testable.rb
```diff
@@ -0,0 +1,38 @@
...
+  def content_pages_nav_test_variant
+    @content_pages_nav_test_variant ||= content_pages_nav_test.requested_variant(request.headers)
+  end
+
+  def set_test_response_header
+    content_pages_nav_test_variant.configure_response(response) if page_in_scope?
+  end
+
+  def show_new_navigation?
+    !content_pages_nav_test_variant.variant?("B") && page_in_scope?
```
andrewgarner: The control variant is "A", so that should remain unchanged? Therefore should this be:

```rb
content_pages_nav_test_variant.variant?("B") && page_in_scope?
```

app/controllers/content_items_controller.rb
```diff
@@ -14,6 +15,8 @@ class ContentItemsController < ApplicationController
   def show
     load_content_item
 
+    load_taxonomy_navigation if page_in_scope?
```
andrewgarner: Should we use the `show_new_navigation?` helper here?

app/views/layouts/application.html.erb
```diff
@@ -24,6 +24,7 @@
   <% if @content_item.description %>
     <meta name="description" content="<%= strip_tags(@content_item.description) %>" />
   <% end %>
+  <%= content_pages_nav_test_variant.analytics_meta_tag.html_safe if page_in_scope? %>
```
andrewgarner: And here?

app/controllers/content_items_controller.rb
sihugh: Yep, that's true

app/views/layouts/application.html.erb
sihugh: We need the analytics tag if the page is in scope even if we're not loading the navigation.  It lets Haur compare A vs B more easily.

app/controllers/concerns/content_pages_nav_ab_testable.rb
sihugh: That's right.  We'll switch this

sihugh:
Closing this as it's been superseded by other work

PR #994, opened 2018-07-25
Add supergroup sections to taxonomy navigation

oscarwyatt:
Adds guidance and regulation, and policy and engagement sections to the nav section at the footer of content pages.

- Policy and engagement content:
  - configured to get latest content
  - shows public updated at date
  - document type
- Guidance and regulation
  - retrieved by popularity (as determined by rummager)
  - shows document type (but not date)

They're both displayed as lists of items rather than highlight boxes.

The content page will show this navigation if they are tagged to a live taxonomy, be rendered by this application and not be an HTML publication (and have results to show)

This PR is to get things generally close to the desired outcome and then we'll polish after testing and a further design review.

screen shot 2018-07-27 at 11 43 18

## Examples:
You will need to have an extension like [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en) installed in your browser to see the B Variant. The values should be:
Request Header Name: GOVUK-ABTest-ContentPagesNav
Request Header Value: B

- [Tagged to one taxon](https://government-frontend-pr-994.herokuapp.com/apply-apprenticeship)
- [Tagged to multiple taxons](https://government-frontend-pr-994.herokuapp.com/government/publications/esfa-update-18-july-2018)
- [Tagged to a taxon without services](https://government-frontend-pr-994.herokuapp.com/government/publications/working-together-to-safeguard-children--2)
- [Not tagged to any taxons](https://government-frontend-pr-994.herokuapp.com/government/case-studies/hard-lines)
- [HTML Publication](https://government-frontend-pr-994.herokuapp.com/government/publications/budget-2016-documents/budget-2016)

Component guide for this PR:
https://government-frontend-pr-994.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -1,0 +1,25 @@
+module Supergroups
+  class Supergroup
+  private
+
+    def format_document_data(documents, include_timestamp: true)
+      documents&.map do |document|
+        data = {
+          link: {
+            text: document["title"],
+            path: document["link"]
+          },
+          metadata: {
+            document_type: document["content_store_document_type"].humanize
+          }
+        }
+
+        if include_timestamp && document["public_timestamp"]
+          data[:metadata][:public_updated_at] = Date.parse(document["public_timestamp"])
+        end
+
+        data
+      end
+    end
+  end
+end
--- a/app/presenters/supergroups/supergroup.rb
+++ b/app/presenters/supergroups/supergroup.rb
@@ -2,12 +2,13 @@ module Supergroups
   class Supergroup
   private
 
-    def format_document_data(documents, include_timestamp: true)
-      documents&.map do |document|
+    def format_document_data(documents, data_category: nil, include_timestamp: true)
+      documents.each.with_index(1).map do |document, index|
         data = {
           link: {
             text: document["title"],
-            path: document["link"]
+            path: document["link"],
+            data_attributes: data_attributes(document["link"], document["title"], index)
           },
           metadata: {
             document_type: document["content_store_document_type"].humanize
@@ -18,8 +19,27 @@ module Supergroups
           data[:metadata][:public_updated_at] = Date.parse(document["public_timestamp"])
         end
 
+        if data_category.present?
+          data[:link][:data_attributes][:track_category] = data_module_label + data_category
+        end
+
         data
       end
+    end
+
+    def data_attributes(base_path, link_text, index)
+      {
+        track_category: data_module_label + "DocumentListClicked",
+        track_action: index,
+        track_label: base_path,
+        track_options: {
+          dimension29: link_text
+        }
+      }
+    end
+
+    def data_module_label
+      self.class.name.demodulize.camelize(:lower)
     end
   end
 end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -69,9 +69,13 @@ private
 
       taxon_ids = taxons.map { |taxon| taxon["content_id"] }
       services = Supergroups::Services.new(content_item_path, taxon_ids)
+      guidance_and_regulation = Supergroups::GuidanceAndRegulation.new(content_item_path, taxon_ids)
+      policy_and_engagement = Supergroups::PolicyAndEngagement.new(content_item_path, taxon_ids)
 
       @taxonomy_navigation = {
-        services: (services.all_services if services.any_services?),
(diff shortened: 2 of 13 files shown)


app/lib/services.rb
```diff
@@ -10,4 +11,8 @@ def self.content_store
       disable_cache: true,
     )
   end
+
+  def self.rummager
```
vanitabarrett: I think once you rebase you won't need this as it's already been added for the Services work

app/presenters/content_item_presenter.rb
```diff
@@ -71,6 +71,24 @@ def cache_control_public?
     !content_item.cache_control.private?
   end
 
+  def has_policy_and_engagement_links?
```
vanitabarrett: I think these should probably be moved out into `policy_and_engagement.rb` as it's specific to that. Same with `policy_and_engagement_links_content`

config/application.rb
```diff
@@ -22,7 +22,10 @@ class Application < Rails::Application
     config.time_zone = 'London'
 
     # Custom directories with classes and modules you want to be autoloadable.
-    config.autoload_paths += %W(#{config.root}/lib)
+    config.autoload_paths += %W(
+                                #{config.root}/lib
+                                #{config.root}/app/services/supergroups
```
vanitabarrett: This is great because it follows the pattern in Collections, but our previous work on Services and News (I think) puts everything in lib for now. It may make sense to follow that pattern for now and perhaps create a card to move everything into `/services` later.

app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -24,3 +24,8 @@
 
   <%= render 'shared/sidebar_navigation' %>
 </div>
+
+<% if @content_item.has_policy_and_engagement_links? %>
+  <%= render "govuk_publishing_components/components/highlight_boxes", {
+      items: @content_item.policy_and_engagement_links_content } %>
+<% end %>
```
vanitabarrett: Missing newline at end of file 🙂 You can probably configure your editor to do this automatically.

app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -24,3 +24,8 @@
 
   <%= render 'shared/sidebar_navigation' %>
 </div>
+
+<% if @content_item.has_policy_and_engagement_links? %>
```
vanitabarrett: This logic looks good, but it needs to be added to `shared/_taxonomy_navigation.html.erb`

sihugh:
@vanitabarrett I've addressed a few things and added tracking.  Do you think you could re-review if you have some time please?

test/integration/answer_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class AnswerTest < ActionDispatch::IntegrationTest
+  include RummagerHelpers
```
vanitabarrett: Is this needed here? Doesn't look like we're using it?

app/presenters/supergroups/supergroup.rb
```diff
@@ -0,0 +1,45 @@
...
+          data[:link][:data_attributes][:track_category] = data_module_label + data_category
+        end
+
+        data
+      end
+    end
+
+    def data_attributes(base_path, link_text, index)
+      {
+        track_category: data_module_label + "DocumentListClicked",
```
vanitabarrett: Looks like some of the tracking stuff won't work because the document list component doesn't have `data-module="track-click"`, it requires you to wrap the document-list in an element with that attribute. Not sure how I didn't spot that with the services work, sorry! That'll need changing there too.

In the long term we probably just want to update the component to do this by default if someone passes tracking info, but it will require updating the other frontend apps.

test/integration/answer_test.rb
sihugh: I'll check that, thanks

app/presenters/supergroups/supergroup.rb
sihugh: Boo! 👎

test/integration/answer_test.rb
sihugh: Removed, thanks

app/presenters/supergroups/supergroup.rb
sihugh: I've wrapped the partial in the data-module now and it seems to work OK.

PR #996, opened 2018-07-26
Use `first_published_at`

tijmenb:
`details.first_public_at` is deprecated. They should be same, so no frontend changes are intended.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/801.

https://trello.com/c/bMv5Tkgp

### DIFF
--- a/app/presenters/content_item/updatable.rb
+++ b/app/presenters/content_item/updatable.rb
@@ -14,7 +14,7 @@ module ContentItem
     end
 
     def first_public_at
-      content_item["details"]["first_public_at"]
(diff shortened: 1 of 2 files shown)


app/presenters/content_item/updatable.rb
```diff
@@ -14,7 +14,7 @@ def history
     end
 
     def first_public_at
-      content_item["details"]["first_public_at"]
```
kevindew: I think for safety on this you might have to use `content_item["details"]["first_public_at"] || content["first_published_at"]`

tijmenb: Updated

PR #995, opened 2018-07-26
Add News & Communications and Transparency content

MahmudH:
- Retrieves most recent content for News and Communications supergroup
- Retrieves most recent content for Transparency supergroup content

PR #997, opened 2018-07-26
Exclude current page from most popular searches

sihugh:
We don't want to show the page that a user is on in any navigation.  It's
confusing and confers no benefit.

An example page with itself in the results is: https://www.gov.uk/government/publications/retiring-teachers-nomination-form-for-official-letter-of-thanks

https://trello.com/c/ljVuTZUL/80-exclude-current-page-from-navigation

## Examples:
You will need to have an extension like [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en) installed in your browser to see the B Variant. The values should be:
Request Header Name: GOVUK-ABTest-ContentPagesNav
Request Header Value: B

- [Tagged to one taxon](https://government-frontend-pr-997.herokuapp.com/apply-apprenticeship)
- [In its own results set](https://government-frontend-pr-997.herokuapp.com/government/publications/retiring-teachers-nomination-form-for-official-letter-of-thanks)

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -68,7 +68,7 @@ private
       taxons = @content_item.taxons.select { |taxon| taxon["phase"] == "live" }
 
       taxon_ids = taxons.map { |taxon| taxon["content_id"] }
-      services = Supergroups::Services.new(taxon_ids)
+      services = Supergroups::Services.new(content_item_path, taxon_ids)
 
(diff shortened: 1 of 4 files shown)


vanitabarrett:
We'll need to update this in https://github.com/alphagov/government-frontend/pull/995 and https://github.com/alphagov/government-frontend/pull/994 too

PR #1008, opened 2018-07-31
Taxonomy navigation collections

andrewgarner:
Moves related collections from the sidebar to the taxonomy navigation.

### Before:

screen shot 2018-07-31 at 14 48 38

### After:

screen shot 2018-07-31 at 14 49 01

https://trello.com/c/zUWqJgX9

---

Visual regression results:
https://government-frontend-pr-1008.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1008.herokuapp.com/component-guide

### DIFF
--- a/app/assets/stylesheets/helpers/_taxonomy-navigation.scss
+++ b/app/assets/stylesheets/helpers/_taxonomy-navigation.scss
@@ -1,11 +1,14 @@
+// govuk_frontend_toolkit imports
+@import 'css3';
+
 .taxonomy-navigation {
   border-top: 2px solid $govuk-blue;
-  padding-top: $gutter-half;
 }
 
-.taxonomy-navigation__title-link {
-  @include bold-24;
-  display: block;
+.taxonomy-navigation__list-item {
+  margin-bottom: $gutter / 6;
+  padding-bottom: $gutter / 6;
+  list-style: none;
 }
 
 .taxonomy-navigation__section {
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/helpers/_taxonomy-navigation.scss
```diff
@@ -14,6 +17,32 @@
   border-top: 1px solid $black;
 }
 
-.taxonomy-navigation__section:first-of-type {
-  margin-top: $gutter * 1.5;
+.taxonomy-navigation__row {
+  display: flex;
```
vanitabarrett: Is there a reason why this needs to use flex rather than just the grid classes (`grid-row` and `column-one-third`)?

andrewgarner: So that they line up with the highlight boxes below.

vanitabarrett: I think it can still line up without flex? Quick testing in the herokuapp looks ok:

screen shot 2018-07-31 at 15 06 37

andrewgarner: The breakpoints are different:

Grid:

ezgif com-video-to-gif
Flex:

ezgif com-video-to-gif-2

vanitabarrett: Ah I see, I didn't spot that - that's annoying! Thanks for explaining. 

I wonder if we need it to match up or if we could just have each column half the width of that row. But using flex isn't a big deal, it just adds a lot of CSS.

andrewgarner: It was something @miaallers mentioned when we went through it together. I should imagine we can find a better way to implement if the AB test is positive.

PR #1035, opened 2018-08-08
Clean up the world location base path class

cbaines:
First, revert a ineffective change I made, then attempt to clean this WorldLocationBasePath class up.

As far as I can work out, this is never used. The only document type
that uses the metadata, which is I think where this is possibly used
is travel advice. I checked if any travel advice content is tagged to
world locations, and it isn't.

This is very confusing, as the exceptional cases aren't
exceptional. Looking at the code, the behaviour looks to be the same
with or without them.

I've also removed the comment, as the situation with the migration of
"world locations to a taxonomy based model" is pretty uncertian, as
the Topic Taxonomy doesn't have anything to do with country
classification (it's about subject, not geography), and this
"temporary solution" both as far as I can work out, doesn't do
anything, and has been here for over a year.

### DIFF
--- a/app/services/world_location_base_path.rb
+++ b/app/services/world_location_base_path.rb
@@ -22,7 +22,7 @@
 class WorldLocationBasePath
   EXCEPTIONAL_CASES = {
     "Democratic Republic of Congo" => "democratic-republic-of-congo",
-    "South Georgia and the South Sandwich Islands" => "south-georgia-and-south-sandwich-islands",
+    "South Georgia and the South Sandwich Islands" => "south-georgia-and-the-south-sandwich-islands",
     "St Pierre & Miquelon" => "st-pierre-miquelon"
   }.freeze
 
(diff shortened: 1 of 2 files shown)


cbaines:
Thanks for taking a look Paul :)

benthorner:
Looks like this is still necessary e.g. for [St Pierre & Miquelon](https://www.gov.uk/government/news/your-passport-application-did-you-send-it-to-the-right-place) (link is at the bottom of the page).

cbaines:
> Looks like this is still necessary e.g. for [St Pierre & Miquelon](https://www.gov.uk/government/news/your-passport-application-did-you-send-it-to-the-right-place) (link is at the bottom of the page).

I don't believe the changes here are directly involved, since the special case removed here doesn't help, and just leads to a non-existent page:
```
"St Pierre & Miquelon" => "st-pierre-miquelon"
```

Using the slug `st-pierre-miquelon` as given above doesn't work currently: https://www.gov.uk/world/st-pierre-miquelon/news

benthorner:
Sure, the actual page is https://www.gov.uk/world/st-pierre-and-miquelon/news. I should clarify my intent here: this was mainly to document the continued need for the original code, which provided for special cases like this one. However, I don't think it's worth investing effort in fixing it right now, as that's a distracting from the underlying problem - no base paths 😨.

PR #1039, opened 2018-08-10
Use hide_chapter_navigation on guides

andysellick:
- only happens if hide_chapter_navigation is true and the guide is in a step by step
- promotes guide part subtitle to be the page title
- hides guide contents list
- hides next/prev navigation and guide print link

**Before:**

screen shot 2018-07-31 at 14 10 07

**After:**

screen shot 2018-07-31 at 14 10 22

Follows on from [this spike](https://github.com/alphagov/government-frontend/pull/1007) done previously.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/809

---

Trello card: https://trello.com/c/RNsA7E4Q/763-hide-chapter-navigation-on-mainstream-guides

Visual regression results:
https://government-frontend-pr-1039.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1039.herokuapp.com/component-guide

### DIFF
--- a/app/presenters/guide_presenter.rb
+++ b/app/presenters/guide_presenter.rb
@@ -41,9 +41,30 @@ class GuidePresenter < ContentItemPresenter
     end
   end
 
+  def show_guide_navigation?
+    multi_page_guide? && !hide_chapter_navigation?
+  end
+
+  def content_title
+    if parts.any?
+      return current_part_title if hide_chapter_navigation?
+    end
+
+    title
+  end
+
 private
 
   def draft_access_token_param
     "token=#{draft_access_token}" if draft_access_token
+  end
+
+  def hide_chapter_navigation?
+    hide_navigation_elements = content_item.parsed_content["details"]["hide_chapter_navigation"]
+    part_of_step_navs? && hide_navigation_elements
+  end
+
+  def part_of_step_navs?
+    content_item.parsed_content["links"]["part_of_step_navs"].present?
   end
 end
--- a/app/views/content_items/guide.html.erb
+++ b/app/views/content_items/guide.html.erb
@@ -9,30 +9,34 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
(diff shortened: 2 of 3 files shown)


app/presenters/guide_presenter.rb
```diff
@@ -41,6 +41,28 @@ def parts
...
   end
 
+  def hide_chapter_navigation?
+    hide_navigation_elements = content_item.parsed_content["details"]["hide_chapter_navigation"]
+    step_navs = content_item.parsed_content["links"]["part_of_step_navs"]
+
+    part_of_step_navs = false
+    part_of_step_navs = step_navs.length if step_navs
+
+    true if part_of_step_navs && hide_navigation_elements
```
leenagupte: You don't need to do `true if`. You can just do:

`part_of_step_navs && hide_navigation_elements`

Ruby will return the result of the last statement regardless.

The same is true for `show_guide_navigation?`.

app/presenters/guide_presenter.rb
```diff
@@ -41,6 +41,28 @@ def parts
     end
   end
 
+  def hide_chapter_navigation?
+    hide_navigation_elements = content_item.parsed_content["details"]["hide_chapter_navigation"]
+    step_navs = content_item.parsed_content["links"]["part_of_step_navs"]
+
+    part_of_step_navs = false
+    part_of_step_navs = step_navs.length if step_navs
```
leenagupte: I'd extract these three lines out to a method, `part_of_step_navs?`

I think it could also be reduced to one line checking whether `content_item.parsed_content["links"]["part_of_step_navs"]` exists and the size is greater than 1.

app/presenters/guide_presenter.rb
```diff
@@ -41,6 +41,28 @@ def parts
     end
   end
 
+  def hide_chapter_navigation?
```
leenagupte: You could make this a private method as it's not being used by anything else.

PR #1032, opened 2018-08-07
Add content item links rules subgroups

oscarwyatt:
Extend content item navigation links to show content filtered by subgroups. Supersedes PR #1028 (add-content-item-links-rules)

screenshot-localhost-3090-2018 08 07-16-16-49

Visual regression results:
https://government-frontend-pr-1032.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1032.herokuapp.com/component-guide

Trello card
https://trello.com/c/A55u68Xw/98-implement-ruleset-for-showing-navigation-in-government-frontend

PR #1042, opened 2018-08-10
Use rummager results for image_urls

sihugh:
This eliminates up to three calls to content store for each page's navigation.

To do this we need to add image_url to the fields that rummager returns.

We can remove the alt text for images that are links because the link text is more relevant (and is preferred by browsers anyway).

We then need a default image in case the content doesn't have an image (not all do).

After rebasing on top of https://github.com/alphagov/government-frontend/pull/1043 and https://github.com/alphagov/government-frontend/pull/1032 it became clear that we can remove some things from the `news_and_communications` and `services` as they don't need to have content split into `promoted` and `tagged` content any more as they only display it in one form now.

I've therefore removed the `promoted` sections from these presenters which allows us to simplify some things and delete a bunch of code.

---

https://government-frontend-pr-1042.herokuapp.com/apply-apprenticeship

### DIFF
--- a/app/presenters/supergroups/news_and_communications.rb
+++ b/app/presenters/supergroups/news_and_communications.rb
@@ -2,6 +2,8 @@ module Supergroups
   class NewsAndCommunications < Supergroup
     attr_reader :content
 
+    PLACEHOLDER_IMAGE = "https://assets.publishing.service.gov.uk/government/assets/placeholder.jpg".freeze
+
     def initialize(current_path, taxon_ids, filters)
       super(current_path, taxon_ids, filters, MostRecentContent)
     end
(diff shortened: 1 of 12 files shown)


app/presenters/supergroups/news_and_communications.rb
```diff
@@ -26,22 +28,33 @@ def tagged_content
...
+  private
 
-      content.each do |document|
-        document_image = news_item_photo(document[:link][:path])
-        document[:image] = {
-          url: document_image["url"],
-          alt: document_image["alt_text"],
-          context: document_image["context"]
+    def format_document_data(documents, data_category: nil)
+      documents.each.with_index(1).map do |document, index|
```
andrewgarner: I had to double check the Ruby API documentation to confirm that `1` is an offset. Could we extract this to provide a little more context?

sihugh: 👍

app/presenters/supergroups/news_and_communications.rb
```diff
@@ -2,6 +2,8 @@ module Supergroups
   class NewsAndCommunications < Supergroup
     attr_reader :content
 
+    PLACEHOLDER_IMAGE = "https://assets.publishing.service.gov.uk/government/assets/placeholder.jpg".freeze
```
andrewgarner: Should we use Plek to make this environment specific?

```rb
PLACEHOLDER_IMAGE = URI::HTTPS.build(host: Plek.current.public_asset_host, path: '/government/assets/placeholder.jpg').freeze
```

sihugh:
I've updated this and rebased after a couple of merges.  It's worth sticking `?w=1` on the end of the URL when looking at the diff as Github's got a little confused about what is a change in a couple of places.

PR #1057, opened 2018-08-16
Fix intermittent failing tests

oscarwyatt:
Improves further on PR #1052. Moved custom configuration to a before_initialize block which seems to solve a problem with the config not being available in test environment. Based on the following stackoverflow post

https://stackoverflow.com/a/44770663

screenshot-stackoverflow com-2018 08 16-14-06-26

Trello: https://trello.com/c/toblYkmD/133-investigate-and-fix-failing-tests-on-government-frontend

---

Visual regression results:
https://government-frontend-pr-1057.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1057.herokuapp.com/component-guide

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,8 +50,7 @@ module GovernmentFrontend
     # Do not swallow errors in after_commit/after_rollback callbacks.
     # config.active_record.raise_in_transactional_callbacks = true
 
-    config.before_initialize do
-      config.taxonomy_navigation_links_out = config_for(:taxonomy_navigation_links_out)
-    end
+    # Load links out configuration for taxonomy navigation
+    config.taxonomy_navigation_links_out = config_for(:taxonomy_navigation_links_out)
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,6 +50,8 @@ module GovernmentFrontend
     # Do not swallow errors in after_commit/after_rollback callbacks.
     # config.active_record.raise_in_transactional_callbacks = true
 
-    config.taxonomy_navigation_links_out = config_for(:taxonomy_navigation_links_out)
+    config.before_initialize do
+      config.taxonomy_navigation_links_out = config_for(:taxonomy_navigation_links_out)
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@oscarwyatt Not sure this is resolved - still getting the same errors, e.g:

```
[build] Failure:
[build] LinksOutConfigTest#test_links_out_configuration_causes_no_errors_and_correct_supergroups_are_displayed_for_each_ruleset [/var/lib/jenkins/workspace/intermittently-failing-test-IAOZESY5I7JDI3LIV3SCFYLXHQ4CKM4DHPPJL6OA4VBZIDUKB4CA/test/integration/links_out_config_test.rb:60]:
[build] Expected false to be truthy.
```

and

```
[build] Error:
[build] LinksOutConfigTest#test_links_out_configuration_causes_no_errors_and_correct_supergroups_are_displayed_for_each_ruleset:
[build] NoMethodError: undefined method `each_key' for nil:NilClass
[build]     test/integration/links_out_config_test.rb:56:in `block (3 levels) in <class:LinksOutConfigTest>'
[build]     test/integration/links_out_config_test.rb:55:in `each_key'
[build]     test/integration/links_out_config_test.rb:55:in `block (2 levels) in <class:LinksOutConfigTest>'
[build]     test/integration/links_out_config_test.rb:54:in `block in <class:LinksOutConfigTest>'
[build] 
[build] bin/rails test test/integration/links_out_config_test.rb:51
```

I've been taking a look this afternoon too, but haven't managed to make any progress on it. I think if we still don't have a solution tomorrow we should probably remove this test for now so we're not slowing other people down in government-frontend (as it's updated quite often). We can add it back in when we're confident it's working 🙂

andrewgarner:
@vanitabarrett @oscarwyatt I've made some changes that I hope will help this:

1. I've made it so we actually stub the Rails configuration object not modify it under test. This is to ensure we don't accidentally break the application which I could happen with randomly ordered tests.

2. I've rewritten the config test so that the tests are dynamically generated based on the contents of the config file. This means that each test only has to make one browser request and a couple of assertions. Hopefully that makes timeouts less likely. Hopefully it also a bit easier to comprehend.

PR #1066, opened 2018-08-29
Remove reject_link filter and fix tests

oscarwyatt:
Refactors most_recent_content and most_popular_content to remove filter_link in Rummager queries and instead filter out any content item with the current path afterwards. This means Rummager is more able to cache results with the same taxonomy for different pages. Benchmarking with 50 random paths showed an improvement in average query time from 0.1276s to 0.0015s.

Following a review request, if there are not enough results after filtering out any results with the same link, it will make a second rummager query which includes the reject_link parameter to ensure it gets enough results. As this will make the overall page load time longer, it is monitored so if we're getting a high number of requests where this is happening we'll be able to make the initial rummager query request more than one extra link thereby mitigating the problem

Additionally refactored some tests to make them slightly more consistent

---

Visual regression results:
https://government-frontend-pr-1066.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1066.herokuapp.com/component-guide

Trello card:
https://trello.com/c/xh5YnCze/136-spike-improve-navigation-performance

### DIFF
--- a/app/services/most_popular_content.rb
+++ b/app/services/most_popular_content.rb
@@ -1,5 +1,3 @@
-require 'gds_api/rummager'
-
 class MostPopularContent
   include RummagerFields
 
@@ -17,7 +15,7 @@ class MostPopularContent
   end
 
   def fetch
-    search_response["results"]
+    search_response
   rescue GdsApi::HTTPErrorResponse => e
     GovukStatsd.increment("govuk_content_pages.most_popular.#{e.class.name.demodulize.downcase}")
     []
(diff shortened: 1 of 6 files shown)


app/services/most_popular_content.rb
```diff
@@ -27,16 +25,15 @@ def fetch
...
   def search_response
     params = {
-      start: 0,
-      count: number_of_links,
-      fields: RummagerFields::TAXON_SEARCH_FIELDS,
-      filter_part_of_taxonomy_tree: content_ids,
-      order: '-popularity',
-      reject_link: current_path,
+        start: 0,
+        count: number_of_links + 1,
```
vanitabarrett: Does `+1` assume that you'd only ever receive one link back that matches the current path? I don't know if this works with multipage formats like guides (see PR fix we did previously here: https://github.com/alphagov/government-frontend/pull/1026)

oscarwyatt: Ah, right, yes, thanks for letting me know. can't think of a way around this without either making another rummager call with the reject added back in to make up the numbers, or just requesting >1 extra links (in which case, how can I figure out how many is a good number?) - do you know any promising directions to go in?

sihugh: For multipage formats, so long as they're only one content item (which I think is true), rummager will only know about them once.  So I think we're fine with the `+ 1` approach.

sihugh:
Closing this as we're removing the AB test code. Nice though, and I would have liked to run some metrics against this :-)

sihugh:
I reopened this as a way to revisit the content pages nav prototype.  I've finished with this now

PR #1100, opened 2018-09-26
Link to people instead of ministers

kevindew:
Trello: https://trello.com/c/3GT17tkH/211-associate-content-with-ministers-l

This is marked as not to be merged due to requiring this: https://github.com/alphagov/govuk-content-schemas/pull/816

The ministers link type is now [deprecated](https://github.com/alphagov/govuk-content-schemas/blob/76fe9441a2b9ec6f030735f012347dbe8eb6e331/formats/news_article.jsonnet#L47) as it serves the same purpose as people links and these contain the same data. This allows widening up to link to people with other role appointments.

As people document_types don't require a base_path it also does a check that those actually have links (this is something that can affect pretty much all links so there should probably be more checks across this app - but I'm not opening up that rabbit hole just yet)

### DIFF
--- a/app/presenters/content_item/linkable.rb
+++ b/app/presenters/content_item/linkable.rb
@@ -3,7 +3,7 @@ module ContentItem
     include ActionView::Helpers::UrlHelper
 
     def from
-      organisations_ordered_by_importance + links_group(%w{worldwide_organisations ministers speaker})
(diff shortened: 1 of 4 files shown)


kevindew:
@benthorner Thanks, I went for a select rather than a reject as that felt more natural, unfortunately there's an annoying extra condition for world_locations

PR #1110, opened 2018-10-01
Fix toggle behaviour when JS disabled

vanitabarrett:
Trello: https://trello.com/c/LQKpXVSw/186-audit-components-that-use-a-toggle-and-update-to-hide-the-toggle-link-when-js-is-disabled

Previously, these components would auto-expand all contents when JS was disabled (which is good), but also continue to show the toggle link which would be confusing to users. This PR only shows the toggle link if JS is enabled.

## Before (published dates)
screen shot 2018-10-01 at 16 08 18

## After (published dates)
screen shot 2018-10-01 at 16 07 28

## Before (publisher metadata)
screen shot 2018-10-01 at 16 08 33

## After (publisher metadata)
screen shot 2018-10-01 at 16 07 08

Links:
https://government-frontend-pr-1110.herokuapp.com/component-guide/publisher-metadata
https://government-frontend-pr-1110.herokuapp.com/component-guide/published-dates

### DIFF
--- a/app/assets/stylesheets/components/_published-dates.scss
+++ b/app/assets/stylesheets/components/_published-dates.scss
@@ -2,6 +2,14 @@
   @include core-16;
 }
 
+.app-c-published-dates__toggle {
+  display: none;
+
+  .js-enabled & {
+    display: inline-block;
+  }
+}
+
 .app-c-published-dates__change-history {
   margin: 20px 0;
 }
--- a/app/assets/stylesheets/components/_publisher-metadata.scss
+++ b/app/assets/stylesheets/components/_publisher-metadata.scss
@@ -26,3 +26,12 @@
     display: inline-block;
   }
 }
+
+.app-c-publisher-metadata__toggle,
+.app-c-publisher-metadata__definition-others {
+  display: none;
+
+  .js-enabled & {
+    display: inline-block;
+  }
+}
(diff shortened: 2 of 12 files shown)


app/assets/stylesheets/components/_published-dates.scss
```diff
@@ -1,5 +1,13 @@
 .app-c-published-dates {
   @include core-16;
+
+  [data-controls] {
```
DilwoarH: is there a reason why you cannot target a class?

vanitabarrett: No, good point. I'll change it now 👍

vanitabarrett: @DilwoarH Done 🙂

app/assets/stylesheets/components/_published-dates.scss
```diff
@@ -2,6 +2,14 @@
   @include core-16;
 }
 
+.app-c-published-dates__toggle {
+  display: none;
+
+  .js-enabled & {
```
DilwoarH: Nit pick: not sure what our standards are but I've always seen the `&` selector used via the following format `&.js-enabled` - correct me if I'm wrong about is. 

Other than this, everything looks fine.

vanitabarrett: My understanding is that they have a slightly different meaning? I may be wrong:

```
.element-one {
    &.element-two {}
}
```
This compiles to `.element-one .element-two`

Whereas
```
.element-one {
    .element-two & {}
}
```
compiles to `.element-two .element-one`

In this case, I think `js-enabled` is set on the `<body>`, so that's why the second variation is needed.

PR #1175, opened 2018-12-03
Use default news image on news article pages

emmabeynon:
For https://trello.com/c/UXUxzEv4/512-amend-government-frontend-logic-to-use-default-organisation-image

In cases where a news article doesn't have an image, we use its primary
publishing organisation's default image.


emmabeynon:
@1pretz1 Sure, at the moment government-frontend is using the url rather than the high res url for images when a news article has one, so we're just using the same logic here for default news images and using url.

1pretz1:
@emmabeynon Okay cool! 👍

PR #1178, opened 2018-12-04
Default to placeholder image for newslike content

emmabeynon:
For https://trello.com/c/UXUxzEv4/512-amend-government-frontend-logic-to-use-default-organisation-image

This ensures that a placeholder image is displays in cases where news-like content does not have an image or a default news image.  We also ensure that world location news articles can display a default news image if they have one in cases where they don't have an image.

### DIFF
--- a/app/presenters/news_article_presenter.rb
+++ b/app/presenters/news_article_presenter.rb
@@ -8,7 +8,7 @@ class NewsArticlePresenter < ContentItemPresenter
   include ContentItem::Metadata
 
   def image
-    content_item["details"].dig("image") || default_news_image
+    content_item["details"].dig("image") || default_news_image || placeholder_image
   end
 
 private
@@ -16,5 +16,9 @@ private
   def default_news_image
     organisation = content_item["links"].dig("primary_publishing_organisation")
     organisation[0].dig("details", "default_news_image") if organisation.present?
+  end
+
+  def placeholder_image
+    "https://assets.publishing.service.gov.uk/government/assets/placeholder.jpg"
   end
 end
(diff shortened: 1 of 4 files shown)


emmabeynon:
Hopefully this should cover it @kevindew.  The only outstanding presenters that call image are:
- Fatality Notices - this returns a hard-coded MoD crest
- Take Part - these don't seem to be tagged to organisations and the placeholder image doesn't feel  appropriate for these pages e.g. https://www.gov.uk/government/get-involved/take-part/volunteer

app/presenters/content_item/image.rb
```diff
@@ -0,0 +1,18 @@
+module ContentItem
+  module Image
```
kevindew: Small suggestion that you rename this as NewsImage

app/presenters/content_item/image.rb
```diff
@@ -0,0 +1,18 @@
...
+    def image
+      content_item["details"].dig("image") || default_news_image || placeholder_image
```
kevindew: Minor - but since we're moving this code around how do you feel about changing `content_item["details"].dig("image")` to `content_item.dig("details", "image")` there's another one below too.

I don't really know why someone did that originally, seems a bit weird to do a lookup then a dig :-/

emmabeynon:
@kevindew Your comments should be addressed now 👍

app/presenters/case_study_presenter.rb
```diff
@@ -2,10 +2,7 @@ class CaseStudyPresenter < ContentItemPresenter
   include ContentItem::Body
   include ContentItem::Metadata
   include ContentItem::TitleAndContext
-
-  def image
-    content_item["details"]["image"]
-  end
+  include ContentItem::NewsImage
```
kevindew: Should this be in case study? I thought we didn't support that type in rummager so probably should be the same here

emmabeynon: You're absolutely right, now removed.

PR #1181, opened 2018-12-06
Fix placeholder image bug

emmabeynon:
It should return a hash instead of a string.


kevindew:
PR begone from the angry seal

PR #1190, opened 2018-12-18
Display image credit for lead images

leenagupte:
Trello: https://trello.com/c/hGmJ5BfV 
Related to: 
* https://github.com/alphagov/govuk-content-schemas/pull/850
* https://github.com/alphagov/content-publisher/pull/627

## Motivation
Content Publisher allows publishers to add an image credit, but we are not displaying this data on the frontend

### Desktop
screen shot 2018-12-18 at 15 22 35

### Mobile
screen shot 2018-12-18 at 16 24 09

Paired with @DilwoarH

### DIFF
--- a/app/assets/stylesheets/components/_figure.scss
+++ b/app/assets/stylesheets/components/_figure.scss
@@ -36,3 +36,11 @@
     margin-top: $gutter-one-third;
   }
 }
+
+.app-c-figure__figcaption-text {
+  margin-bottom: 0;
+
+  @include media(desktop, tablet) {
+    margin-bottom: $gutter-one-third;
+  }
+}
--- a/app/views/components/_figure.html.erb
+++ b/app/views/components/_figure.html.erb
@@ -1,12 +1,21 @@
 <%
   alt ||= ''
   caption ||= ''
+  credit ||= ''
 %>
 <figure class="app-c-figure">
   <img class="app-c-figure__image" src="<%= src %>" alt="<%= alt %>">
-  <% unless caption.blank? %>
     <figcaption class="app-c-figure__figcaption">
-      <%= caption %>
+      <% unless caption.blank? %>
+        <p class="app-c-figure__figcaption-text">
+          <%= caption %>
+        </p>
+      <% end %>
+
+      <% if credit.present? %>
+          <p class="app-c-figure__figcaption-credit">
+            Image credit: <%= credit %>
+          </p>
+      <% end %>
     </figcaption>
-  <% end %>
 </figure>
--- a/app/views/components/_figure.html.erb
+++ b/app/views/components/_figure.html.erb
@@ -14,7 +14,7 @@
 
       <% if credit.present? %>
           <p class="app-c-figure__figcaption-credit">
-            Image credit: <%= credit %>
+            <%= I18n.t("components.figure.image_credit", credit: credit) %>
           </p>
       <% end %>
     </figcaption>
--- a/app/views/content_items/case_study.html.erb
+++ b/app/views/content_items/case_study.html.erb
@@ -23,6 +23,7 @@
       <%= render 'components/figure',
           src: @content_item.image["url"],
           alt: @content_item.image["alt_text"],
+          credit: @content_item.image["credit"],
           caption: @content_item.image["caption"] if @content_item.image %>
 
       <%= render 'govuk_publishing_components/components/govspeak',
--- a/app/views/content_items/fatality_notice.html.erb
+++ b/app/views/content_items/fatality_notice.html.erb
@@ -25,6 +25,7 @@
       <%= render 'components/figure',
           src: @content_item.image["url"],
           alt: @content_item.image["alt_text"],
+          credit: @content_item.image["credit"],
           caption: @content_item.image["caption"] if @content_item.image %>
       <%= render 'govuk_publishing_components/components/govspeak',
           content: @content_item.body.html_safe,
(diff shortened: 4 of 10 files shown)


app/views/components/_figure.html.erb
```diff
@@ -9,4 +10,10 @@
       <%= caption %>
     </figcaption>
   <% end %>
+
+  <% if credit.present? %>
+    <figcredit class="app-c-figure__figcredit">
```
alex-ju: I assume this wasn't meant to be custom element but rather part of the `<figcaption>` or a separate `<span>`?

leenagupte: What is `figcaption`? We were trying to follow the pattern defined by captions, but weren't sure where this element is defined.

leenagupte: Should we just make it a `span`?

alex-ju: `<figcaption>` is a defined element representing a caption or a legend associated with an image/figure. They must be placed inside a ` <figure>` element to be semantically correct. There is no html element for credits though, so we can use a generic tag (i.e. `<span>` or we can include the credits inside the `<figcaption>` element.

leenagupte: TIL! I thought figcaption was something we'd made up! 

I'll use `figcaption` for consistency.

vanitabarrett:
@leenagupte Would it make more sense to just include the credit as part of the original image caption element? E.g:

```
<figcaption class="app-c-figure__figcaption">
      <%= caption %>

     <% if credit.present? %>
          <p>Image credit: <%= credit %></p>
     <% end %>
    </figcaption>
  <% end %>
```

app/views/components/_figure.html.erb
alex-ju: As @vanitabarrett mentioned, `<figure>` should only have one `<figcaption>` so it probably makes more sense to be included in the exiting tag.

app/assets/stylesheets/components/_figure.scss
```diff
@@ -36,3 +36,11 @@
     margin-top: $gutter-one-third;
   }
 }
+
+.app-c-figure__figcaption-text {
+  margin-bottom: $gutter-one-third;
+
+  @include media(mobile) {
```
vanitabarrett: We normally try and do mobile-first development, which means our CSS should be based on mobile devices and we should add any other styling for larger screen sizes in media queries. So in this case, it would mean swapping round the CSS you've added here.

app/views/components/_figure.html.erb
```diff
@@ -1,12 +1,21 @@
...
   caption ||= ''
+  credit ||= ''
 %>
 <figure class="app-c-figure">
   <img class="app-c-figure__image" src="<%= src %>" alt="<%= alt %>">
-  <% unless caption.blank? %>
     <figcaption class="app-c-figure__figcaption">
-      <%= caption %>
+      <% unless caption.blank? %>
+        <div class="app-c-figure__figcaption-text">
```
vanitabarrett: Could we use paragraph tags here instead of divs? For both `app-c-figure__figcaption-text` and `app-c-figure__figcaption-credit`

app/views/components/_figure.html.erb
```diff
@@ -1,12 +1,21 @@
...
+          <%= caption %>
+        </div>
+      <% end %>
+
+      <% if credit.present? %>
+          <div class="app-c-figure__figcaption-credit">
+            Image credit: <%= credit %>
```
vanitabarrett: Should "Image credit" come from a translation file?

app/views/components/_figure.html.erb
leenagupte: Sure! Not sure what the difference re: `<p>` and `<div>` is to be honest.

app/views/components/_figure.html.erb
leenagupte: Do you know what happens if there isn't a translation available for the other languages?

app/views/components/_figure.html.erb
vanitabarrett: Visually, they probably look the same unless you add your own styling. It's more about the semantics - a <p> (paragraph tag) is designed to wrap text, whereas a <div> is just a general containing element (which we tend to use for grouping or layout). Using semantic HTML means we are properly describing what the document/page is made up of: a load of `div` elements won't tell you a lot, but things like `p`, `image`, `a` etc give you more information.

vanitabarrett: I hope that makes sense! 🙂

app/views/components/_figure.html.erb
vanitabarrett: I thought it fell back to the english one... but you can provide a default too?

app/assets/stylesheets/components/_figure.scss
leenagupte: @vanitabarrett Updated and fixed up.

app/views/components/_figure.html.erb
leenagupte: You're right.

> The I18n library will use English as a default locale, i.e. if a different locale is not set, :en will be used for looking up translations.

From: https://guides.rubyonrails.org/i18n.html

app/views/components/_figure.html.erb
```diff
@@ -14,7 +14,7 @@
 
       <% if credit.present? %>
           <p class="app-c-figure__figcaption-credit">
-            Image credit: <%= credit %>
+            <%= I18n.t("components.figure.image_credit") %>: <%= credit %>
```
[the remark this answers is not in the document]
leenagupte: I like that better!

app/views/components/_figure.html.erb
leenagupte: @vanitabarrett Updated and fixed up.

vanitabarrett:
Looks great! 💯

PR #1211, opened 2019-01-17
Add related links A/A test to all routes

karlbaker02:
This PR adds a new concern `AATestable`, which will be used to run an A/A test to determine
variance in our metrics and help guide metrics to use for upcoming A/B tests on related links.

Solo: @karlbaker02

---

Visual regression results:
https://government-frontend-pr-1211.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1211.herokuapp.com/component-guide


karlbaker02:
@sihugh Thanks, that is being sorted as we speak.

sihugh:
Actually we can merge and deploy this because the cookie doesn't get set unless I've deployed the CDN changes

PR #1298, opened 2019-04-03
Add language names to new locales

bevanloon:
We have received some language names to apply to the new locales we've
added. Include these so the frontend will display the language names
nicely translated.


thomasleese:
Can't say anything about the validity of the translations...

bevanloon:
Thanks Tom - the translations have been sent to us so should be good 🤞

Issue #254, opened 2017-02-22
Translation files aren't being updated with `regenerate` task


fofr:
Running `bundle exec rake translation:regenerate` should only update the strings that have been modified.

There's a couple of problems:
* When adding translations we haven't re-run this task
* The task strips out the i18n property when it's run, it shouldn't

Suggested fix:
* Fix the i18n problem
* Cleanup locales
* Add test step that fails if `bundle exec rake translation:regenerate` leads to changes (like the build steps in content schemas)

This is what happens on master:
https://github.com/alphagov/government-frontend/compare/cleanup-locales

cc @gpeng @bevanloon @andrewgarner

PR #1382, opened 2019-06-19
Add new locale

MuriloDalRi:
Adds support for Kazakh (kk)

[Trello](https://trello.com/c/fVLk3RBY/1085-add-translation-option-for-kazakh)

---

Visual regression results:

Component guide for this PR:


MuriloDalRi:
@bevanloon I regenerated the translations and imported all as YML.

The docs say `There's no timeline for how frequently this is done, so you can expect many translation values to be missing in non EN locales.`

So I'm assuming that's what happened here and that's ok? Or should I import just the kk one?

bevanloon:
@MuriloDalRi yeah -I think I just added what I needed last time rather than all the extras, but it's a judgement call really. This [deletion](https://github.com/alphagov/government-frontend/pull/1382/files#diff-2bb925571600311d122bd27feec93485L2), however, doesn't look quite right to me and might account for your test failure.

Perhaps the translation manager isn't smart enough to deal with that.

MuriloDalRi:
@bevanloon Okay I'll have a look at fixing that but if it's too much faff I'll just add the Kazakh one. I found this old issue which seems like it might be the same problem: https://github.com/alphagov/government-frontend/issues/254

PR #1386, opened 2019-06-21
Display new recommended related links

karlbaker02:
This PR adds support for the new recommended related links to be shown based on the existence of a HTTP header set by our CDN. To achieve this we add support in general for feature flags via a `FeatureToggler` class.

Should the header exist, the new recommended related links will only be shown if there are no existing `ordered_related_items` in the content item, to ensure that we do not override manually curated links.

Dependent on https://github.com/alphagov/govuk-content-schemas/pull/908.

---

Visual regression results:
https://government-frontend-pr-1386.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1386.herokuapp.com/component-guide

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -56,6 +56,11 @@ private
 
   def load_content_item
     content_item = Services.content_store.content_item(content_item_path)
+
+    if Services.feature_toggler.use_recommended_related_links?(content_item['links'], request.headers)
+      content_item['links']['ordered_related_items'] = content_item['links'].fetch('suggested_ordered_related_items', [])
+    end
+
     @content_item = PresenterBuilder.new(content_item, content_item_path).presenter
   end
 
--- a/app/models/feature_toggler.rb
+++ b/app/models/feature_toggler.rb
@@ -1,0 +1,12 @@
+class FeatureToggler
+  def initialize(feature_flags)
+    @feature_flags = feature_flags
+  end
+
+  def use_recommended_related_links?(content_item_links, request_headers)
+    return false if content_item_links.nil?
+
+    content_item_links.fetch('ordered_related_items', []).empty? &&
+      @feature_flags.feature_enabled?(FeatureFlagNames.recommended_related_links, request_headers)
+  end
+end
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -130,6 +130,38 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_equal content_item['title'], assigns[:content_item].title
   end
 
+  test "gets item from content store and keeps existing ordered_related_items when feature flag header not specified" do
+    content_item = content_store_has_schema_example('case_study', 'case_study')
+
+    get :show, params: { path: path_for(content_item) }
+    assert_response :success
+    assert_equal content_item['links']['ordered_related_items'], assigns[:content_item].content_item['links']['ordered_related_items']
+  end
+
+  test "gets item from content store and keep existing ordered_related_items when feature flag header is specified but links already exist" do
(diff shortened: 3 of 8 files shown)


app/models/feature_toggler.rb
```diff
@@ -0,0 +1,12 @@
+class FeatureToggler
+  def initialize(feature_flags)
+    @feature_flags = feature_flags
+  end
+
+  def use_recommended_related_links?(content_item_links, request_headers)
+    return false if content_item_links.nil?
+
+    (!content_item_links.has_key?('ordered_related_items') || content_item_links['ordered_related_items'].empty?) &&
```
barrucadu: I think `content_item_links.fetch('ordered_related_items', []).empty?` would be clearer

app/controllers/content_items_controller.rb
```diff
@@ -56,6 +56,11 @@ def set_guide_draft_access_token
 
   def load_content_item
     content_item = Services.content_store.content_item(content_item_path)
+
+    if Services.feature_toggler.use_recommended_related_links?(content_item['links'], request.headers)
+      content_item['links']['ordered_related_items'] = content_item['links'].fetch('suggested_ordered_related_items', [])
```
barrucadu: I think you need to set a response header here to avoid the two variants being mixed up in the cache.  Have you seen https://github.com/alphagov/govuk_ab_testing?  It does that and also some analytics stuff.

karlbaker02: We're not using the govuk_ab_testing gem here purposefully, as this isn't a test but instead a mechanism to achieve rollback if we ever need to (for example, if we realise that some proportion of generated links are in some way bad). We'll be setting the request header in the CDN which will effectively invalidate the cache if we change this header, which means we shouldn't need to set a response header (again, we're only ever going to be showing one of these 'variants' at any time).

app/models/feature_toggler.rb
karlbaker02: Good shout.

test/controllers/content_items_controller_test.rb
```diff
@@ -130,6 +130,34 @@ class ContentItemsControllerTest < ActionController::TestCase
...
+
+  test "gets item from content store and replaces ordered_related_items when feature flag header is specified and there are no existing links" do
+    request.headers[FeatureFlagNames.recommended_related_links] = 'true'
+
+    content_item = content_store_has_schema_example('case_study', 'case_study')
+
+    get :show, params: { path: path_for(content_item) }
+    assert_response :success
+    assert_equal assigns[:content_item].content_item['links']['ordered_related_items'], content_item['links']['suggested_ordered_related_items']
+  end
```
barrucadu: I'm not sure why this test and the one above it are testing different things.  Shouldn't you explicitly set the `ordered_related_items` in the above test and delete it in this one?

If `guide` and `case_study` just happen to have the property that examples of the former have `ordered_related_items` and examples of the latter don't, I still think it should be made explicit here.

karlbaker02: I see where you're coming from - they are testing different things because one of the content schema examples has existing `ordered_related_items` and the other has no `ordered_related_items` but does have `suggested_ordered_related_items`. This was through an intentional change to the content schemas some time ago, and it is necessary to test these behaviours as we don't want to be overriding existing related links where they exist.

To your second point, how do you feel this would be best made explicit?

barrucadu: Maybe something like

```
# first test
content_item = ...
content_item['links']['ordered_related_items'] = [...]
content_item['links']['suggested_ordered_related_items'] = [...]
# check the ordered_related_items is unchanged

# first test
content_item = ...
content_item['links']['ordered_related_items'] = []
content_item['links']['suggested_ordered_related_items'] = [...]
# check the ordered_related_items is changed
```

I don't see why overwriting the links would be a problem here.

PR #1398, opened 2019-06-27
Change feature flag names to represent headers set at the CDN

karlbaker02:
This PR changes the feature flag names used in `FeatureFlagNames` to represent headers as we set them at the CDN. This allows us to output response headers for the feature flags and in the `Vary` header which are consistent with others headers that we set (see below):

**Before (first and last headers)**
Screen Shot 2019-06-27 at 16 56 44

**After (first and last headers)**
Screen Shot 2019-06-27 at 16 45 15

---

Visual regression results:
https://government-frontend-pr-1398.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1398.herokuapp.com/component-guide

### DIFF
--- a/app/models/http_feature_flags.rb
+++ b/app/models/http_feature_flags.rb
@@ -3,19 +3,25 @@ class HttpFeatureFlags
     @feature_flags = {}
   end
 
-  def add_http_feature_flag(header_name, val)
-    @feature_flags[header_name] = val
+  def add_http_feature_flag(feature_flag_name, val)
+    @feature_flags[get_header_name(feature_flag_name)] = val
   end
 
-  def get_feature_flag(header_name)
-    @feature_flags[header_name]
+  def get_feature_flag(feature_flag_name)
+    @feature_flags[get_header_name(feature_flag_name)]
   end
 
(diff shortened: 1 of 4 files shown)


bilbof:
It looks like visual regression link is missing screenshots.

app/models/http_feature_flags.rb
```diff
@@ -3,19 +3,25 @@ def initialize
...
   end
 
   def self.instance
     @instance ||= HttpFeatureFlags.new
   end
+
+private
+
+  def get_header_name(feature_flag_name)
+    "HTTP_#{feature_flag_name.upcase.tr('-', '_')}"
```
bilbof: Is the `feature_flag_name` string only used here? If so, could you store it as that string so you don't need to transform it?

```ruby
module FeatureFlagNames
    RECOMMENDED_RELATED_LINKS = 'GOVUK_USE_RECOMMENDED_RELATED_LINKS'.freeze
end
```

karlbaker02: As discussed 🙂

PR #1400, opened 2019-06-28
Use request header for related links to determine response header

karlbaker02:
This PR fixes the related links response header to be determined by the request header. If the request header is not set or does not match the required value as defined by the `feature_flags` initializer, we set the response header to `false`; if the request header is set and matches the required value, it's set to `true`.

This will enable us to change the request header at CDN level if we need to change related links.

---

Visual regression results:
https://government-frontend-pr-1400.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1400.herokuapp.com/component-guide

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -115,7 +115,10 @@ private
 
   def set_use_recommended_related_links_header
     response.headers['Vary'] = [response.headers['Vary'], FeatureFlagNames.recommended_related_links].compact.join(', ')
-    response.headers[FeatureFlagNames.recommended_related_links] = Services.feature_toggler.feature_flags.get_feature_flag(FeatureFlagNames.recommended_related_links)
+
+    related_links_request_header = RequestHelper.get_header(FeatureFlagNames.recommended_related_links, request.headers)
+    required_header_value = Services.feature_toggler.feature_flags.get_feature_flag(FeatureFlagNames.recommended_related_links)
+    response.headers[FeatureFlagNames.recommended_related_links] = (related_links_request_header == required_header_value).to_s
   end
 
   def set_expiry
--- a/app/lib/request_helper.rb
+++ b/app/lib/request_helper.rb
@@ -1,0 +1,9 @@
+module RequestHelper
+  def self.get_header(header_name, request_headers)
+    request_headers[headerise(header_name)]
+  end
+
+  def self.headerise(header_name)
+    "HTTP_#{header_name.upcase.tr('-', '_')}"
+  end
+end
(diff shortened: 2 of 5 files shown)


app/lib/request_helper.rb
```diff
@@ -0,0 +1,9 @@
+module RequestHelper
+  def self.get_header(header_name, request_headers)
+    request_headers[headerise(header_name)]
+  end
```
bilbof: Did you add this somewhere else in this codebase?

app/controllers/content_items_controller.rb
```diff
@@ -115,7 +115,9 @@ def set_access_control_allow_origin_header
 
   def set_use_recommended_related_links_header
     response.headers['Vary'] = [response.headers['Vary'], FeatureFlagNames.recommended_related_links].compact.join(', ')
-    response.headers[FeatureFlagNames.recommended_related_links] = Services.feature_toggler.feature_flags.get_feature_flag(FeatureFlagNames.recommended_related_links)
+
+    related_links_response_header = RequestHelper.get_header(FeatureFlagNames.recommended_related_links, request.headers) == Services.feature_toggler.feature_flags.get_feature_flag(FeatureFlagNames.recommended_related_links)
```
bilbof: A couple of variable assignments might help readability here.

e.g. but with proper variable names :)

```ruby
x = FeatureFlagNames.recommended_related_links
y = RequestHelper.get_header(x, request.headers) 
z = Services.feature_toggler.feature_flags.get_feature_flag(x)
response.headers[x] = y == z
```

test/controllers/content_items_controller_test.rb
```diff
@@ -178,24 +178,36 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_equal assigns[:content_item].content_item['links']['ordered_related_items'], content_item['links']['suggested_ordered_related_items']
   end
 
-  test "sets the Govuk-Use-Recommended-Links header to true when using recommended links" do
-    HttpFeatureFlags.instance.add_http_feature_flag(FeatureFlagNames.recommended_related_links, 'true')
+  test "sets the Govuk-Use-Recommended-Links response header to false when request header is not set" do
+    # HttpFeatureFlags.instance.add_http_feature_flag(FeatureFlagNames.recommended_related_links, 'false')
```
bilbof: Did you mean to remove this?

app/lib/request_helper.rb
karlbaker02: I think I am using this in a few places in this PR now 🙂

app/controllers/content_items_controller.rb
karlbaker02: Sure thing

PR #1429, opened 2019-07-26
Ensure suggested related links respect overrides

karlbaker02:
This PR adds an additional check to ensure that `suggested_ordered_related_items` are only assigned if there are both no `ordered_related_items` and no `ordered_related_items_overrides`. This is necessary because publishers can manually add overrides and expect them to be displayed instead of any existing links.

Depends on https://github.com/alphagov/govuk-content-schemas/pull/918.

---

Visual regression results:
https://government-frontend-pr-1429.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1429.herokuapp.com/component-guide

### DIFF
--- a/app/models/feature_toggler.rb
+++ b/app/models/feature_toggler.rb
@@ -9,6 +9,7 @@ class FeatureToggler
     return false if content_item_links.nil?
 
     content_item_links.fetch('ordered_related_items', []).empty? &&
+      content_item_links.fetch('ordered_related_items_overrides', []).empty? &&
       @feature_flags.feature_enabled?(FeatureFlagNames.recommended_related_links, request_headers)
   end
 end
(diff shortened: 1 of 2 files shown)


barrucadu:
What's the difference between `ordered_related_items` and `ordered_related_items_overrides`?  Could a document have both of them set?

app/models/feature_toggler.rb
```diff
@@ -9,6 +9,7 @@ def use_recommended_related_links?(content_item_links, request_headers)
     return false if content_item_links.nil?
 
     content_item_links.fetch('ordered_related_items', []).empty? &&
+      content_item_links.fetch('ordered_related_items_overrides', []).empty? &&
       @feature_flags.feature_enabled?(FeatureFlagNames.recommended_related_links, request_headers)
```
hannako: should this be `feature_flags.feature_enabled?` because there's an attr_reader

karlbaker02:
@barrucadu From my understanding, a document can have both `ordered_related_items` and `ordered_related_items_overrides`. The latter is something that can manually be set by publishers to override whatever is in the former.

PR #1447, opened 2019-08-19
Add detection of contextual title translation

maxgds:
https://trello.com/c/iQcgz7lt/25-2-mark-language-on-page-kicker-section-of-page

This adds a helper method which detects whether the string requested
was returned by the current locale's translation file. If not, it returns
the default locale.

This helper is in turn used on the contextual headings, allowing us to
pass the information on to the component that displays the heading, meaning
we will be able to markup the contextual header's actual language, supporting
screenreaders and making the page more semantic.

---

Visual regression results:
https://government-frontend-pr-1447.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1447.herokuapp.com/component-guide

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -3,6 +3,18 @@ module ApplicationHelper
     I18n.t("i18n.direction", locale: I18n.locale, default: "ltr")
   end
 
+  def t_locale_fallback(key, options = {})
+    options['locale'] = I18n.locale
+    options[:fallback] = nil
+    translation = I18n.t(key, options)
+
+    if translation.nil? || translation.include?("translation missing")
+      I18n.default_locale
+    else
+      return nil
+    end
+  end
+
   def wrapper_class
     "direction-#{page_text_direction}" if page_text_direction
   end
--- a/app/presenters/content_item/title_and_context.rb
+++ b/app/presenters/content_item/title_and_context.rb
@@ -4,6 +4,7 @@ module ContentItem
       {
         title: title,
         context: I18n.t("content_item.schema_name.#{document_type}", count: 1),
+        context_locale: t_locale_fallback("content_item.schema_name.#{document_type}", count: 1),
         average_title_length: "long"
       }
     end
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -1,5 +1,6 @@
 class ContentItemPresenter
   include ContentItem::Withdrawable
+  include ApplicationHelper
 
   attr_reader :content_item,
               :requested_content_item_path,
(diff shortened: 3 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -3,6 +3,18 @@ def page_text_direction
     I18n.t("i18n.direction", locale: I18n.locale, default: "ltr")
   end
 
+  def t_locale_fallback(key, options = {})
+    options['locale'] ||= I18n.locale
```
vanitabarrett: Are we ever setting `options['locale']` outside of the tests? It would be great not to have logic specific to tests. If you want to set the locale inside your test, you should be able to set it: `I18n.locale = :de` (for example).

vanitabarrett:
@maxgds have you got an example URL for this?

app/helpers/application_helper.rb
maxgds: We don't currently, no, but when I set I18n.locale = :de it caused a cascade of failing tests because I couldn't figure out how to unset it afterwards.

maxgds:
> @maxgds have you got an example URL for this?

Unfortunately not. After sinking loads of time into making this work I found that all the real-world examples I could find seem to be those dodgy exceptions that think they're English despite the content being another language.

app/helpers/application_helper.rb
vanitabarrett: Are you able to do something like this:

```
I18n.with_locale(:de) do
    assert .....
end
```

maxgds:
I've now found a relevant url: https://government-frontend-pr-1447.herokuapp.com/government/news/changes-to-uk-visa-application-process-in-germany.de however it's not going to do anything until the publishing components are accepting the new parameter - that's queued up in a separate PR.

app/helpers/application_helper.rb
maxgds: Your test suggestion is working, thanks!

app/helpers/application_helper.rb
```diff
@@ -3,6 +3,18 @@ def page_text_direction
...
 
+  def t_locale_fallback(key, options = {})
+    options['locale'] = I18n.locale
+    options[:fallback] = false
+    translation = I18n.t(key, options)
+
+    if translation.nil? || translation.include?("translation missing")
+      I18n.default_locale
+    else
+      false
```
injms: Can we use `nil` instead of `false` please? I found that `false` was cast to a string and displayed.
```suggestion
      nil
```

maxgds: > Can we use `nil` instead of `false` please? I found that `false` was cast to a string and displayed.

Fixed.

maxgds:
@injms please revisit this and confirm if it's good to merge now.

PR #1449, opened 2019-08-20
Prevent errors for missing public_updated_at

kevindew:
This stops document collections raising errors when a public_updated_at
value is not set on a document. This is often the case with draft GOV.UK
documents, however Whitehall tends to fill in this field which is why
this error hasn't been seen before.

---

Visual regression results:

Component guide for this PR:

### DIFF
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -40,7 +40,7 @@ class DocumentCollectionPresenter < ContentItemPresenter
           }
         },
         metadata: {
-          public_updated_at: Time.zone.parse(link["public_updated_at"]),
+          public_updated_at: link["public_updated_at"]&.then { |time| Time.zone.parse(time) },
           document_type: I18n.t("content_item.schema_name.#{link['document_type']}",
                                 count: 1,
                                 default: nil),
--- a/test/presenters/document_collection_presenter_test.rb
+++ b/test/presenters/document_collection_presenter_test.rb
@@ -88,6 +88,21 @@ class DocumentCollectionPresenterTest
 
       assert_nil grouped.first[:metadata][:document_type]
     end
+
+    test 'it handles public_updated_at not being specified' do
+      schema_data = schema_item
+
+      document = schema_data["links"]["documents"].first.tap do |link|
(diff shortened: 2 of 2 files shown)


app/presenters/document_collection_presenter.rb
```diff
@@ -40,7 +40,7 @@ def group_document_links(group, group_index)
           }
         },
         metadata: {
-          public_updated_at: Time.zone.parse(link["public_updated_at"]),
+          public_updated_at: link["public_updated_at"]&.then { |time| Time.zone.parse(time) },
```
1pretz1: I tried to quickly re-write this as I hadn't seen `then` be given a block before (cool you can do something like this btw), here's my attempt, don't think it's better though so feel free to merge :) `link.fetch("public_updated_at", nil) || Time.zone.parse(link["public_updated_at"])`

kevindew: Sure, `then` was added in Ruby 2.6 as an alias for `yield_self` added in 2.5 so you don't see them much in the wild. It's a really useful method though: https://mlomnicki.com/yield-self-in-ruby-25/

Your suggestion there won't work as the right side of the `||` will be called when the left side returns nil. The alternative approach to this is through the ternary operator: `link["public_updated_at"] ? Time.zone.parse(link["public_updated_at"]) : nil,` or a normal conditional: `Time.zone.parse(link["public_updated_at"]) if link["public_updated_at"]` - though I always think normal conditionals are a bit confusing inside a hash.

1pretz1: Oh yeah of course, my bad! Personally I like `Time.zone.parse(link["public_updated_at"]) if link["public_updated_at"]` as it reads quite nicely to me although it doesn't explicitly return nil.

kevindew: I'm not a big fan of having that when it's within a hash. For instance this would be:

```
metadata: {
  public_updated_at: Time.zone.parse(link["public_updated_at"]) if link["public_updated_at"],
}
```

and I read that as it being ambiguous exactly what on the line the if applies to. 

I'm going to roll with the `then`

PR #1460, opened 2019-09-03
Add basic auth to heroku app

barrucadu:
Copied from: https://github.com/alphagov/finder-frontend/blob/master/app.json

---

Visual regression results:
https://government-frontend-pr-1460.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1460.herokuapp.com/component-guide

### DIFF
--- a/app.json
+++ b/app.json
@@ -25,7 +25,14 @@
     },
     "HEROKU_APP_NAME": {
       "required": true
-    }
+    },
+    "BASIC_AUTH_USERNAME": {
+      "required": true
+    },
+    "BASIC_AUTH_PASSWORD": {
+      "required": true
+    },
+    "REQUIRE_BASIC_AUTH": "true"
   },
   "image": "heroku/ruby",
   "buildpacks": [


nickcolley:
Can we document the username and password clearly for developers?

barrucadu:
It'll be the same username and password as all the other preview apps, though I don't know where that's documented.

bevanloon:
The password is documented in a few places on our internal wiki.
It's possibly also somewhere in secrets?

nickcolley:
Does this mean documentation such as component guides will be locked down?

http://government-frontend.herokuapp.com/component-guide

If possible it'd be good to have pages like this available to everyone as other people such as designers will need to reference them.

https://docs.publishing.service.gov.uk/manual/components.html#component-guides

barrucadu:
Yes, everything will be locked down.   Maybe a robots.txt would be better

nickcolley:
I think the username and password is a really good idea for the indexed pages, but I'm just wondering if we can do it in a way that doesnt close off the documentation that's also served by our frontend applications.

https://www.gov.uk/guidance/government-design-principles#make-things-open-it-makes-things-better

sihugh:
How about this then: https://github.com/alphagov/government-frontend/pull/1461?

PR #1483, opened 2019-09-12
Pass _ga client id to external signon services

bilbof:
On pages such as https://www.gov.uk/log-in-file-self-assessment-tax-return/sign-in/prove-identity
we redirect users to services such as tax.service.gov.uk to complete their sign in.

This will pass the `_ga` cookie as the `_ga` param as part of cross-domain tracking, so that we can measure user journeys across domains.

This will only pass the _ga param to other .gov.uk services. Currently these are:
- www.tax.service.gov.uk
- www.signin.service.gov.uk
- www.ruralpayments.service.gov.uk

Pages to check: 

- https://government-frontend-pr-1483.herokuapp.com/log-in-file-self-assessment-tax-return/sign-in/prove-identity
- https://government-frontend-pr-1483.herokuapp.com/check-state-pension/sign-in/prove-identity
- https://government-frontend-pr-1483.herokuapp.com/update-company-car-details/sign-in/prove-identity
- https://government-frontend-pr-1483.herokuapp.com/personal-tax-account/sign-in/prove-identity
- https://government-frontend-pr-1483.herokuapp.com/update-company-car-details/sign-in/prove-identity

The urls we redirect to are defined in service_sign_on content item `details.choose_sign_in.options` ([example](https://www.gov.uk/api/content/check-state-pension/sign-in)). 

Further work is required on the signin.service.gov.uk to make use of the _ga param and ensure it is included in further redirects.

Trello: https://trello.com/c/ORdXWurC/101

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,2 +1,11 @@
 //= require_tree ./modules
 //= require govuk_publishing_components/all_components
+//= require set-ga-client-id-on-form
+
+jQuery(function ($) {
+  var $form = $('.js-service-sign-in-form')
+
+  if ($form.length) {
+    new GOVUK.SetGaClientIdOnForm({ $form: $form })
+  }
+})
--- a/app/assets/javascripts/set-ga-client-id-on-form.js
+++ b/app/assets/javascripts/set-ga-client-id-on-form.js
@@ -1,0 +1,20 @@
+(function () {
+  'use strict'
+
+  window.GOVUK = window.GOVUK || {}
+  var GOVUK = window.GOVUK
+
+  function SetGaClientIdOnForm (options) {
+    if (!options.$form || !window.ga) { return }
+
+    var form = options.$form
+
+    window.ga(function(tracker) {
+      var clientId = tracker.get('clientId')
+      var action = form.attr('action')
+      form.attr('action', action + "?_ga=" + clientId)
+    });
+  }
+
+  GOVUK.SetGaClientIdOnForm = SetGaClientIdOnForm
+})(window, window.GOVUK);
--- a/spec/javascripts/set-ga-client-id-on-form.spec.js
+++ b/spec/javascripts/set-ga-client-id-on-form.spec.js
@@ -1,0 +1,23 @@
+/* global describe beforeEach it expect */
+
+var $ = window.jQuery
+
+describe('SetGaClientIdOnForm', function () {
+
+  var GOVUK = window.GOVUK
+  var tracker = { clientId: 'clientId' }
+  tracker.get = function(arg) { return this[arg] }
+  window.ga = function(callback) { callback(tracker) }
+  var form
+
+  beforeEach(function () {
+    form = $(
+      '<form class="js-service-sign-in-form" action="/endpoint"></form>'
+    )
+    setter = new GOVUK.SetGaClientIdOnForm({ $form: form })
+  })
+
+  it('sets the _ga client id as a query param on the form action', function () {
+    expect(form.attr('action')).toBe('/endpoint?_ga=clientId')
+  })
+})
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -34,7 +34,7 @@ class ContentItemsController < ApplicationController
         return
       end
 
-      redirect_to selected[:url]
+      redirect_to service_url(selected[:url])
     end
   end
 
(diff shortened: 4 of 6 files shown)


app/assets/javascripts/set-ga-client-id-on-form.js
```diff
@@ -0,0 +1,20 @@
+(function () {
+  'use strict'
+
+  window.GOVUK = window.GOVUK || {}
+  var GOVUK = window.GOVUK
```
andysellick: Nitpick: something weird going on with the indenting here?

app/assets/javascripts/set-ga-client-id-on-form.js
```diff
@@ -0,0 +1,20 @@
...
+
+   function SetGaClientIdOnForm (options) {
+    if (!options.$form || !window.ga) { return }
```
andysellick: Is it necessary to check for `options.$form` here? Can this code be called by anything other than line 9 of `application.js`?

spec/javascripts/set-ga-client-id-on-form.spec.js
```diff
@@ -0,0 +1,23 @@
+/* global describe beforeEach it spyOn expect */
```
andysellick: Is `spyOn` needed here?

bilbof: No, I grabbed this from another spec file. Should I remove the ones I didn't use?

andysellick: Yeah, just spyOn, everything else is used.

app/assets/javascripts/set-ga-client-id-on-form.js
bilbof: It could be called by something else later, so doing an argument check makes the function a bit more robust.

app/assets/javascripts/set-ga-client-id-on-form.js
```diff
@@ -0,0 +1,20 @@
...
+
+  function SetGaClientIdOnForm (options) {
+    if (!options.$form || !window.ga) { return }
+
+    var form = options.$form
+
+    window.ga(function(tracker) {
+      var clientId = tracker.get('clientId')
+      var action = form.attr('action')
+      form.attr('action', action + "?_ga=" + clientId)
```
cbaines: I guess this won't work as intended if the action URL already includes query parameters. Do you know if any action URLs include query parameters?

app/assets/javascripts/application.js
```diff
@@ -1,2 +1,11 @@
...
 //= require govuk_publishing_components/all_components
+//= require set-ga-client-id-on-form
+
+jQuery(function ($) {
+  var $form = $('.js-service-sign-in-form')
+
+  if ($form.length) {
+    new GOVUK.SetGaClientIdOnForm({ $form: $form })
+  }
+})
```
cbaines: It looks like the module system in the `govuk_frontend_toolkit` would be able to handle this. https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules

That's used elsewhere in this app, in the `track-radio-group.js` file. Is that something you considered?

app/assets/javascripts/set-ga-client-id-on-form.js
bilbof: The form currently submits without query parameters for all URLs (apart from the `_ga` param).

https://github.com/alphagov/government-frontend/pull/1483/files#diff-1e002ab60ce763d5443df324a62b2bcfR22

I checked the service sign in pages published by publisher from https://github.com/alphagov/publisher/tree/master/lib/service_sign_in and they seem to work (some of those rendered by the Heroku app are linked to in the description).

I considered using `url = new URL(action)` and then do something like `url.searchParams.set('_ga', 'clientId')` which would handle the case of multiple query params. But the constructor isn't supported by Internet Explorer (https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility).

cbaines: Ok :+1:

app/assets/javascripts/application.js
bilbof: Yes I can do, I'd need to roll the behaviour into the track-radio-group module since that module is already applied to the form at https://github.com/alphagov/government-frontend/pull/1483/files#diff-1e002ab60ce763d5443df324a62b2bcfR17, unless there's a way to apply two modules to the same element?

cbaines: Hmm, I didn't anticipate a one module per element restriction, but it looks from the code like that's the case [1].

1: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/modules.js#L31

In which case, I'd just stick with your current approach.

bilbof: OK, will do.

PR #1527, opened 2019-10-25
Add hardcoded FAQPage schema for voting-in-the-uk

sihugh:
We think that the voting intents around this page mean that users will benefit from a more curated schema.  This is intended to be a one of a kind for now, and is obviously time sensitive, hence the somewhat blunt approach.

The content for the FAQPage schema is loaded from a local config file, and presented.  It's hard to do this programmatically from the content in the content item because the sections we want don't necessarily map to an h2 and the following paragraphs (for example).  

We'll be looking into how to structure our content/apis to make things like this easier in the near future.

I've added a couple of light touch tests to ensure that they aren't broken inadvertently.

We'll be keeping the content in the yaml file in step with the published content until we find a better approach.

---

https://government-frontend-pr-1527.herokuapp.com/voting-in-the-uk

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -5,6 +5,7 @@ class ContentItemPresenter
   attr_reader :content_item,
               :requested_content_item_path,
               :base_path,
+              :slug,
               :title,
               :description,
               :schema_name,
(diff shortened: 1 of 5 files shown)


sihugh:
Thanks Bill!

PR #1534, opened 2019-10-31
Bump govuk_publishing_components

sihugh:
I think dependabot is being really slow/ignoring this version of govuk_publishing_components.

This change adds the [`data-nosnippet` attribute](https://developers.google.com/search/reference/robots_meta_tag#data-nosnippet-attr) to the cookie banner

[Changelog](https://github.com/alphagov/govuk_publishing_components/blob/master/CHANGELOG.md#2181)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -114,7 +114,7 @@ GEM
     govuk_frontend_toolkit (8.2.0)
       railties (>= 3.1.0)
       sass (>= 3.2.0)
-    govuk_publishing_components (21.8.0)
+    govuk_publishing_components (21.8.1)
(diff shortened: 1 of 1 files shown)


sihugh:
@andysellick I had originally made this PR to bump from 21.7.0 --> 21.8.1 which needed a Gemfile change, but dependabot eventually gazumped me (about an hour after 21.8.1 was released) with a PR to get to 21.8.0.  So TLDR, it's only Gemfile.lock required.

sihugh:
Looks like the cookie banner is actually used in static anyway, so I don't need this right now :-D

PR #1529, opened 2019-10-28
Remove govuk_frontend_toolkit sass dependency

andysellick:
Remove toolkit sass dependency. Should be no visual changes.

(some) Pages to check:

- https://government-frontend-pr-1529.herokuapp.com
- https://government-frontend-pr-1529.herokuapp.com/government/organisations/attorney-generals-office/about
- https://government-frontend-pr-1529.herokuapp.com/government/case-studies/get-britain-building-carlisle-park
- https://government-frontend-pr-1529.herokuapp.com/government/publications/car-show-me-tell-me-vehicle-safety-questions/car-show-me-tell-me-vehicle-safety-questions
- https://government-frontend-pr-1529.herokuapp.com/log-in-file-self-assessment-tax-return

---

Component guide for this PR:
https://government-frontend-pr-1529.herokuapp.com/component-guide

PR #1554, opened 2019-11-26
Remove govuk_frontend_toolkit dependency

vanitabarrett:
## What
Removes govuk_frontend_toolkit as a dependency, as it's not being used any more. The bulk of the work was already done by @andysellick in https://github.com/alphagov/government-frontend/pull/1529


andysellick:
Wow that was all that needed doing? Amazing! 😁

PR #1558, opened 2019-11-26
Replace govuk-lint with rubocop-govuk and scss_lint-govuk

cbaines:
The GOV.UK Lint gem is deprecated in favour of using Rubocop and
scss_lint directly with a set of shared configs.


Rakefile
```diff
@@ -2,5 +2,19 @@
...
+task :default => [:spec, 'jasmine:ci']
+RuboCop::RakeTask.new(:rubocop) do |t|
+  t.patterns = %w(app config test Gemfile)
+end
+
+SCSSLint::RakeTask.new do |t|
+  t.files = Dir.glob(["app/assets/stylesheets"])
+end
+
+task default: [:spec, "jasmine:ci", :rubocop, :scss_lint]
```
issyl0: This line and line 10 won't work together, I don't think.

Also, rubocop-govuk and scss_lint-govuk are dev/test dependencies, and all these Rakefiles run on Production deploys. You'll have to do something like in https://github.com/alphagov/email-alert-api/pull/1045.

cbaines: Ah yeah, I obviously didn't read this closely enough...

I've also added a conditional to avoid the Production environment issue :+1:

PR #1585, opened 2019-12-13
HUH-74: Verify sign-in hint on interstitial page - live test

jakubmiarka:
This is a time-limited live test of showing a Verify hint to the users
who have verified with GOV.UK Verify before.

This test should be teared down before 20th December 2019.

The way this works, is on page load, an Ajax call is made to Verify domain to check whether
the user has successfully verified with an IDP before, if they haven't nothing happens. If they
have a new box is displayed at the top of the page showing their IDP and an option to continue.

A few points to consider when reviewing:
 - the hint will only show on personal tax account sign-in page
 - CSP rules have been adjusted by adding a new domain to our S3 where we store
   IDP and Verify logos. It's an identical copy of the gov_uk_app_config gem (had to add and exclusion to robocop as it's failing the linter)
 - added tests to confirm correct behaviour of the hint show logic

Screenshot:
image

---

Visual regression results:
https://government-frontend-pr-1585.surge.sh/gallery.html

Component guide for this PR:
https://government-frontend-pr-1585.herokuapp.com/component-guide

### DIFF
--- a/app/assets/javascripts/modules/show-gov-uk-verify-hint.js
+++ b/app/assets/javascripts/modules/show-gov-uk-verify-hint.js
@@ -36,17 +36,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     function generateHtml(data) {
-      return '\
-        <div class="verify-hint-box"> \
-          <h2 class="govuk-heading-m">Someone recently signed in with '+ data['displayName'] +' on this device</h2> \
-          <div class="verify-hint-logos"> \
-            <img class="verify-hint-logos-idp" src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/'+ data['simpleId'] +'.png" alt=""> \
-            <img class="verify-hint-logos-verify" src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/govuk-verify-small-black-text-454fe97ff5e3edfb6eebdc648930c0ff675616ae7956f1d87e67b30f479d7b8d.svg" alt=""> \
-          </div> \
-          <a class="gem-c-button govuk-button" role="button" href="https://www.signin.service.gov.uk/initiate-journey/hmrc-personal-tax-account?journey_hint=idp_' + data['simpleId'] + '">Continue with '+ data['displayName'] +'</a> \
-        </div> \
-        <p class="govuk-body">If this wasn\'t you, choose one option:</p> \
-      '
+      return  '<div class="verify-hint-box">' +
+                '<h2 class="govuk-heading-m">' +
+                  'Someone recently signed in with '+ data['displayName'] +' on this device' +
+                '</h2>' +
+                '<div class="verify-hint-logos">' +
+                  '<img ' +
+                    'class="verify-hint-logos-idp"' +
+                    'src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/'+ data['simpleId'] +'.png"' +
+                    'alt=""'+
+                  '>' +
+                  '<img '+
+                    'class="verify-hint-logos-verify"' +
+                    'src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/govuk-verify-small-black-text-454fe97ff5e3edfb6eebdc648930c0ff675616ae7956f1d87e67b30f479d7b8d.svg"' +
+                    'alt=""' +
+                  '>' +
+                '</div>' +
+                '<a '+
+                  'class="gem-c-button govuk-button"' +
+                  'role="button"' +
+                  'href="https://www.signin.service.gov.uk/initiate-journey/hmrc-personal-tax-account?journey_hint=idp_' + data['simpleId'] + '"' +
+                  'onClick="GOVUK.analytics.trackEvent(\'verify-hint\', \'proceed\', { transport: \'beacon\'})"' +
+                '>' +
+                  'Continue with '+ data['displayName'] +'</a>' +
+              '</div>' +
+              '<p class="govuk-body">If this wasn\'t you, choose one option:</p>'
     }
   }
 })(window, window.GOVUK);
--- a/app/assets/javascripts/modules/show-gov-uk-verify-hint.js
+++ b/app/assets/javascripts/modules/show-gov-uk-verify-hint.js
@@ -1,0 +1,52 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.ShowGovUkVerifyHint = function () {
+    this.start = function (element) {
+      if (window.location.href.indexOf("/personal-tax-account/sign-in/prove-identity") > -1) {
+        checkLastSuccessfulIdp(element)
+      }
+    }
+    
+    this.render = function (element, data) {
+      renderHint(element, data)
+    }
+
+    function checkLastSuccessfulIdp (element) {
+      $.ajax({
+        url: 'https://www.signin.service.gov.uk/successful-idp',
+        cache: false,
+        dataType: 'jsonp',
+        timeout: 3000
+      }).then(function(data){
+          renderHint(element, data);
+      }, function(e){console.log("error", e)})
+    }
+
+    function renderHint (element, data) {
+      if (data != null && data['found'] == 'true') {
+        $(element).html(generateHtml(data)).show()
+        $('button:contains("Continue")').addClass('govuk-button--secondary')
+      }
+    }
+
+    function generateHtml(data) {
+      return '\
+        <div class="verify-hint-box"> \
+          <h2 class="govuk-heading-m">Someone recently signed in with '+ data['displayName'] +' on this device</h2> \
+          <div class="verify-hint-logos"> \
+            <img class="verify-hint-logos-idp" src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/'+ data['simpleId'] +'.png" alt=""> \
+            <img class="verify-hint-logos-verify" src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/govuk-verify-small-black-text-454fe97ff5e3edfb6eebdc648930c0ff675616ae7956f1d87e67b30f479d7b8d.svg" alt=""> \
+          </div> \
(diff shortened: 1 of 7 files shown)


config/initializers/csp.rb
```diff
@@ -1 +1,76 @@
...
+
+  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src
+  policy.img_src :self,
+                :data, # Base64 encoded images
+                *GOVUK_DOMAINS,
+                *GOOGLE_ANALYTICS_DOMAINS, # Tracking pixels
+                # Some content still links to an old domain we used to use
+                "assets.digital.cabinet-office.gov.uk",
+                # For a Verify enhanced hint live test (temporary)
+                "gds-verify-frontend-assets.s3.amazonaws.com"
```
issyl0: This feels very hacky. In an ideal world, I'd like for that extra domain to be configured in `govuk_app_config`. Would that be the plan for if the tests succeed goes live permanently, or getting them onto GOV.UK's asset mangement system somehow?

jakubmiarka: yes, it does (and it's not the only hacky thing :( ).. the `ENV['GOVUK_APP_DOMAIN_EXTERNAL']` could be used to pass it in, but that will also allow scripts and everything.. we only need images... If this was going to live, the CSP would have to be considered. At its current form it will get teared down...

issyl0: By "this" I meant "reproducing the whole of CSP for this one thing that deviates from the gem for everything in this app now". :-)

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
```diff
@@ -0,0 +1,52 @@
...
+
+    function renderHint (element, data) {
+      if (data != null && data['found'] == 'true') {
+        $(element).html(generateHtml(data)).show()
+        $('button:contains("Continue")').addClass('govuk-button--secondary')
+      }
+    }
+
+    function generateHtml(data) {
+      return '\
```
injms: I think this could be made more legible by concatenating strings, which is the way I've seen construction of markup done in other pieces of JavaScript on GOV.UK - for example:

```javascript
return '<div class="verify-hint-box">'
  + '<h2 class="govuk-heading-m">'
  + 'Someone recently signed in with '
  + data['displayName']
```

app/assets/stylesheets/views/_choose_sign_in.scss
```diff
@@ -0,0 +1,22 @@
+.verify-hint-box {
+  width: 100%;
+  border: 5px solid $govuk-border-colour;
+  padding: 15px;
```
injms: Whilst this is being torn down, so it's not too big a deal, spacing are generally done with [`govuk-spacing()` functions](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components).

For example, `15px` can be replaced with `govuk-spacing(3)` - which makes it easier to change in the future if the spacings are updated.

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
```diff
@@ -27,6 +27,7 @@
   <%= render "govuk_publishing_components/components/fieldset", legend_text: legend_text do %>
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
+      <div id="verify-hint" data-module="show-gov-uk-verify-hint" style="display:none"></div>
```
injms: `GOVUK.modules` should be triggered with a `data-module='moduleName'`, rather than triggering them by passing an element into the constructor. This has an added benefit of preventing the same module from running twice on a single element.

When [`GOVUK.modules.start`](https://github.com/alphagov/govuk_publishing_components/blob/b9b46987c8e1800fcf1f392dec404846cc7d171e/app/assets/javascripts/govuk_publishing_components/modules.js#L25) is called, it runs through the DOM looking for every element that has a `data-module` attribute. If the value of that attribute corresponds to a module that's available, it then initialises the module.

```suggestion
      <div id="verify-hint" data-module="show-gov-uk-verify-hint" style="display:none" data-module="show-gov-uk-verify-hint"></div>
```

jakubmiarka: that's what it's doing already?

jakubmiarka:
- I have updated the sass to use the govuk spacing function instead of fixed numbers
- changed the way the html string is concentanted so hopefully it's more readable
- @injms not sure about your blocking comment, since what you're describing is already happening and your suggested change duplicates attributes on the element

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
injms: It is. Apologies!

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
```diff
@@ -0,0 +1,52 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  var $ = global.jQuery
+
+  GOVUK.Modules.ShowGovUkVerifyHint = function () {
+    this.start = function (element) {
+      if (window.location.href.indexOf("/personal-tax-account/sign-in/prove-identity") > -1) {
```
alex-ju: Can we store this path in a variable at the top of the script to make it easier to spot/update?

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
```diff
@@ -0,0 +1,52 @@
...
+        checkLastSuccessfulIdp(element)
+      }
+    }
+    
+    this.render = function (element, data) {
```
alex-ju: Is this being called anywhere?

app/assets/stylesheets/views/_choose_sign_in.scss
```diff
@@ -0,0 +1,22 @@
...
+  box-sizing: border-box;
+  margin: -30px 0 30px;
```
alex-ju: As Ian mentioned, can we use the spacing scale at least for vertical spacing (e.g.  `- govuk-spacing(6) 0 govuk-spacing(6)`)

app/assets/stylesheets/views/_choose_sign_in.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  .verify-hint-logos {
+    margin-bottom: 15px;
+
+    .verify-hint-logos-idp {
```
alex-ju: We should use BEM for class names (e.g. `verify-hint-box`, `verify-hint-box__logos`, `verify-hint-box__logo-idp`, `.verify-hint-box__logo-verify`) which will make SCSS nesting unnecessary.

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
alex-ju: We should avoid in-line styling. Can we rely on CSS and `.js-enabled` class to hide this?
Also I don't see where we need an `id` for this element.

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
```diff
@@ -0,0 +1,66 @@
...
+        timeout: 3000
+      }).then(function(data){
+          renderHint(element, data);
+      }, function(e){console.log("error", e)})
+    }
+
+    function renderHint (element, data) {
+      if (data != null && data['found'] == 'true') {
+        $(element).html(generateHtml(data)).show()
+        $('button:contains("Continue")').addClass('govuk-button--secondary')
```
alex-ju: The `:contains` pseudoselector seems brittle to me since the copy is not controlled by this script – if content will be updated this line will fail to execute.

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+
+    function generateHtml(data) {
+      return '\
+        <div class="verify-hint-box"> \
+          <h2 class="govuk-heading-m">Someone recently signed in with '+ data['displayName'] +' on this device</h2> \
+          <div class="verify-hint-logos"> \
+            <img class="verify-hint-logos-idp" src="https://gds-verify-frontend-assets.s3.amazonaws.com/4af94ca-c1e26b4/'+ data['simpleId'] +'.png" alt=""> \
```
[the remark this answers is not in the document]
jakubmiarka: These are just logos, we felt they don't convey any information in this context, as the text itself is.

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
jakubmiarka: Yeah, there isn't an easier way and we don't expect the copy to change in the next 3 days anyway.

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
jakubmiarka: in tests

app/assets/stylesheets/views/_choose_sign_in.scss
```diff
@@ -0,0 +1,22 @@
+.verify-hint-box {
+  width: 100%;
```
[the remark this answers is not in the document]
jakubmiarka: well spotted, it originally wasn't a block level element. Will update.

app/views/content_items/service_sign_in/_choose_sign_in.html.erb
jakubmiarka: As much as I hate in-inline styling I have resorted to this as I don't want to make the element (even empty) visible. It's not only for people with no-js (so `.js-enabled` class does not help) but people who haven't used Verify successfully before shouldn't have get the element visible... 

The ID is a leftover from original implentation, it didn't get used. I'll remove it.

app/assets/javascripts/modules/show-gov-uk-verify-hint.js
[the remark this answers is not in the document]
jakubmiarka: I added it anyway in https://github.com/alphagov/government-frontend/pull/1585/commits/86655243dce49a78c3ee04a34bd5071d03c22b5c#diff-e4386657576fa7f7ba825873c292c692L48

app/assets/stylesheets/views/_choose_sign_in.scss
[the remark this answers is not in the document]
jakubmiarka: fixed

jakubmiarka:
@andysellick and @alex-ju thanks for having a look at the PR and your comments. I've addressed in the latest commit or responded to them. Could you have another look and review (and approve) since we only less than 4 days to run this test. Thanks!

PR #1638, opened 2020-01-28
Show no deal notice with no links

laurentqro:
Trello: https://trello.com/c/fBavxD33/426-no-deal-notice-should-display-if-no-links-provided

When the publisher in Whitehall ticks the checkbox for the notice to be displayed with no links, the `brexit_no_deal_notice` is present, with a value of `[]`.

We were expecting to see the notice in this case. However, the code was assuming we should only show the notice if links were present.

### DIFF
--- a/app/assets/stylesheets/components/_header-notice.scss
+++ b/app/assets/stylesheets/components/_header-notice.scss
@@ -18,6 +18,14 @@
   padding: govuk-spacing(4);
 }
 
+.app-c-header-notice__description {
+  margin-bottom: 0;
+}
+
+.app-c-header-notice__links {
+  margin-top: govuk-spacing(2);
+}
+
 .app-c-header-notice__list {
   margin-bottom: 0;
 }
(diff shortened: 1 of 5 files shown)


vanitabarrett:
@laurentqro I don't think the `header-notice` component checks for links or not. This might mean the `link_intro` is displayed even when there are no links, which wouldn't make sense. Can we add an example with link_intro and without links to [this file](https://github.com/alphagov/government-frontend/blob/master/app/views/components/docs/header-notice.yml), so we can test the component?

app/assets/stylesheets/components/_header-notice.scss
```diff
@@ -18,12 +18,21 @@
   padding: govuk-spacing(4);
 }
 
+.app-c-header-notice__text {
+  margin-bottom: govuk-spacing(2);
```
huwd: Don't need that new fancy fangled responsive mixin here?
:D

vanitabarrett: @huwd 10px doesn't get smaller on lower breakpoints and I think it would be too small if it did, so I think it's ok as is.

PR #1649, opened 2020-02-10
Remove web_isbn and print_meta_data_contact_address

JonathanHallam:
Trello - https://trello.com/c/w5dsqKuc/1381-remove-redundant-fields-for-whitehall-attachments-org-contact-and-web-isbn

These file types are redundant and should be removed. They are no longer being sent to the publishing api by whitehall

These fields are also being removed from govuk content schemas and whitehall
whitehall - https://github.com/alphagov/whitehall/pull/5333
content schemas - https://github.com/alphagov/govuk-content-schemas/pull/958

---

Visual regression results:

### DIFF
--- a/app/views/content_items/html_publication/_print_meta_data.html.erb
+++ b/app/views/content_items/html_publication/_print_meta_data.html.erb
@@ -13,21 +13,9 @@
 <p>
   This publication is available at <%= @content_item.full_path(request) %>
 </p>
-<% unless @content_item.print_meta_data_contact_address.blank? %>
-    <p>
-      Any enquiries regarding this publication should be sent to us at:<br/>
-    </p>
-    <%= simple_format(@content_item.print_meta_data_contact_address) %>
-<% end %>
 
 <% unless @content_item.isbn.blank? %>
     <p>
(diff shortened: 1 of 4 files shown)


app/views/content_items/html_publication/_print_meta_data.html.erb
```diff
@@ -13,21 +13,9 @@
...
-<% unless @content_item.print_meta_data_contact_address.blank? %>
-    <p>
-      Any enquiries regarding this publication should be sent to us at:<br/>
-    </p>
-    <%= simple_format(@content_item.print_meta_data_contact_address) %>
-<% end %>
 
 <% unless @content_item.isbn.blank? %>
     <p>
       Print ISBN: <%= @content_item.isbn %>
```
benthorner: It would make sense to remove `Print` here, although that would be a user-facing change. Perhaps raise in a separate PR, as it's confusing to see it used without context of knowing there used to be another ISBN.

JonathanHallam: Yeah, good point, I'll raise a separate one after this :)

PR #1655, opened 2020-02-17
Override old "documents" fields with new "attachments" fields

barrucadu:
If both are set and the attachments field is the same size as the
documents field.  If they're not the same size an error is logged.  We
can use this to be confident that, after republishing everything,
we've not somehow lost any data.

When we're happy to remove the old field, we can revert this commit
and just use the new field directly.

---

[Trello card](https://trello.com/c/iOfOQesh/1389-make-frontend-apps-render-featuredattachments)

Visual regression results:
https://government-f-msw-rfc-11-dk2x9g.surge.sh/gallery.html

### DIFF
--- a/app/presenters/content_item/featured_attachments_migration.rb
+++ b/app/presenters/content_item/featured_attachments_migration.rb
@@ -1,0 +1,23 @@
+module ContentItem
+  module FeaturedAttachmentsMigration
+    def choose_field(new_field_name:, old_field_name:)
+      new_list = content_item["details"][new_field_name]
+      old_list = content_item["details"][old_field_name] || []
+
+      # don't raise an error just because a document hasn't been
+      # republished to have the new field yet.
+      return old_list if new_list.nil?
+
+      if new_list.length == old_list.length
+        new_list
+      else
+        GovukError.notify(
+          "Mismatch between attachments and documents",
+          extra: { error_message: "Document with #{new_list.length} #{new_field_name} but #{old_list.length} #{old_field_name} at #{base_path}" },
+        )
+
+        old_list
+      end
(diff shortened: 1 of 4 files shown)


test/presenters/content_item/featured_attachments_migration.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class FeaturedAttachmentsMigrationTest < ActiveSupport::TestCase
+  class DummyContentItem
+    include ContentItem::FeaturedAttachmentsMigration
+    attr_accessor :content_item
+
+    def initialize(old_list, new_list)
+      @content_item = {
+        "base_path" => "/a/base/path",
+        "details" => { "old_badness" => old_list, "new_hotness" => new_list }.compact,
```
bilbof: Love those key names 😆

app/presenters/content_item/featured_attachments_migration.rb
```diff
@@ -0,0 +1,22 @@
+module ContentItem
+  module FeaturedAttachmentsMigration
+    def choose_field(new_field_name:, old_field_name:)
+      new_list = content_item["details"].fetch(new_field_name, [])
+      old_list = content_item["details"].fetch(old_field_name, [])
+
+      if new_list != []
```
bilbof: You could do `return old_list unless new_list.any?` here, then you don't need to nest the rest of this in the if block.

barrucadu:
Got an error on the heroku preview app when I tried to preview a random consultation, just checking that out now

barrucadu:
Looks unrelated, and the master branch preview for consultations is broken too

PR #1686, opened 2020-03-11
Header notice spacing tweaks

vanitabarrett:
## What
- Adjust padding on black header to take border into account
- Remove bottom margin on last paragraph inside the notice

## Before
Screenshot 2020-03-11 at 11 50 48
Screenshot 2020-03-11 at 11 50 40

## After
Screenshot 2020-03-11 at 11 54 45
Screenshot 2020-03-11 at 11 55 00


vanitabarrett:
@alex-ju That makes sense! This change isn't urgent so will amend as part of this PR 🙂

PR #1698, opened 2020-03-19
Fix the appearance of the dev landing page

andysellick:
## What / why

The dev landing page was looking a bit broken. It's always looked a little unstyled, but with the recent change to the components gem it was looking worse (the search box was unstyled, footer layout was broken). This fixes that.

## Visual changes

Before:

Screenshot 2020-03-19 at 16 30 10

After:

Screenshot 2020-03-19 at 16 29 53

### DIFF
--- a/app/views/development/index.html.erb
+++ b/app/views/development/index.html.erb
@@ -5,6 +5,7 @@
     government-frontend development page
   </title>
   <meta name="robots" content="noindex, nofollow">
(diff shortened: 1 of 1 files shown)


alex-ju:
Nice one, @andysellick! 👏

andysellick:
Thanks @alex-ju, although I'm pretty sure I broke it in the first place... 😁

PR #1709, opened 2020-03-25
Render featured attachments directly

benthorner:
https://trello.com/c/wM9wqe9H/1587-support-featured-attachments-directly-on-govuk-frontend

Depends on: https://github.com/alphagov/govuk-content-schemas/pull/975

This changes government-frontend to render attachments 
directly, when a pre-rendered payload is not supplied (see RFC 116) [1].

Note that we no longer support PDF preview for documents rendered 
using this new method. This won't impact any documents at the moment, 
since all of them still have a pre-rendered 'documents' payload.

## Before (Publication)
screencapture-government-frontend-dev-gov-uk-government-publications-number-of-ex-regular-service-personnel-in-fr20-2020-03-25-13_39_45

## After (Publication)
screencapture-government-frontend-dev-gov-uk-government-publications-number-of-ex-regular-service-personnel-in-fr20-2020-03-25-13_40_05

## Before (Consultation)
screencapture-government-frontend-dev-gov-uk-government-consultations-setting-the-grade-standards-of-new-gcses-in-england-2017-2018-2020-03-25-13_38_36

## After (Consultation)
screencapture-government-frontend-dev-gov-uk-government-consultations-setting-the-grade-standards-of-new-gcses-in-england-2017-2018-2020-03-25-13_38_52

## Before (Mobile)
Screenshot 2020-03-25 at 12 19 19

## After (Mobile)
Screenshot 2020-03-25 at 12 19 15

[1]
https://github.com/alphagov/govuk-rfcs/blob/master/rfc-116-store-attachment-data-in-content-items.md


app/views/content_items/publication.html.erb
```diff
@@ -29,7 +29,27 @@
   <div class="govuk-grid-column-two-thirds responsive-bottom-margin">
     <%= render 'components/important-metadata', items: @content_item.important_metadata %>
     <div class="responsive-bottom-margin">
-       <%= render 'publication_inline_body' %>
+      <%= render 'govuk_publishing_components/components/heading',
+        text: t("publication.documents", count: 5), # This should always be pluralised.
```
ChrisBAshton: Maybe in another commit after the first commit, you could change this. The `count: 5` feels weirdly arbitrary. Even `2` would be clearer!

app/views/content_items/_attachments.html.erb
```diff
@@ -0,0 +1,14 @@
+<% attachment_title_id = "#{title.parameterize}-title" %>
+
+<% if legacy_pre_rendered_documents.present? %>
+  <%= render 'govuk_publishing_components/components/heading',
+      id: attachment_title_id,
+      text: title,
+      mobile_top_margin: true %>
+
+  <div aria-labelledby="<%= attachment_title_id %>">
```
ChrisBAshton: Wouldn't this have the effect of reading the heading out twice (once on the heading itself, and once immediately afterwards)? (NB, the examples I've seen on MDN tend to show the heading inside the div):

```erb
<div aria-labelledby="<%= attachment_title_id %>">

  <%= render 'govuk_publishing_components/components/heading',
      id: attachment_title_id,
      text: title,
      mobile_top_margin: true %>
```

I'm sure Alex has already looked at this, I'm just not immediately clear on how this is an improvement 🤔

test/integration/consultation_test.rb
```diff
@@ -87,21 +95,36 @@ class ConsultationTest < ActionDispatch::IntegrationTest
...
-  test "consultation outcome documents render" do
+  test "renders consultation outcome attachments (as-is and directly)" do
     setup_and_visit_content_item("consultation_outcome")
 
+    assert page.has_text?("Download the full outcome")
     within '[aria-labelledby="download-the-full-outcome-title"]' do
       assert page.has_text?("Employee Share Schemes: NIC elections - consulation response")
     end
+
+    setup_and_visit_content_item("consultation_outcome_with_featured_attachments")
```
ChrisBAshton: Each of these `test` blocks seem like they should be split into two `test` blocks - one for `as-is`, and one for `directly`. I appreciate there's no other way of grouping them together, save for a comment in between every two blocks, but it would make it easier to delete the `as-is` blocks when we come to deprecate that behaviour.

benthorner: While I agree they could be split up, I think that would decrease the readability of the rest of the test file, as there is indeed no way of grouping them, and they are effectively testing the same thing. One of the commit messages says this, somewhere.

benthorner: I don't think I would find it that much harder to delete the lines within the test, when we (finally) disable the legacy functionality.

app/views/content_items/_attachments.html.erb
benthorner: Looks like this is a case of me assuming we were doing a good thing in the first place 🤦‍♂. This is replicating a behaviour that [previously applied only to publication documents](https://github.com/alphagov/government-frontend/pull/1709/commits/6287f6847acf2fb6ef9024fd3fce5042186bedb8#diff-24095b2b1c8348342f68a199663a50b7R37), and applying it to all lists of attachments (with a title).

Apparently it was introduced [here](https://github.com/alphagov/government-frontend/pull/256/commits/da1bc39eaf8b3839af79ea5f4f4adfa910f60203), but with no explanation. I've added another commit so it conforms [to the spec](https://www.w3.org/TR/wai-aria/#region). Removing it makes the tests harder, as there's nothing to 'grab hold of', so I've gone with fixing it, instead.

How does it look now?

app/views/content_items/publication.html.erb
benthorner: It's tempting, but that means going through all 61 locale files and working out what to do with it. There's definitely a bigger problem with the use of I18n in this app, so I'd like to leave it for someone else to tackle.

benthorner:
Ah, that's useful to know, thanks. I've updated it to use `section`. Shame it doesn't actually do anything exciting with my screenreader - I assume others will benefit 🙏.

benthorner:
Ah, thanks for pointing that out - I like learning a new thing, even if it's dubiously documented 👓. I've updated the last commit to implement that, which has the nice effect of simplifying the tests 👍.

PR #1714, opened 2020-03-27
[DO NOT MERGE] Move webchat section to be above the email section

thomasleese:
There was some confusion in the last change (5b00d3e40ec26778a918be73aeb58f99f82c585f) and it looks like they would prefer the webchat section to appear above the email section (rather than above the phone section).

[Trello Card](https://trello.com/c/L5LSGY8T/1879-move-webchat-section-on-hmrc-contact-pages)


thomasleese:
I won't merge this until we have confirmation this is what they want.

PR #1737, opened 2020-04-16
Replace placeholder image

MuriloDalRi:
This changes the Whitehall street sign as request here: https://govuk.zendesk.com/agent/tickets/3992513


MuriloDalRi:
> The ticket is only to change the image for world location news articles, not all news articles. World location news articles are https://github.com/alphagov/government-frontend/blob/master/app/presenters/world_location_news_article_presenter.rb

🤦 That seemed too easy

PR #1740, opened 2020-04-21
Fix beta banner

laurentqro:
Rendering the govuk component in the view helper breaks the styling
because govuk-publishing-components gem relies on said components
being rendered from inside a view to detect which component is being
rendered, and thus, which stylesheets to require.

## Before

Screenshot 2020-04-21 at 17 23 26

## After

Screenshot 2020-04-22 at 09 31 29

https://trello.com/c/ZITqImYk/109-adding-alpha-beta-banner-to-start-page-to-indicate-this-is-a-trial


app/views/layouts/application.html.erb
```diff
@@ -35,7 +35,9 @@
   <% end %>
 
   <div id="wrapper" class="<%= wrapper_class %>">
-    <%= render_phase_label @content_item, content_for(:phase_message) %>
```
laurentqro, line 38: `:phase_message` was unused.

vanitabarrett:
As part of this PR, we'll need to pull in the CSS for the component like we do for others here: https://github.com/alphagov/government-frontend/blob/master/app/assets/stylesheets/application.scss#L6

PR #1744, opened 2020-04-23
Add GA params to /ask survey link

chao-xian:
We need to be able to compare the number of people who go on to see the survey vs the people who complete so that we can work out the rate.

Trello https://trello.com/c/La96nEZw

---

Visual regression results:
https://government-frontend-pr-1744.surge.sh/gallery.html


app/assets/javascripts/set-ga-client-id-on-url-in-element.js
```diff
@@ -0,0 +1,23 @@
...
+    if (!options.$linkedElement || !options.attribute || !window.ga) {
+      return
+    }
+
+    var linkedElement = options.$linkedElement
+    var attribute = options.attribute
+
+    window.ga(function (tracker) {
+      var clientId = tracker.get('clientId')
+      var href = linkedElement.attr(attribute)
```
vanitabarrett, line 17: Is it confusing to call this `href` when it depends on the attribute itself (so might not always be a href)?

chao-xian: Good point - [fixed](https://github.com/alphagov/government-frontend/pull/1744/commits/2286ccabb85d83bfaaf5e5cefabc4f2873f7832e)

PR #1812, opened 2020-07-03
Update global travel advice callout to reflect 4th July updates

issyl0:
- As per request from FCO.

### DIFF
--- a/app/views/content_items/travel_advice.html.erb
+++ b/app/views/content_items/travel_advice.html.erb
@@ -26,7 +26,7 @@
       </div>
       <div class="travel-advice-notice__content">
         <p class="govuk-body">
-          <strong>The Foreign &amp; Commonwealth Office currently advises British nationals against all but essential international travel.</strong> This advice is being kept under constant review.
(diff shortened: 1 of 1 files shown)


app/views/content_items/travel_advice.html.erb
```diff
@@ -26,7 +26,7 @@
       </div>
       <div class="travel-advice-notice__content">
         <p class="govuk-body">
-          <strong>The Foreign &amp; Commonwealth Office currently advises British nationals against all but essential international travel.</strong> This advice is being kept under constant review.
+          <strong>The Foreign &amp; Commonwealth Office currently advises British nationals against all but essential international travel.</strong> This advice is being kept under constant review. From 4 July <a href="https://www.gov.uk/guidance/coronavirus-covid-19-countries-and-territories-exempt-from-advice-against-all-but-essential-international-travel">travel to some countries and territories will be exempt</a>.
```
andysellick, line 29: Bit late, but do these URLs need to be absolute?

PR #1851, opened 2020-09-08
Add additional skip link on guide views

owenatgov:
## What
Adds an additional skip link above the guide contents that goes to the actual content of the guide.

## Why
This was identified as a WCAG fail (specifically WCAG SC 2.4.1) as for keyboard only users, it meant that they had the option to skip navigation in the govuk header but still have to tab through the navigation of the guide contents to get to the "actual" content of the guide. This solution of using an additional skip link is used on govuk search pages to skip past filters.

**Card:** https://trello.com/c/x3QKWP0z/340-cannot-bypass-repeated-navigation

### DIFF
--- a/app/views/content_items/guide.html.erb
+++ b/app/views/content_items/guide.html.erb
@@ -19,13 +19,17 @@
     <%= render 'govuk_publishing_components/components/title', { title: @content_item.content_title } %>
 
     <% if @content_item.show_guide_navigation? %>
+      <%= render "govuk_publishing_components/components/skip_link", {
+        text: "Skip to contents of guide",
+        href: "#guide-contents"
+      } %>
       <aside class="part-navigation-container" role="complementary">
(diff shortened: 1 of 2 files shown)


app/views/content_items/guide.html.erb
```diff
@@ -19,13 +19,17 @@
     <%= render 'govuk_publishing_components/components/title', { title: @content_item.content_title } %>
 
     <% if @content_item.show_guide_navigation? %>
+      <%= render "govuk_publishing_components/components/skip_link", {
+        text: "Skip to contents of guide",
+        href: "#guide-contents"
+      } %>
       <aside class="part-navigation-container" role="complementary">
```
injms, line 26: I looks like the `aside` overlaps slightly with the focussed skip link:

image

injms (approved):
Small visual discrepancy that a bit of margin will sort out - but nothing that stops this from being a solid improvement over not having a skip link at all.

PR #1860, opened 2020-09-16
Remove ordered list markup for organisation logos

owenatgov:

## What
Move logos at the top of organisation pages out of the ordered list element.

## Why
These logos should not be in an ordered list as there's nothing to denote them as ordered. This could've been changed into an unordered list but this itself isn't ideal as they aren't really intended as a list. [Example of a page with multiple logos](https://www.gov.uk/government/publications/face-coverings-when-to-wear-one-and-how-to-make-your-own/face-coverings-when-to-wear-one-and-how-to-make-your-own).

No visual changes.

**Card:** https://trello.com/c/zeNDQ9hA/374-ordered-list-used-when-list-is-not-ordered

### DIFF
--- a/app/assets/stylesheets/helpers/_organisation-logos.scss
+++ b/app/assets/stylesheets/helpers/_organisation-logos.scss
@@ -9,11 +9,7 @@
 
 .organisation-logos__logo {
   padding-bottom: govuk-spacing(3);
+  margin-right: govuk-spacing(3);
   flex-basis: 25%;
   min-width: 130px;
-}
-
-.organisation-logo__inner {
-  max-width: 215px;
-  padding-right: govuk-spacing(3);
 }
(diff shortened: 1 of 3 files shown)


maxgds:
Can you link to some example pages? Particularly interesting would be ones with more than one logo.

danacotoran:
@owenatgov this may very well be a daft question but why do you say they are they not intended to be a list? 

Isn't the purpose of that markup to list the organisations that own that particular document/publication/page/whathaveyou?
I can get behind the argument that it does not particularly make sense to have them as an `<ol>` as there's nothing ordered about it, but I'm not sure I understand why it shouldn't be a `<ul>` 🤔

owenatgov:
@danacotoran So in short, I don't really have an explanation. I was on the fence about if it should be a `ul` or not. When I took the `ol` out I also thought it made sense to de-list it altogether because when I observed this it felt like improper use of list semantics because it wasn't a traditional list aka: items "listed" one after the other. However, I also wondered if this perspective came from a sighted bias and, removing the visual context, this is perfectly acceptable as a list. I feel like your reasoning makes perfect sense, I'll swap this out for a `ul` instead.

danacotoran (approved):
👍🏼

PR #1877, opened 2020-10-14
Replace bespoke nav with contents list component

maxgds:
This removes the bespoke 2 column layout for parts navigation and replaces it with the contents-list component. This brings in a header of "Contents" and makes the navigation more consistent with the rest of the site which is good for accessibility. 
https://trello.com/c/LbVmpaK6/441-inconsistent-naming-of-table-of-contents
Sample page: https://www.gov.uk/foreign-travel-advice/switzerland

### Before
Screenshot 2020-10-14 at 15 03 41

### After
Screenshot 2020-10-14 at 15 03 31

### DIFF
--- a/app/assets/stylesheets/helpers/_parts.scss
+++ b/app/assets/stylesheets/helpers/_parts.scss
@@ -14,16 +14,10 @@
   }
 
   .part-navigation {
-    margin-bottom: 0;
-    @include govuk-font(19);
+    margin-left: 0;
 
-    @include govuk-media-query($from: tablet) {
-      @include govuk-font(16);
-    }
-
-    li {
-      list-style: none;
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
💪

owenatgov:
Non-blocking comment that I noticed post approval: The contents section is now quite snug to the left side on mobile.
Screenshot 2020-10-15 at 16 40 11

maxgds:
> Non-blocking comment that I noticed post approval: The contents section is now quite snug to the left side on mobile.

Thanks, I realised it was too far over at all widths, actually. Have amended, could you re-check, please?

injms (approved):
👍 🚢

PR #1939, opened 2020-12-11
Render Get Involved page on Government Frontend

Tetrino:
**DRAFT PR DO NOT MERGE**

The summation of a bunch of work to migrate the Get Involved static page from Whitehall to Government Frontend. This PR is designed so people can look and comment on any outstanding concerns that may be present.

This took more time than it really should of, in part due to modernising elements in Whitehall that we do not support in Government Frontend, and in part due to other unrelated concerns.

Testing needs to be completed on Integration before a merge can go ahead. It is likely minor CSS changes will have to be brought in, and possibly a modification to some of the publishing-API calls for accuracy and/or load limit.


app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+        org_acronyms << Services.publishing_api.get_live_content(org_id).parsed_content['details']['acronym']
+      end
+      consultation['links']['organisation_acronyms'] = org_acronyms
+    end
+  end
+
+  def sort_take_part(take_part_pages)
+    take_part_pages.sort_by{ |page| page["details"]["ordering"] }
+  end
+end
```
ChrisBAshton: Missing a newline here.

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
+class GetInvolvedController < ApplicationController
```
ChrisBAshton, line 1: Might be useful to link to this in the commit message, for posterity: https://github.com/alphagov/whitehall/blob/bcda05ec43d87b3bec1541e644c1373bafa2a3e8/app/controllers/home_controller.rb#L17-L25

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    @do_not_show_breadcrumbs = true
+    render template: "content_items/get_involved"
+  end
+
+  def load_content_item
+    content_item = Services.content_store.content_item("/government/get-involved")
+
+    if Services.feature_toggler.use_recommended_related_links?(content_item["links"], request.headers)
+      content_item["links"]["ordered_related_items"] = content_item["links"].fetch("suggested_ordered_related_items", [])
+    end
```
ChrisBAshton, line 16: This feels like it needs a test.

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    @content_item = PresenterBuilder.new(
+      content_item,
+      content_item_path,
+      view_context,
+    ).presenter
+  end
+
+  def load_get_involved_data
+    @open_consultation_count = retrieve_given_document_type("open_consultation")['total']
```
ChrisBAshton: I suspect the mixture of single and double quotes is upsetting the linter here (and elsewhere in this file).

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    Services.publishing_api.get_content_items(document_type: document_type).to_hash
+  end
+
+  def retrieve_special(params)
+    Services.publishing_api.get_content_items(params).to_hash
+  end
+
+  # Aims to find a count of consultations closed in the last n months.
+  # Could feasibly be done via SearchAPI calls but there is no direct or indirect way to search by closing date
+  # there so we have to go around the houses a bit via a single large P-API call.
```
ChrisBAshton, line 44: It tripped me up for a second, and there's no reason why it needs to :) 

```suggestion
  # there so we have to go around the houses a bit via a single large Publishing API call.
```

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  def retrieve_given_document_type(document_type)
+    Services.publishing_api.get_content_items(document_type: document_type).to_hash
+  end
+
+  def retrieve_special(params)
+    Services.publishing_api.get_content_items(params).to_hash
+  end
+
+  # Aims to find a count of consultations closed in the last n months.
+  # Could feasibly be done via SearchAPI calls but there is no direct or indirect way to search by closing date
```
ChrisBAshton, line 43:
```suggestion
  # Could feasibly be done via Search API calls but there is no direct or indirect way to search by closing date
```

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    cl_docs.each do |doc|
+      if DateTime.parse(doc['details']['closing_date']) > DateTime.now.prev_month(months)
+        closed_count += 1
+      else
+        break
+      end
+    end
+
+    return closed_count # hard return for clarity
```
ChrisBAshton: I think Ruby/Rails developers reading this would be fine with implicit return - not sure why this method should be treated differently to any other.

```suggestion
   closed_count
```

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def retrieve_next_closing
+    open_consults = retrieve_given_document_type("open_consultation")['results']
+    open_consults.sort_by! { |k| k['details']['closing_date'] }[0]
+  end
+
+  def retrieve_new_consultations
```
ChrisBAshton, line 68: Given that this method and the one above it have the same call to `retrieve_given_document_type`, I'd suggest passing that in as an argument instead. It would mean defining what the 'open consults' are in only one place (load_get_involved_data) and make it easier to test, too.

```suggestion
  def retrieve_new_consultations(open_consults)
```

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      organisations = consultation['links']['organisations']
+      organisations.each do |org_id|
+        org_acronyms << Services.publishing_api.get_live_content(org_id).parsed_content['details']['acronym']
+      end
+      consultation['links']['organisation_acronyms'] = org_acronyms
+    end
+  end
+
+  def sort_take_part(take_part_pages)
+    take_part_pages.sort_by{ |page| page["details"]["ordering"] }
```
ChrisBAshton: As this only saves adding a `.sort_by` call in `load_get_involved_data`, I question the value in having it in its own function (unless you're writing a test for it, but even then there's an argument that it should be declared below a `private` keyword, thus have no 'testable' advantage to being in its own method). I think I'd put it inline instead.

app/controllers/get_involved_controller.rb
```diff
@@ -46,6 +46,7 @@ def retrieve_special(params)
   def retrieve_date_filtered_closed_consultations(months)
     closed_count = 0
 
+    # Has to be a better way of doing this, will fail to count more than 500 results
```
ChrisBAshton, line 48: What happens if you remove the `per_page: 500`? Or set it to some extremely large number?

app/helpers/get_involved_helper.rb
```diff
@@ -0,0 +1,31 @@
+module GetInvolvedHelper
+  def date_microformat(attribute_name)
+      attribute_name.to_date.strftime("%d %B %Y")
```
ChrisBAshton: Too much indentation here.

app/helpers/get_involved_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  # Gets the link to the search page for all consultations
+  def get_consultations_link(filters = %w[open_consultations closed_consultations])
+    "search/policy-papers-and-consultations?#{filters.to_query("content_store_document_type")}"
+  end
+
+  private
+
+  def render_datetime_microformat(object, method, &block)
```
ChrisBAshton, line 13: Is this actually used anywhere? 🤔

app/helpers/get_involved_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+    when :zero?.to_proc
+      "Closing today"
+    when 1
+      "Closes tomorrow"
+    else
+      "#{days_left} days left"
+    end
+  end
+
+end
```
ChrisBAshton: Missing newline.

app/helpers/get_involved_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+  def render_datetime_microformat(object, method, &block)
+    tag.time(class: method, datetime: object.send(method).iso8601, &block)
+  end
+
+  def time_until_closure(consultation)
+    days_left = (consultation['details']['closing_date'].to_date - Time.zone.now.to_date).to_i
+    case days_left
+    when :negative?.to_proc
+      "Closed"
+    when :zero?.to_proc
```
ChrisBAshton, line 22: These are clever! 👍

app/helpers/application_helper.rb
```diff
@@ -28,6 +28,29 @@ def current_path_without_query_string
...
+    # rubocop:disable Rails/HelperInstanceVariable
+    if title_parts.any?
+      title_parts.push("Admin") if params[:controller].match?(/^admin\//)
+      title_parts.push("GOV.UK")
+      @page_title = title_parts.reject(&:blank?).join(" - ")
+    else
+      @page_title
+    end
+    # rubocop:enable Rails/HelperInstanceVariable
+  end
```
ChrisBAshton, line 48: I'm looking at commit f7434f239bf5b3e66a34286e65a4642e3cecbcaa - it seems everything above this line isn't actually used in this commit (or at least not explicitly). Could it be moved to its own commit where it can be read and understood in isolation?

app/views/layouts/application.html.erb
```diff
@@ -47,8 +47,10 @@
       <% end %>
     <% end %>
 
-    <main role="main" id="content" class="<%= @content_item.schema_name.dasherize %>" lang="<%= I18n.locale %>">
-      <%= yield %>
+    <main role="main" id="content" class="government-frontend-content" lang="<%= I18n.locale %>">
+      <div id="page" class="<%= content_for?(:page_class) ? yield(:page_class) : '' %>" >
```
ChrisBAshton, line 51: I don't see any styling associated with `#page` or `.history` - what effect does this change have?

app/assets/stylesheets/views/_get-involved.scss
```diff
@@ -1,22 +1,42 @@
...
+  .govuk-grid-column-one-third-from-desktop{
+    padding: 0px;
+  }
+
+  .govuk-grid-column-two-thirds-from-desktop{
+    padding: 0px;
+  }
+
+  .govuk-grid-column-one-half-from-desktop{
+      padding: 0px;
```
ChrisBAshton: All of these can be changed to `padding: 0;` (omit the `px` when it's a zero value).

app/assets/stylesheets/views/_get-involved.scss
```diff
@@ -137,40 +156,40 @@
...
-            margin: $gutter-half $gutter-half $gutter;
+          @include govuk-media-query($from: tablet) {
+            margin: govuk-spacing(3) govuk-spacing(3) govuk-spacing(6);
           }
         }
-
         .column {
-          @include media(tablet) {
+          @include govuk-media-query($from: tablet) {
+            /*@include govuk-grid-column-one-half;*/
```
ChrisBAshton: Can remove this comment.

app/views/content_items/get_involved.html.erb
```diff
@@ -94,31 +91,32 @@
...
+      <div class="column govuk-grid-column-two-thirds-from-desktop">
         <div class="content">
           <p>For an instant way to interact with government departments, try their social media streams. These are listed under 'Follow us' on the department's home page. As well as access to blogs, audio, video and more, you can comment, debate and rate.</p>
           <p class="see-all"><a href="/government/organisations/" class="govuk-link">See all government departments</a></p>
           <p>Some active government blogs include:</p>
           <ul class="index-list">
             <li><a href="https://history.blog.gov.uk" class="govuk-link" rel="external">History of Government</a></li>
             <li><a href="https://quarterly.blog.gov.uk" class="govuk-link" rel="external">Civil Service Quarterly</a></li>
-            <li><a href="https://blogs.fcdo.gov.uk/" class="govuk-link" rel="external">FCDO bloggers</a></li>
+            <li><a href="http://blogs.fco.gov.uk" class="govuk-link" rel="external">FCO bloggers</a></li>
```
ChrisBAshton, line 116: This - and the change below - has slipped into the wrong commit.

app/controllers/content_items_controller.rb
```diff
@@ -63,7 +63,6 @@ def show_history_page
 
     if valid_page_ids.include?(page_id)
       @do_not_show_breadcrumbs = true
-
```
ChrisBAshton, line 66: This should be fixed up to the commit where it was added.

app/assets/stylesheets/application.scss
```diff
@@ -85,7 +86,10 @@ $govuk-use-legacy-palette: false;
 @import 'views/specialist-document';
 @import 'views/answer';
 @import 'views/help-page';
-@import "views/guide";
+@import 'views/guide';
+@import 'views/get-involved';
+@import 'views/headings';
+@import 'helpers/featured_news';
```
ChrisBAshton, line 92: This seems quite a strange change to include with the route addition. I guess adding the route caused the asset compilation to fail on these imports, hence why they were added. But I think it would be better to add these in earlier on, when the associated CSS was added.

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+    sorted_desc = closed_consults.sort_by! { |k| k['details']['closing_date'] }.reverse!.values_at(0..2)
+    parse_organisation_acronyms(sorted_desc)
+  end
+
+  def parse_organisation_acronyms(consultations)
+    consultations.each do |consultation|
+      org_acronyms = []
+      organisations = consultation['links']['organisations']
+      organisations.each do |org_id|
+        org_acronyms << Services.publishing_api.get_live_content(org_id).parsed_content['details']['acronym']
```
ChrisBAshton: A little concerned about the nested loops and an API call in the middle of that nest - how many requests does it end up making? But I can't really see an alternative either.

ChrisBAshton (commented):
First of all, congrats on getting through this very chunky piece of work - I know it wasn't easy!

I have no major concerns about the overall approach in the code, but it does seem to be missing quite a few tests which ought to be added before merging. The commit history is also quite confusing to follow. I've made some suggestions below, hopefully it's not _too_ painful to rebase!

app/controllers/get_involved_controller.rb
Tetrino: You'd think, but this is like this _because_ of the linter :P

app/controllers/get_involved_controller.rb
Tetrino: The default value is 20 results per page when retrieving from the publishing api. This is all handled that end but it handles large numbers sensibly. If I remove it we don't get anywhere near the correct values because we don't have them all to count. The alternative is a bunch of looping logic doing it page by page with multiple calls to the API.

My concern is just how large a packet is being requested here. I could - and should - refine it into the loop for smaller calls, but if anything happens to the API calls in the loop who knows what will happen?

app/helpers/get_involved_helper.rb
Tetrino: This is a remnant from the change in Line 3. Previously it did all of this extra to format the time but I optimised it in Line 3 and didn't actually remove the original function. Good catch!

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+  def retrieve_given_document_type(document_type)
+    Services.publishing_api.get_content_items(document_type: document_type).to_hash
+  end
+
+  def retrieve_special(params)
+    Services.publishing_api.get_content_items(params).to_hash
+  end
+
+  # Aims to find a count of consultations closed in the last n months.
+  # Could feasibly be done via SearchAPI calls but there is no direct or indirect way to search by closing date
```
thomasleese, line 43: This does concern me a little bit. It's quite unusual for a frontend application to need to talk to the Publishing API (except [for publishing its routes which itself is considered tech debt](https://trello.com/c/hJVUjqbs/197-frontend-apps-publish-content-items-to-publishing-api)) and it definitely feels like this could become a source of confusion and potential issues if those pages becomes highly trafficked (the Publishing API isn't always the quickest thing to respond to requests asking for content).

I wonder if we can dig a little more in the reasons why we can't get this information from the Search API and whether it's feasible to make changes there?

Sorry this is a bit of a spanner in the works! But I think it's good to make sure we get this right.

Tetrino: I agree. I'll clean up git first then investigate.

app/controllers/get_involved_controller.rb
Tetrino: Believe me if there was a better way... Might be able to do it with SearchAPI now that I'm in that life, but it'd still likely be a loop unless I have a bunch of OR tags in the search (which might be possible but I don't know until I look at it).

app/views/content_items/get_involved.html.erb
Tetrino: Only real way of solving this one is to include it in the original import of the html.erb as the change happened mid-development.

app/controllers/get_involved_controller.rb
thomasleese: This kind of thing should be done via [link expansion](https://github.com/alphagov/publishing-api/blob/master/docs/link-expansion.md). You can add `acronym` to the list of fields [that get included in the representation of the organisation in the `links`](https://github.com/alphagov/publishing-api/blob/63a973916fcca6062dbfe61c51cdb60b4c1526c0/lib/expansion_rules.rb#L82) of the consultation. Then no API calls needs to be made, and the content is already there.

Although now that the data itself is coming from the Search API this might be harder, perhaps a better solution would be to get Search API to expose that in the search results. I'm not super familiar with the Search API, but perhaps something like this: https://github.com/alphagov/search-api/blob/35ce835ddde96c5a7cbdab61f7beeda9daf3b03f/lib/govuk_index/presenters/expanded_links_presenter.rb#L80-L83

app/assets/stylesheets/views/_get-involved.scss
Tetrino: oops

app/controllers/get_involved_controller.rb
Tetrino: @thomasleese now we're in the search API it's actually more viable as said linking is already done! - See https://www.gov.uk/api/search.json?filter_format=consultation (literally any result)

I posted my comment this morning before I gave a solid look at it, but my internet is frankly so crap that it took this long to post.

app/views/layouts/application.html.erb
Tetrino: No longer needed.

app/controllers/get_involved_controller.rb
Tetrino: Resolving as this code is no longer present.

Tetrino:
Be aware that GemFile.lock and the publishing components commits are not part of this work, and are there as an artefact of rebasing I can't seem to clear. They will not overwrite the work already done on main.

app/controllers/get_involved_controller.rb
```diff
@@ -23,9 +23,9 @@ def load_content_item
   end
 
   def load_get_involved_data
-    @open_consultation_count = retrieve_given_document_type("open_consultation")['total']
+    @open_consultation_count = Services.search_api.search({ filter_display_type: "Open consultation", count: 0 })["total"]
```
brucebolt, line 26: Could this be filtered on a static key rather than text that could change?

I think [`filter_content_store_document_type: "open_consultation"`](https://www.gov.uk/api/search.json?filter_content_store_document_type=open_consultation) should suffice.

Same further down each time this query is being made.

app/controllers/get_involved_controller.rb
```diff
@@ -37,48 +37,56 @@ def retrieve_given_document_type(document_type)
 
   # Aims to find a count of consultations closed in the last n months.
   def retrieve_date_filtered_closed_consultations(months)
-    closed_count = 0
+    cutoff_date = Time.zone.now.prev_month(months)
 
-    cl_docs = retrieve_special({document_type: 'closed_consultation', per_page: 500, fields: ['details']})['results']
-    cl_docs.sort_by! { |k| -k['details']['closing_date'] }.reverse!
+    query = {
+      filter_display_type: "Closed consultation",
```
brucebolt, line 43: Same as above, `filter_content_store_document_type: "closed_consultation"` would be better here.

app/lib/services.rb
```diff
@@ -24,4 +24,11 @@ def self.publishing_api
       bearer_token: ENV.fetch("PUBLISHING_API_BEARER_TOKEN", "example"),
     )
   end
+
+  def self.search_api
+    @search_api = GdsApi::Search.new(
+      Plek.find("search"),
+      bearer_token: ENV.fetch("RUMMAGER_BEARER_TOKEN", "example"),
```
brucebolt, line 31: You shouldn't need a bearer token here as this application will only be reading from Search API (like [finder-frontend](https://github.com/alphagov/finder-frontend/blob/main/app/lib/services.rb#L23-L25)).

app/views/content_items/get_involved.html.erb
```diff
@@ -36,17 +36,13 @@
...
+          <h3><%= time_until_closure(@next_closing_consultation) %></h3>
           <ol class="document-list">
-            <% @next_closing_consultations.each do |consultation| %>
-              <%= content_tag(:li, "consultation", class: 'document-row') do %>
-                <h3><%= consultation['title'] %></h3>
-                <ul class="attributes">
-                  <li><%= link_to "Read and respond", consultation['base_path'], class: "govuk-link" %></li>
-                </ul>
-              <% end %>
+            <%= content_tag(:li, "consultation", class: 'document-row') do %>
```
brucebolt, line 41: Maybe I've missed something, but it looks like this code previously showed multiple consultations, but now only shows one?  Is that intentional?

app/helpers/get_involved_helper.rb
```diff
@@ -0,0 +1,34 @@
+module GetInvolvedHelper
+  def date_microformat(attribute_name)
+    attribute_name.to_date.strftime("%d %B %Y")
+  end
+
+  # Gets the link to the search page for all consultations
+  def get_consultations_link(filters = %w[open_consultations closed_consultations])
+    "search/policy-papers-and-consultations?#{filters.to_query('content_store_document_type')}"
```
brucebolt, line 8: Does this need to be relative (i.e. with a slash at the start), otherwise it could link to somewhere unintended?

brucebolt (commented):
Thanks for this @Tetrino.  It's great to see such a substantial piece of work come together nicely.

I've done a quick first pass review and left some comments, that shouldn't be too much effort to resolve.

I can't say I'm content with reviewing the frontend code as there's quite a lot of moving parts, so maybe a frontend developer could be sought to look over this?

app/views/content_items/get_involved.html.erb
```diff
@@ -0,0 +1,167 @@
+<% page_title %>
+<% page_class "get-involved" %>
+
+<header class="block headings-block">
+  <div class="inner-block floated-children">
+    <%= render partial: 'shared/heading',
+              locals: { big: true,
+                        heading: "Get involved" } %>
+    <p class="intro-paragraph">Find out how you can <a class="govuk-link" href="#engage-with-government">engage with government</a> directly, and <a class="govuk-link" href="#take-part">take&nbsp;part</a> locally, nationally or internationally.</p>
```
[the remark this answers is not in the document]
Tetrino: I'll make a pass to update for locale, as you can imagine a good chunk of this was completed before locale was brought in.

app/controllers/get_involved_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+  
+  def show
+    load_content_item
+    load_get_involved_data
+    @do_not_show_breadcrumbs = true
+    render template: "content_items/get_involved"
+  end
+
+  def load_content_item
+    content_item = Services.content_store.content_item("/government/get-involved")
```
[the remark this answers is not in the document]
Tetrino: Oops, good catch. I thought I'd changed this but it apparently got lost in the rebase hell.

app/assets/stylesheets/helpers/_featured_news.scss
```diff
@@ -1,6 +1,8 @@
+// stylelint-disable max-nesting-depth
```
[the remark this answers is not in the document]
Tetrino: I had rebase issues doing so, meant a lot of work had to be recovered from the tree. I'll try again.

app/views/content_items/get_involved.html.erb
Tetrino: In this case yes, this particular entry is for the consultation that explicitly shows on the top of the Get Involved page, in the blue box (which as of writing is an entry that closes tomorrow: https://www.gov.uk/government/get-involved )

Below on lines 54 and 72 are the entries for multiple consultations in their respective areas.

brucebolt: Thanks for explaining!  I missed the `[0]` from `open_consults.sort_by! { |k| k['details']['closing_date'] }[0]` in the `retrieve_next_closing` method.  Seems strange that the code in Whitehall returned a single item from the method but then iterated over an array that always contained one item in the views 🤷‍♂️.  Good that this strange logic has been refactored and simplified here 👍.

app/helpers/get_involved_helper.rb
Tetrino: Oops

brucebolt:
> Important that while this repo is continuously deployed, **this work will not be visible until enabled by a related rake task.** With this in mind it is safe to deploy.

Have you got some instructions for running this locally in `govuk-docker`?  The content item contains the existing Whitehall version of the page, so I think I'll need to update something in this content item to get this page to render in `government-frontend`.

app/controllers/get_involved_controller.rb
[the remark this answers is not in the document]
brucebolt: After some discussion on Slack, it was decided that looking this up by path is actually better, so I'll resolve this comment.

app/controllers/get_involved_controller.rb
```diff
@@ -37,50 +37,70 @@ def retrieve_given_document_type(document_type)
...
-      organisations.each do |org_id|
-        org_acronyms << Services.publishing_api.get_live_content(org_id).parsed_content['details']['acronym']
-      end
-      consultation['links']['organisation_acronyms'] = org_acronyms
+  def retrieve_take_part_pages
+    detailed_results = []
+    results = Services.search_api.search(filter_format: "take_part", order: "ordering", count: 100)["results"]
+
+    results.each do |r| 
+      detailed_results << Services.content_store.content_item(r["link"]).to_h
```
brucebolt, line 97: I'm a little concerned by the number of API calls here and the nested nature of them meaning (whilst we only have 18 at the moment) the number of calls could become rather large.

Instead I'd suggest using [link expansion](https://github.com/alphagov/publishing-api/blob/master/docs/link-expansion.md) (as recommended by @thomasleese in [this comment](https://github.com/alphagov/government-frontend/pull/1939#discussion_r555845242)).  This would mean you don't need to put the ordering into `search-api` at all.  You'd need to add the content IDs for these into the `links` section of the `/government/get-involved` content item when Whitehall Publisher presents the content item to `publishing-api`.  Publishing API will then expand the links so all the information you need is in the one content item, so no need to make a request to Search API or the content store at all as you're already retrieving the content item for the page.  This basically moves the lookup stuff to the point of publishing, rather than doing it at the point of rendering the page to a user.

You can see an example of how this is used on organisation pages, where contacts are added as `links` to the content item (e.g. [for the AAIB](https://www.gov.uk/api/content/government/organisations/air-accidents-investigation-branch)).

test/controllers/get_involved_controller_test.rb
```diff
@@ -0,0 +1,87 @@
...
+      "results" => {},
+      "total" => 83,
+      "start" => 0,
+    }
+    stub_search(body)
+
+    # Overrides for calls to publishing API made during load_get_involved_data
+    # These are tested in Smokey as failure prevents a load of some objects
+    def @controller.retrieve_given_document_type(_document_type)
+      "empty_type"
```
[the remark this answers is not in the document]
Tetrino: Resolving this as we no longer make calls to publishing API or anything related that needs stubbing.

brucebolt (commented):
Thank you for moving the hardcoded strings into locale files @Tetrino 👍 .  This will make it much easier for this page to be translated, or edited by content designers in the future.

I've added a few small comments in the latest commit.  None of them are particularly blocking, except the two where either translations are being lost or whitespace is being appended to the end of a line in the locale files.

config/locales/ar.yml
```diff
@@ -2,751 +2,798 @@
 ar:
   brexit:
     business_link:
+      path:
```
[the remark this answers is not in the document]
Tetrino: Annoyingly I ran this after generating the files, but obviously didn't check it in. Oops. Having said that, the path and text under buisiness_link and citizen_link seem genuinely new, yet unrelated to anything I've knowingly done?

lib/publish_static_pages.rb
```diff
@@ -1,42 +1,63 @@
 module PublishStaticPages
   PAGES = [
+    {
+      content_id: "dbe329f1-359c-43f7-8944-580d4742aa91",
```
brucebolt, line 4: I think I wrote this on Slack last week, but I'll write it here so it doesn't disappear into the ether.

The publishing of this page would probably be more suited to remain in Whitehall, as that is a publishing app and that's where the patched links will be coming from.

The change that will need to be made in Whitehall is to update the `rendering_app` to be `government-frontend`, then ensure the links get populated when the special routes task is published.

That would mean all the changes to `PublishStaticPages` here won't be needed, which should make things simpler.

Tetrino: https://github.com/alphagov/whitehall/pull/6303

PR #1946, opened 2020-12-18
Add filter to prevent empty image tags from rendering in figures

Tetrino:
Previously if the image source was missing the tag would still render.
This was fine in many browsers but caused a big empty space in IE.
This pull request should fix that issue.

Zendesk: https://govuk.zendesk.com/agent/tickets/4391436

### DIFF
--- a/app/views/components/_figure.html.erb
+++ b/app/views/components/_figure.html.erb
@@ -4,20 +4,19 @@
   credit ||= ''
 %>
 <figure class="app-c-figure" lang="en">
-    <% unless src.blank? %>
-      <img class="app-c-figure__image" src="<%= src %>" alt="<%= alt %>">
+  <% if src.present? %>
+    <img class="app-c-figure__image" src="<%= src %>" alt="<%= alt %>">
+  <% end %>
+  <figcaption class="app-c-figure__figcaption">
+    <% if caption.present? %>
+      <p class="app-c-figure__figcaption-text">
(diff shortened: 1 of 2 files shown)


app/views/components/_figure.html.erb
```diff
@@ -4,18 +4,19 @@
   credit ||= ''
 %>
 <figure class="app-c-figure" lang="en">
-  <img class="app-c-figure__image" src="<%= src %>" alt="<%= alt %>">
-    <figcaption class="app-c-figure__figcaption">
-      <% unless caption.blank? %>
-        <p class="app-c-figure__figcaption-text">
-          <%= caption %>
+  <% unless src.blank? %>
```
brucebolt, line 7: It's a really minor nitpick, but I know our linter prefers `if src.present?` over `unless src.blank?`.  We don't lint our views, but it's probably good practice to do that here too.

brucebolt: On the same point, I know you didn't write that code, but `unless caption.blank?` below should also be changed to `if caption.present?`.

brucebolt (commented):
Looks good.  I've left a really minor nitpick comment that should be a very minor change.

app/views/components/_figure.html.erb
Tetrino: Good shout. Updated.

brucebolt (approved):

PR #1950, opened 2020-12-22
Fix to search input

chris-gds:
## What?

Fix to resolve visual issue on `government-frontend` app within search as it doesn't import the scss from the gem.

## Why?

There was an issue highlighted with search [on this page ](https://www.gov.uk/guidance/access-the-public-register-for-environmental-information)which was related to a [search PR](https://github.com/alphagov/govuk_publishing_components/pull/1793)

## Visuals

Before
Screenshot 2020-12-22 at 18 00 57

After
Screenshot 2020-12-22 at 18 01 15

Before
Screenshot 2020-12-22 at 18 00 04

After
Screenshot 2020-12-22 at 18 00 19

## Anything else?

URL test links
https://www.gov.uk/guidance/access-the-public-register-for-environmental-information
https://www.gov.uk/guidance/alcohol-licensing

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -35,6 +35,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/radio';
 @import 'govuk_publishing_components/components/related-navigation';
 @import 'govuk_publishing_components/components/share-links';
+@import 'govuk_publishing_components/components/search';
 @import 'govuk_publishing_components/components/step-by-step-nav';
 @import 'govuk_publishing_components/components/step-by-step-nav-header';
 @import 'govuk_publishing_components/components/step-by-step-nav-related';


owenatgov (approved):
👍 🎅

36degrees:
Whilst this does fix the problem, it's because the search styles are now being included _twice_ somehow – once with a broken call to the `em` function the exact same CSS again but with the `em` function working.

I believe the correct fix would be to work out why the call to `em` in the original CSS is failing.

## Live version

Screenshot 2020-12-23 at 14 01 50

## From the review app

The sections highlighted in red and blue are the same, except for the broken call to `em`.

Screenshot 2020-12-23 at 14 02 27

36degrees:
It looks like the same issue is present across other pages on GOV.UK, it's just masked because most (?) other applications are importing the search component. You can see the same duplication and the same broken `em` call on https://www.gov.uk/browse/benefits, for example.

On that basis, it _might_ make sense to merge this as as short term fix, given it's just making it consistent with other pages on GOV.UK…?

kr8n3r:
`em` is broken in static as the [core-layout imports the search component via the header-footer-only stylesheet](https://github.com/alphagov/static/blob/master/app/assets/stylesheets/header-footer-only.scss#L10)
but it doesn't import the` px-to-em` function via `@import govuk_publishing_components/component_support` like individual apps do (or import directly

chris-gds:
> It looks like the same issue is present across other pages on GOV.UK, it's just masked because most (?) other applications are importing the search component. You can see the same duplication and the same broken `em` call on https://www.gov.uk/browse/benefits, for example.

>  em is broken in static as the core-layout imports the search component via the header-footer-only stylesheet
but it doesn't import the px-to-em function via @import govuk_publishing_components/component_support like individual apps do (or import directly)

Thanks for these insights @36degrees, @kr8n3r!  [Different PR](https://github.com/alphagov/static/pull/2380) raised on static to resolve issue, this could prompt wider discussion about isolated components & static in general.  Closing PR

PR #1951, opened 2020-12-23
Add cookieless A/A test

karlbaker02:
This PR adds a new Cookieless A/A test, which sends hit-only data to analytics up until cookies have been accepted by the user. This is necessary to understand whether we can rely on the approach offered by our CDN to reliably assign users to a variant of an A/B test and maintain that variant, such that we'd expect to see an even distribution of users across all active variants of a test (or according to any prescribed percentages of variants).

**This PR should not be merged until we're ready to turn on the A/A test.**


DilwoarH (approved):

DilwoarH:
A Review date has been put in for 27th of Jan 2021 to decide on how we will approach this going forward.

PR #1956, opened 2020-12-29
Add tracking for briefings page

DilwoarH:
Add event tracking for links within the briefings page

This will help us understand which links users are interacting with.

https://trello.com/c/BHqdvxKg/688-set-up-analytics-for-briefing-video-page

### DIFF
--- a/app/assets/javascripts/modules/track-briefing.js
+++ b/app/assets/javascripts/modules/track-briefing.js
@@ -1,0 +1,34 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function TrackBriefing () {}
+
+  TrackBriefing.prototype.start = function($element) {
+    var scope = $element[0]
+
+    this.trackLinks(scope)
+  }
+
+  TrackBriefing.prototype.trackLinks = function(scope) {
+    var links = scope.getElementsByTagName('a')
+
+    for (var i = 0; i < links.length; i++) {
+      var link = links[i]
+      link.addEventListener('click', this.sendLinkClickEvent)
+    }
+  }
+
+  TrackBriefing.prototype.sendLinkClickEvent = function(event) {
+    GOVUK.analytics.trackEvent(
+      'Briefings page',
+      event.target.innerText,
+      {
+        transport: 'beacon',
+        label: event.target.getAttribute('href')
+      }
+    )
+  }
+
+  Modules.TrackBriefing = TrackBriefing
+})(window.GOVUK.Modules)
--- a/spec/javascripts/track-briefings.spec.js
+++ b/spec/javascripts/track-briefings.spec.js
@@ -1,0 +1,73 @@
+/* global describe beforeEach it spyOn expect */
+
+var $ = window.jQuery
+
+describe('Track Briefing', function () {
+
+  var GOVUK = window.GOVUK
+  var tracker
+  var element
+
+  GOVUK.analytics = GOVUK.analytics || {}
+  GOVUK.analytics.trackEvent = function () {}
+
+  describe('Single link', function () {
+    beforeEach(function () {
+      spyOn(GOVUK.analytics, 'trackEvent')
+  
+      element = document.createElement('div')
+      element.innerHTML = '<a href="/blah/blahhhh">Blahh</a>'
+  
+      tracker = new GOVUK.Modules.TrackBriefing()
+      tracker.start($(element))
+    })
+
+    afterEach(function () {
+      GOVUK.analytics.trackEvent.calls.reset()
+    })
+  
+    it('send ga event when link is clicked', function () {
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/modules/track-briefing.js
```diff
@@ -0,0 +1,34 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function TrackBriefing () {}
```
theseanything, line 5: Could we generalise this function name as there's nothing specific here to the briefings page? 
```suggestion
  function TrackLinks () {}
```

app/assets/javascripts/modules/track-briefing.js
```diff
@@ -0,0 +1,34 @@
...
+
+  TrackBriefing.prototype.start = function($element) {
+    var scope = $element[0]
+
+    this.trackLinks(scope)
+  }
+
+  TrackBriefing.prototype.trackLinks = function(scope) {
```
theseanything, line 13:
```suggestion

```

Is there a reason to create a separate function for trackLinks here?

spec/javascripts/track-briefings.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    tracker.start($(element))
+  })
+
+  it('send ga event when link is clicked', function () {
+    element.querySelector('a').dispatchEvent(new Event('click'))
+
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+      'Briefings page', 'Blahh', { transport: 'beacon', label: "/blah/blahhhh" }
+    )
+  })
```
theseanything, line 30: Probably be good to have a test with multiple links to make sure its sending multiple events.

app/assets/javascripts/modules/track-briefing.js
```diff
@@ -0,0 +1,34 @@
...
+    var links = scope.getElementsByTagName('a')
```
theseanything, line 14: Do we wanna make this for all links? Or only external links to gov.uk? I assume we collect user journey anyways when they stay on gov.uk domain.

app/assets/javascripts/modules/track-briefing.js
```diff
@@ -0,0 +1,34 @@
...
+
+    for (var i = 0; i < links.length; i++) {
+      var link = links[i]
+      link.addEventListener('click', this.sendLinkClickEvent)
+    }
+  }
+
+  TrackBriefing.prototype.sendLinkClickEvent = function(event) {
+    GOVUK.analytics.trackEvent(
+      'Briefings page',
```
theseanything, line 24: If we generalise this function - does this need be specific about the particular page? Will the event will be assigned with a url of the page it was sent from?

theseanything (changes requested):

app/assets/javascripts/modules/track-briefing.js
DilwoarH: TrackLinks conflicts with an existing module I believe called TrackLinks - I can call it `TrackContentLinks` or something similar?

app/assets/javascripts/modules/track-briefing.js
DilwoarH: No reason, just wanted to keep it a bit clean - in case in the future other things need to be added in

app/assets/javascripts/modules/track-briefing.js
DilwoarH: We already collect external - I will ask the PA if we want both

app/assets/javascripts/modules/track-briefing.js
DilwoarH: It will be assigned with the URL of the page but we should probably change the category too

app/assets/javascripts/modules/track-briefing.js
DilwoarH: PA said all links

spec/javascripts/track-briefings.spec.js
DilwoarH: Done

app/assets/javascripts/modules/track-briefing.js
theseanything: Ah do you have a link to the module? I can't seem to find it. Also I wonder if that module has the functionality we need?

DilwoarH: Sorry, it's called TrackClick:
https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/track-click.js

theseanything: Ah thank you! - so I think TrackLink might be better just because realistically you could use this anywhere on the page - not just content

DilwoarH:
superseded by https://github.com/alphagov/govuk_publishing_components/pull/1854

PR #1958, opened 2020-12-29
Refactor Brexit callout

benthorner:
https://trello.com/c/kUqdYsCD/746-update-the-wording-of-the-whitehall-brexit-callout-box

This makes a couple of changes to make future updates to the
callout easier. Please see the commits for more details.


app/presenters/content_item/brexit_notice.rb
```diff
@@ -1,40 +1,40 @@
 module ContentItem
-  module NoDealNotice
-    def has_no_deal_notice?
+  module BrexitNotice
+    def has_brexit_notice?
       content_item.dig("details").key?("brexit_no_deal_notice")
```
huwd, line 4: mmm yeah, that's a good reminder for future eh?
overly descriptive names for things that might get hijacked in future!

huwd (approved):
Yep great work.
I partcularly agree with the thinking beind the second commit.

Things like that made tearing down some parts of the old brexit business finder much easier!
Especially when the app was old and it was approached with less context.

app/presenters/content_item/brexit_notice.rb
benthorner: Yeah, it's an interesting one: going in it may have appeared like a more specific naming was better (the old "avoid premature refactoring" mantra). I expect the product thinking at the time was all around "a brexit notice _for a no deal scenario_". As developers, the onus was really on us to ask: would we _ever_ have more than _one_ brexit notice? Even if we did, it would still be possible to distinguish them i.e. the future risk of confusion was always low.

Oh well. Easy to say all this with hindsight!

PR #1973, opened 2021-01-11
No longer render text as <strong> importance

Rosa-Fox:
The last line should not be `<strong>`, so this has been amended.

## Before

government-frontend-before

## After

gov-frontend-after


GDSNewt (approved):
:shipit:

kevindew (approved):
Code looks good 👍 

It does seem a bit surprising not having the whole paragraph bold but I'm assuming you've already dug into whether it's really wanted this way.

PR #1982, opened 2021-01-14
Fix component references

andysellick:

Update component references following this change to correct component file names: https://github.com/alphagov/govuk_publishing_components/pull/1848

Also taking the opportunity to add missing print styles for the `skip link` component.

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -8,6 +8,7 @@
 @import 'govuk_publishing_components/components/print/govspeak-html-publication';
 @import 'govuk_publishing_components/components/print/metadata';
 @import 'govuk_publishing_components/components/print/share-links';
+@import 'govuk_publishing_components/components/print/skip-link';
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/print.scss
```diff
@@ -8,6 +8,7 @@
 @import 'govuk_publishing_components/components/print/govspeak-html-publication';
 @import 'govuk_publishing_components/components/print/metadata';
 @import 'govuk_publishing_components/components/print/share-links';
+@import 'govuk_publishing_components/components/print/skip-link';
```
maxgds, line 11: This change doesn't look like one that's described in the PR.

maxgds (approved):
All looks fine, just think the skip-link addition might want to be mentioned in the PR as it doesn't look like rename.

app/assets/stylesheets/print.scss
andysellick: Good point, have updated the PR description, thanks.

PR #1999, opened 2021-01-26
Change call out html to markdown

DilwoarH:
## What
Changes HTML content to markdown.

## Why
This will allow content designers to update content without dev requirement.

## Visual Change

TLDR; No Change

### Before

Screenshot 2021-01-26 at 12 00 01

### After

Screenshot 2021-01-26 at 12 00 12


leenagupte (commented):
This is a nice idea but I think we should hold off on making these changes at the moment. The whole Ireland / not-Ireland thing makes updating travel advice complicated, and I don’t think adding govspeak is going to make anyone’s life easier at the moment.

It’s also very likely that other countries will start having differences in their callouts because COVID, so I don’t think we can escape developer involvement.

app/views/content_items/travel_advice.html.erb
```diff
@@ -26,23 +26,9 @@
       </div>
       <div class="travel-advice-notice__content">
         <% if @content_item.ireland? %>
-          <%= render 'content_items/ireland_covid_travel_guidance' %>
+          <%= render_govspeak render('content_items/travel_advice/ireland_covid_travel_guidance.md') %>
```
kevindew, line 29: A thing to note is that you're suffixing this file with md to imply it's a markdown file but then the function is treating the content as govspeak.

It might be overkill to have govspeak in the mix and better to just have a render_markdown function. That'd save having to bring over all the other dependencies that come with govspeak - similar to how we do this on https://github.com/alphagov/govuk_publishing_components/blob/5d4ed0788b5dd2a9865b45eb7d062eb7f59dd73a/app/models/govuk_publishing_components/component_doc.rb#L86-L88

DilwoarH: Ahh that's a great suggestion, thanks Kevin!

DilwoarH:
> This is a nice idea but I think we should hold off on making these changes at the moment. The whole Ireland / not-Ireland thing makes updating travel advice complicated, and I don’t think adding govspeak is going to make anyone’s life easier at the moment.
> 
> It’s also very likely that other countries will start having differences in their callouts because COVID, so I don’t think we can escape developer involvement.

Agree with your points, I'll close this for now then we can reopen this again if need

PR #2003, opened 2021-02-01
Enable 'standard' publication types to be rendered

1pretz1:
Adds locales needed to render the new 'standard' document type.

I've added the foreign translation placeholders as they are used unlike the the ones located in [Whitehall](https://github.com/alphagov/whitehall/pull/5969). We'll need to decide if we want a translation for all locales, some or just for English.

Dependent on:

- [x] Foreign locale translations being agreed upon and implemented (decision is for no translations to be created initially) 

Trello:
https://trello.com/c/ugeapNY0/2318-5-create-new-standard-document-sub-type-in-whitehall

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -267,6 +267,13 @@ cy:
         few:
         many:
         other: Areithiau
+      standard:
+        zero:
+        one: Safon
+        two:
+        few:
+        many:
+        other: Safonau
       statement_to_parliament:
         zero:
         one: Datganiad i'r senedd
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -284,6 +284,12 @@ ar:
         few:
         many:
         other: "كلمات"
+      standard:
+        one:
+        two:
+        few:
+        many:
+        other:
       statement_to_parliament:
         zero:
         one: "تصريح للبرلمان"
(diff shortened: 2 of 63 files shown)


brucebolt (approved):

config/locales/cy.yml
```diff
@@ -267,6 +267,13 @@ cy:
         few:
         many:
         other: Areithiau
+      standard:
+        zero:
+        one: Safon
+        two:
+        few:
+        many:
```
ChrisBAshton, line 275: Can we remove the empty `zero`/`two`/`few`/`many` properties, like in the English version?

ChrisBAshton (approved):
LGTM, though I've been a little out of the loop on the i18n standardisation stuff. Is there a reason why the English version only has `one` and `other` defined, and all the translations have other (empty) counts which will likely never have values?

config/locales/cy.yml
brucebolt: The [pluralisation rules for Welsh](https://github.com/svenfuchs/rails-i18n/blob/7a273036d94373477ec1abc2be154ccdf0a252f7/rails/pluralization/cy.rb) require us to have a `zero`, `two`, `few` and `many` value.  We weren't provided with translations for these, so the empty values should help us identify outstanding translations when we do [this piece of work](https://trello.com/c/ifbyj6pJ/1831-identify-keys-in-locale-files-that-require-translations).  I've also done the same for other languages that don't have a `one-other` pluralisation rule (e.g. Croatian).

PR #2026, opened 2021-02-22
Switch to meta block

chris-gds:
## What?
Update local app Meta block to [gem component](https://components.publishing.service.gov.uk/component-guide/metadata)

## Why?

As part of on-going accessibility team work to align to gem based architecture to ensure future a11y improvements to components are cascaded to all apps

## Visuals
(Before > After)
 
Screenshot 2021-02-19 at 19 14 00

Screenshot 2021-02-19 at 19 14 37

image

## Anything else?

[Ticket](https://trello.com/c/VSxcJHxI)
Testing Links:

https://www.gov.uk/government/news/statement-from-greg-clark-secretary-of-state-for-business-energy-and-industrial-strategy

https://www.gov.uk/government/publications/dwp-welsh-language-leaflets

https://www.gov.uk/government/news/christmas-2016-prime-ministers-message

Translations are out of scope of this particular card and have been [raised here](https://trello.com/c/mEGHKNGt/2371-5-investigate-and-fix-dates-not-being-translated):

Screenshot 2021-02-19 at 10 52 16

image

### DIFF
--- a/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
+++ b/app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
@@ -9,6 +9,7 @@
 
   .metadata-column {
     padding-left: 0;
+    padding-top: govuk-spacing(3);
   }
 
   .metadata-logo {
--- a/app/presenters/content_item/metadata.rb
+++ b/app/presenters/content_item/metadata.rb
@@ -21,12 +21,10 @@ module ContentItem
 
     def publisher_metadata
       {
-        published: published,
+        from: from,
+        first_published: published,
         last_updated: updated,
-        link_to_history: updated.present?,
-        other: {
-          from: from,
-        },
+        see_updates_link: true,
       }
     end
   end
--- a/app/views/shared/_publisher_metadata_with_logo.html.erb
+++ b/app/views/shared/_publisher_metadata_with_logo.html.erb
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="metadata-logo-wrapper<%= ' responsive-bottom-margin' if @content_item.try(:logo) %>">
     <div class="govuk-grid-column-two-thirds metadata-column">
-      <%= render "components/publisher-metadata", @content_item.publisher_metadata %>
+      <%= render 'govuk_publishing_components/components/metadata', @content_item.publisher_metadata %>
     </div>
     <div class="govuk-grid-column-one-third">
       <% if @content_item.try(:logo) %>
--- a/test/controllers/content_items_controller_test.rb
+++ b/test/controllers/content_items_controller_test.rb
@@ -381,7 +381,7 @@ class ContentItemsControllerTest < ActionController::TestCase
     assert_select ".metadata-logo-wrapper", false
     assert_select ".app-c-figure", false
     assert_select ".gem-c-contextual-sidebar", false
-    assert_select ".app-c-published-dates", false
+    assert_select ".gem-c-metadata__definition", false
     assert_select ".gem-c-contextual-footer", false
   end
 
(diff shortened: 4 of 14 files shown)


chris-gds:
@injms I've noticed a footer meta element that is not part of the gem but appears on these pages.  I'm wondering if this would qualify to extend the gem, however as it appears on a different part of the page this would need some thought - I don't think there's anything that _quite_ works this way at present.  

Do you think this is worth raising as an issue on the gem to capture?

image

owenatgov:
@chris-gds Re your comment on the "published on" paragraph, I don't think this is worth raising in our gem. It's simple enough a pattern (one p tag with content within) that we wouldn't gain anything from componentising it. It does have some style associated with it I can see but I wonder if that's even necessary. Possibly removing said styling would be beneficial?

@injms Interested to hear your take if you disagree.

app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
```diff
@@ -9,6 +9,7 @@
 
   .metadata-column {
     padding-left: 0;
+    padding-top: govuk-spacing(3);
```
owenatgov, line 12: I initially wondered what this was and I can see now it's part of the design of articles that the metadata block is the same width as the main content. As someone who isn't a designer, I can see why this design has been put in place but I question how necessary it is. I personally think it looks squashed now but I'm not gonna let my **opinions** block this PR.

owenatgov (approved):
Looks lovely. One little comment but otherwise I've got no probs with this going out 👍

app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
chris-gds: image

Oh - this does the top 👆🏼 
Is this what you mean?

owenatgov: Sort of. If you look at the div where this class is being used, which wraps around the metadata component call, there's also a `govuk-grid` class that's setting the width of the metadata. If you remove it it looks like this:
Screenshot 2021-02-26 at 16 59 27

I don't think there's a problem with that design over the current, `two-thirds` one and it means lower html verbosity because we can remove this wrapping div entirely. This however is a design change and I don't think this PR shouldn't get merged just because of my opinions on said design.

chris-gds: image

I see what you mean re the code, I think this is intentional as the [visuals restriction also appear here](https://www.gov.uk/guidance/uk-carbon-capture-and-storage-government-funding-and-support) which is on `collections`. Course, still worth a noting

chris-gds:
@owenatgov ah I've not got ~~the power~~ auth to merge this - do the honours Monday (to be safe as it's CD)?

app/assets/stylesheets/helpers/_publisher-metadata-with-logo.scss
owenatgov: That's fair. This feels like a wider decision by govuk designers. Feel free to resolve 👍

owenatgov:
@chris-gds If you're ok to resolve the conflicts I can do this Monday 👍

chris-gds:
> @chris-gds Re your comment on the "published on" paragraph, I don't think this is worth raising in our gem. It's simple enough a pattern (one p tag with content within) that we wouldn't gain anything from componentising it. It does have some style associated with it I can see but I wonder if that's even necessary. Possibly removing said styling would be beneficial?
> 
> @injms Interested to hear your take if you disagree.

Makes sense re the gem. Should it be appear on some pages with meta blocks and not on others across the wider site....probably not - but yeah not part of this. Maybe a Design Enhancement issue. I'll leave for now.

PR #2031, opened 2021-03-04
Replace sanitize with raw for legacy document list

injms:
## What
Replace `sanitize` with `raw` for the legacy pre-rendered documents list.

## Why
`sanitize` was stripping out the `tabindex` attribute from the legacy document list markup provided by Whitehall. `sanitize` requires an allowlist of attributes, but a more future-proof option is to use `raw` since this is markup from a trusted source.

## Visual differences

The document icon was focusable with an invisible focus state - so a user could tab there but there was no visual indication that the document icon was focused. This removes that link from the tab index, so it's no longer reachable by tabbing through the page. This means that the tab focus jumps from the document title to the next document title without detouring.

### DIFF
--- a/app/views/content_items/_attachments.html.erb
+++ b/app/views/content_items/_attachments.html.erb
@@ -8,10 +8,7 @@
       <%= render 'govuk_publishing_components/components/govspeak', {
         direction: page_text_direction,
       } do %>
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
🎉

PR #2034, opened 2021-03-05
Updates for How To Vote guide

sihugh:
We're going to run https://www.gov.uk/how-to-vote as a single page guide up to the local elections. This allows us to use the FAQPage schema to extend the reach of the content of this page to other platforms.

We wrote about [doing this for the general election](https://gds.blog.gov.uk/2019/12/19/making-gov-uk-more-than-a-website/).

This PR resurrects the old way of doing this, by:
- reverting two old commits
- updating the slugs (because it used to be called /how-to-vote
- updating the content from the [document agreed with stakeholders](https://docs.google.com/document/d/17i_0rH67XgBn2dy0DIQbCC0PLxiZofTkohbw042qtV0/edit?ts=603e18cb#)
- fixing a failing test due to changes in stubbing methods

Note, we may want to review the single page guide template, as some elements of GOV.UK frontend have moved on.

https://government-f-voting-int-jrhxkr.herokuapp.com/how-to-vote


config/routes.rb
```diff
@@ -15,6 +15,9 @@
 
   get "/government/uploads/*path" => "asset_manager_redirect#show", format: false
 
+  # Testing guides as a single page so we redirect parts to the default page
+  get "/voting-in-the-uk/:chapter", to: redirect("/voting-in-the-uk#%{chapter}")
```
alex-ju: Nice one! 👍

config/machine_readable/how-to-vote.yml
```diff
@@ -34,12 +34,11 @@ faqs:
 
   - question: When you can vote
     answer: >
-      <p>Polling stations are open from 7am to 10pm on the day of an election ('polling day').</p>
+      <p>Polling stations are open from 7am to 10pm on 6 May ('polling day').</p>
```
alex-ju, line 43: `Polling stations will be open from 7am to 10pm on 6 May (‘polling day’).`

app/views/content_items/guide_single.html.erb
```diff
@@ -0,0 +1,53 @@
...
+
+  <div class="govuk-grid-column-two-thirds">
+    <% if @content_item.has_parts? %>
+
+      <% @content_item.parts.each_with_index do |part, index| %>
+      <section class="govuk-!-padding-bottom-9" aria-labelledby="<%= part["slug"] %>">
+        <%= render "govuk_publishing_components/components/heading", {
+            text: part["title"],
+            id: part["slug"],
+            heading_level: 1,
```
alex-ju, line 36: This would need to be a level 2 heading – Although this means the content we pass trough govspeak on L42 would need to have the heading levels moved down (H2->H3). While the heading change won't affect the page visually, the content change (coming from `part["body"]`) will 😬 .

Given this is not a change introduced by this PR (we already have this structure on the current page, I'd be tempted to leave it as is (H1) and address it separately (it may already be in the A11y team's backlog, I suspect).

app/views/content_items/guide_single.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <%= render "govuk_publishing_components/components/contents_list", aria_label: 'Pages in this guide', contents: @content_item.part_link_elements, underline_links: true %>
+      </aside>
+    <% end %>
+  </div>
+
+  <div class="govuk-grid-column-two-thirds">
+    <% if @content_item.has_parts? %>
+
+      <% @content_item.parts.each_with_index do |part, index| %>
+      <section class="govuk-!-padding-bottom-9" aria-labelledby="<%= part["slug"] %>">
```
alex-ju, line 32: The padding class `govuk-!-padding-bottom-9` is not ideal, we could create a class for this section in future. We already used them widely across the app, it's not something to be changed in this PR.

alex-ju (approved):
Looks good overall. Only one content-related comment to be checked. I don't mind longer pages as long as users can navigate to them using the contents list and share direct links to those sections. I get that this is a trial and we can make the 'bespoke' code more generic if we decide to move more guides to this model.

sihugh:
> Looks good overall. Only once content-related comment to be checked. I don't mind longer pages as long as users can navigate to them using the contents list and share direct links to those sections. I get that this is a trial and we can make the 'bespoke' code more generic if we decide to move more guides to this model.

Totally agree 👍

app/views/content_items/guide_single.html.erb
sihugh: Yeah, I spotted that [the first time I did this](https://github.com/alphagov/government-frontend/pull/1481). I think the risks of trying to cascade changes to HTML heading levels are probably too high to pick up here. I think the A11y team graded it as low priority (see my comment on the previous PR, though I can't see the Trello board any more)

config/machine_readable/how-to-vote.yml
sihugh: Good spot, thank you!

PR #2035, opened 2021-03-05
March 8th Travel Advice callout update

kevindew:
This changes the copy to the newly requested copy. With this change
there is no longer separate advice for Ireland so this file has been
removed.

Before:

Screenshot 2021-03-05 at 17 51 04

After:

Screenshot 2021-03-05 at 17 52 11


sihugh (approved):
✈️

PR #2059, opened 2021-03-24
Remove translation keys that do not exist in the English version

benjamineskola:
Most of these are blank keys anyway. The exception is the `taxonomy_navigation` key which only exists in Arabic and Russian.

https://trello.com/c/qlDQFq2I/2428-3-remove-locale-keys-with-no-english-translation

### DIFF
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -14,11 +14,6 @@ ar:
       topical_events: "أحداث ذات صلة بالموضوع"
       topics: "المواضيع"
       world_locations: "العالم"
-    taxonomy_navigation:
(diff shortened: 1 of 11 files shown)


MuriloDalRi:
Is there any reason why those translations only exist in Arabic and Russian?

MuriloDalRi (approved):

benjamineskola:
> Is there any reason why those translations only exist in Arabic and Russian?

I'm not really sure. It does make me a little suspicious — but I also can't find anywhere where they would be used from, either. (And after all if they were used then we would notice odd behaviour on the English version of the pages.)

PR #2086, opened 2021-04-23
Remove unused i18n keys

ChrisBAshton:
This PR adds the [i18n-coverage gem](https://github.com/hiptest/i18n-coverage) to monitor untested i18n keys. I've then removed used the coverage reports to identify and remove unused keys. See commits for details.

Trello: https://trello.com/c/Z3Pitrfg/2454-identify-and-remove-unused-keys-in-locale-files-5

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ end
 group :test do
   gem "capybara"
   gem "faker"
+  gem "i18n-coverage"
   gem "minitest-reporters"
   gem "mocha"
   gem "simplecov"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -152,6 +152,7 @@ GEM
       domain_name (~> 0.5)
     i18n (1.8.10)
       concurrent-ruby (~> 1.0)
+    i18n-coverage (0.2.0)
     image_size (2.1.0)
     io-like (0.3.1)
     jasmine (3.6.0)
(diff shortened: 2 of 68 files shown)


thomasleese:
Do we know if there's any risk or problems with lines like https://github.com/alphagov/government-frontend/blob/e5c0c97dc614b9513b82e106074e9293826498a4/app/presenters/content_item/withdrawable.rb#L33 causing us to think a translation isn't used when actually it is?

ChrisBAshton:
> Do we know if there's any risk or problems with lines like
> 
> https://github.com/alphagov/government-frontend/blob/e5c0c97dc614b9513b82e106074e9293826498a4/app/presenters/content_item/withdrawable.rb#L33
> 
> causing us to think a translation isn't used when actually it is?

Yes, that is a risk. I manually searched for portions of the key, rather than the entire key, so that I could spot programmatic occurrences like this, in which case I've kept the keys rather than removed them.

I think the bigger risk is that there are some schema names in the i18n which aren't ever called upon programmatically. If, for example, there is no `national` schema, then [the `national` key would be redundant](https://github.com/alphagov/government-frontend/blob/1d7e2e1d0e3502e30c47df7c294d3548c71da5aa/config/locales/en.yml#L136-L138), but there's no way to easily check. So the risk is we're not deleting enough (rather than we're deleting too much).

thomasleese (approved):

benthorner:
I spotted a slight issue with the config here while testing: https://github.com/alphagov/government-frontend/pull/2112

PR #2098, opened 2021-05-05
Retire hardcoded schema at 10pm on voting day

sihugh:
The polls close at 10:00pm, so we should turn the hardcoded schema off at that time.

Updates to the content are scheduled to be published at that time too, so the cache will be cleared by our standard processes which is handy :-) .


hannako (approved):
LGTM 👍

PR #2112, opened 2021-05-19
Fix error when running a single test

benthorner:
The require statement for the printer doesn't replace the general one for
the module. It's actually unclear how this worked without it, but I don't
think that's worth investigating.

Error, for reference:

```
 /govuk/government-frontend/test/test_helper.rb:10:in `<top (required)>':
undefined method `config' for I18n::Coverage:Module (NoMethodError)
```


ChrisBAshton (approved):

benthorner:
Thanks for the review @ChrisBAshton! You'll need to merge, by the way...

PR #2116, opened 2021-05-24
Remove hardcoded strings from controllers, presenters and views

brucebolt:
Moves all hardcoded English strings from controllers, presenters and views to the English locale file and adds placeholder keys for these in all other languages.

This will allow us to gain visibility of content that needs to be translated in order to fully support non-English content on GOV.UK.

This does not translate [the "histories" pages](https://github.com/alphagov/government-frontend/tree/release_1749/app/views/histories) which contain substantial amounts of hardcoded English content.

[Trello card](https://trello.com/c/vbFb9GT2)

### DIFF
--- a/app/views/content_items/contact.html.erb
+++ b/app/views/content_items/contact.html.erb
@@ -17,7 +17,7 @@
 
     <% body = capture do %>
       <% if @content_item.online_form_links.any? %>
-        <h2 id="online-forms-title">Online</h2>
+        <h2 id="online-forms-title"><%= t("contact.online") %></h2>
         <% @content_item.online_form_links.each do |link| %>
           <p>
             <%= link_to(link[:title], link[:url]) %>
@@ -28,12 +28,12 @@
       <% end %>
 
       <% if @content_item.show_webchat? %>
-        <h2 id="webchat-title">Webchat</h2>
+        <h2 id="webchat-title"><%= t("contact.webchat") %></h2>
         <%= render 'shared/webchat' %>
       <% end %>
 
       <% if @content_item.email.any? %>
-        <h2 id="email-title">Email</h2>
+        <h2 id="email-title"><%= t("contact.email") %></h2>
         <% @content_item.email.each do |email_group| %>
           <p class="vcard">
             <% email_group[:v_card].each do |v_card_part| %>
@@ -50,7 +50,7 @@
       <% end %>
 
       <% if @content_item.phone.any? %>
-        <h2 id="phone-title">Phone</h2>
+        <h2 id="phone-title"><%= t("contact.phone") %></h2>
         <% @content_item.phone.each do |phone_group| %>
           <% if @content_item.phone.size > 1 %>
             <h3><%= phone_group[:title] %></h3>
@@ -77,7 +77,7 @@
         <%= @content_item.phone_body %>
 
         <p class="call-charges">
-          <%= link_to "Find out about call charges", "/call-charges" %>
+          <%= link_to t("contact.find_call_charges"), "/call-charges" %>
         </p>
       <% end %>
 
--- a/app/views/content_items/gone.html.erb
+++ b/app/views/content_items/gone.html.erb
@@ -14,7 +14,7 @@
 
     <% if @content_item.alternative_path.present? %>
       <p class="alternative">
-        <%= t("gone.visit") %> <%= alternative_path_link(request, @content_item.alternative_path)%>
+        <%= t("common.visit") %> <%= alternative_path_link(request, @content_item.alternative_path)%>
       </p>
     <% end %>
   </div>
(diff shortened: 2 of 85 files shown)


app/views/content_items/contact.html.erb
```diff
@@ -28,12 +28,12 @@
...
 
       <% if @content_item.show_webchat? %>
-        <h2 id="webchat-title">Webchat</h2>
+        <h2 id="webchat-title"><%= t("contact.webchat") %></h2>
         <%= render 'shared/webchat' %>
       <% end %>
 
       <% if @content_item.email.any? %>
-        <h2 id="email-title">Email</h2>
+        <h2 id="email-title">E<%= t("contact.email") %></h2>
```
thomasleese, line 36:
```suggestion
        <h2 id="email-title"><%= t("contact.email") %></h2>
```

app/views/content_items/gone.html.erb
```diff
@@ -14,7 +14,7 @@
 
     <% if @content_item.alternative_path.present? %>
       <p class="alternative">
-        Visit: <%= alternative_path_link(request, @content_item.alternative_path)%>
+        <%= t("gone.visit") %>: <%= alternative_path_link(request, @content_item.alternative_path)%>
```
thomasleese, line 17: Do we want to include the `:` in the string? Do all languages use the same `:` character to describe an explanation?

app/views/content_items/unpublishing.html.erb
```diff
@@ -12,7 +12,7 @@
 
     <% if @content_item.alternative_url.present? %>
       <p class="alternative">
-        Visit: <%= link_to(@content_item.alternative_url, @content_item.alternative_url, class: "govuk-link") %>
+        <%= t("unpublishing.visit") %>: <%= link_to(@content_item.alternative_url, @content_item.alternative_url, class: "govuk-link") %>
```
thomasleese, line 15: Is it worth re-using the same key here as `gone.visit`?

app/views/content_items/html_publication/_print_meta_data.html.erb
```diff
@@ -2,20 +2,20 @@
...
 </p>
 <p>
-  This publication is available at <%= @content_item.full_path(request) %>
+  <%= sanitize(t("html_publication.print_meta_data.available_at", url: @content_item.full_path(request))) %>
 </p>
 
 <% unless @content_item.isbn.blank? %>
     <p>
-      ISBN: <%= @content_item.isbn %>
+      <%= t("html_publication.print_meta_data.isbn") %>: <%= @content_item.isbn %>
```
thomasleese, line 19: Similar question here about the `:` as with `Visit:`.

thomasleese (commented):
Looks good, I've left a few comments, although nothing major.

app/views/content_items/contact.html.erb
brucebolt: Good spot!

app/views/content_items/gone.html.erb
brucebolt: Good idea, I'll move this to the locale file

app/views/content_items/unpublishing.html.erb
brucebolt: Yeah, I'll use a shared key for this

app/views/content_items/html_publication/_print_meta_data.html.erb
brucebolt: Thanks, I'll add that to the translation here too

1pretz1:
Nice work @brucebolt, just have one question. Are we not including specs in these changes? It's quite useful to have the specs using i18n as it means locale changes don't require the tests to be updated.

brucebolt:
> Nice work @brucebolt, just have one question. Are we not including specs in these changes? It's quite useful to have the specs using i18n as it means locale changes don't require the tests to be updated.

That's a good question that I hadn't considered 🙈 .  I'll see how much work it would be to update these too.

brucebolt:
> Nice work @brucebolt, just have one question. Are we not including specs in these changes? It's quite useful to have the specs using i18n as it means locale changes don't require the tests to be updated.

Looks like this could be a rather difficult task.  We've got strings in the tests that come from `government-frontend`, but also others from `whitehall` and other places so there will be a mixture of translations and hardcoded strings.  Maybe we should leave the tests as there are for now, then they can be updated when the tests are next updated?

1pretz1:
Sure thing, I don't think it's important enough to block this work but is certainly a nice to have. As you say, these can be updated as and when they're changed, hopefully 🤞 

I'll probably add them in for collections as I've already done some, and it hasn't got hairy, yet... 😄

thomasleese (approved):

1pretz1 (approved):
Looks good to me 🍏

PR #2117, opened 2021-05-25
Replace context link with breadcrumb

maxgds:
## What
https://trello.com/c/3ENWcE93/729-confusing-contextual-link-on-some-pages
Remove the link from the context text on the main heading, add breadcrumb to replace it.

## Why

The styling of the contextual text in the component doesn't indicate in any way that it is a link without user interaction. We plan to remove the ability to add links from the components gem.  We decided to preserve the linking back to the History page by adding a breadcrumb. The wording there more explicitly describes where the link goes.

### Before
Screenshot 2021-05-25 at 15 44 16

### After
Screenshot 2021-05-25 at 15 43 10

### DIFF
--- a/app/views/histories/10_downing_street.html.erb
+++ b/app/views/histories/10_downing_street.html.erb
@@ -1,9 +1,20 @@
 <header class="govuk-grid-row">
   <div class="govuk-grid-column-full">
+    <%= render "govuk_publishing_components/components/breadcrumbs", {
+      breadcrumbs: [
+        {
+          title: "Home",
+          url: "/",
+        },
+        {
+          title: "History of the UK Government",
+          url: "/government/history",
+        },
+      ]
+    } %>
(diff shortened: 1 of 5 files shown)


chris-gds (approved):

chris-gds:
All looks good to me

app/views/histories/10_downing_street.html.erb
```diff
@@ -1,9 +1,20 @@
...
   <div class="govuk-grid-column-full">
+    <%= render "govuk_publishing_components/components/breadcrumbs", {
+      breadcrumbs: [
+        {
+          title: "Home",
+          url: "/",
+        },
+        {
+          title: "History of the UK Government",
+          url: "/government/history",
```
chris-gds, line 11: I do wonder if there's a way to automate these for breadcrumbs so no matter where it appears on the site - via the gem -  it's pre-populated.

PR #2138, opened 2021-06-09
Fix exposed html from translation

maxgds:
### What

The text content from this page comes from a translation file. This means that any coding gets exposed unless we deliberately render it as raw. Since we control the translations we can be confident that doing this is safe. The error that was exposed was noted by a user in https://govuk.zendesk.com/agent/tickets/4582865

Sample URL: https://www.gov.uk/government/consultations/tobacco-and-related-products-legislation-introduced-between-2015-to-2016-reviewing-effectiveness

### Before
Screenshot 2021-06-09 at 13 13 28

### After
Screenshot 2021-06-09 at 13 13 14

### DIFF
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -21,7 +21,7 @@
 
     <% if @content_item.unopened? %>
       <% content_item_unopened = capture do %>
-        <%= t("consultation.opens") %>
+        <%= raw(t("consultation.opens")) %>
         <%= @content_item.opening_date_midnight? ? t("consultation.on") : t("consultation.at") %>
         <time datetime="<%= @content_item.opening_date_time %>"><%= @content_item.opening_date %></time>
       <% end %>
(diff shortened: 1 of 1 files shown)


theseanything:
Could we just remove the nbsp from the local text? https://github.com/alphagov/government-frontend/blob/bff5929da7075f74cdf0d849ec3edde67bc7d31f/config/locales/en.yml#L52. Not sure why we have a non-breaking space there anyways?

thomasleese:
> . Not sure why we have a non-breaking space there anyways?

I wonder if it's intentional to make sure those words always appear together rather than split over multiple lines?

thomasleese (approved):

maxgds:
> Could we just remove the nbsp from the locale text?
> 
> https://github.com/alphagov/government-frontend/blob/bff5929da7075f74cdf0d849ec3edde67bc7d31f/config/locales/en.yml#L52
> 
> . Not sure why we have a non-breaking space there anyways?

Like Tom said - it means you don't end up with a single word on one line. It's hacky but it works and is a valid thing people do.

theseanything:
Sure - it just seems inconsistent as we don't put non-breaking spaces before the last word on other bit of locales. Don't think I understand the user issue of having a single word by itself on the next line, in this particular instance.

PR #2144, opened 2021-06-14
Brexit child taxon descriptions

hannako:
## What

No visual changes in this PR, I'm just amending the language we use in the code base to keep it in line with the terminology being used more widely in the team, and to make a few minor refactors before we get too far along. 


app/presenters/content_item/brexit_taxons.rb
```diff
@@ -8,15 +8,15 @@ module BrexitHubPage
 
     def brexit_links
       {
-        ContentItem::BrexitHubPage::BREXIT_BUSINESS_PAGE_CONTENT_ID => {
+        ContentItem::BrexitTaxons::BREXIT_BUSINESS_PAGE_CONTENT_ID => {
```
sihugh, line 11: We should be able to get rid of `ContentItem::BrexitTaxons::` in this bit as it's referring to things within this class, shouldn't we? We'll obvs need to keep them in the tests

sihugh (approved):
Minor comment, but nice gardening 👍

PR #2162, opened 2021-06-29
Fix button colour

chris-gds:
## What 

Fix to resolve buttons text colours on buttons
https://www.gov.uk/government/organisations/department-for-work-pensions/about/accessible-documents-policy

## Why

Button text colours are being overwritten due to the styles applying department branding, this is an WCAG fails as the colour contrast makes this text unreadable therefore impact low vision and blind users.

## Visuals

image

## Anything else

As part of an on-going Epic corporate information templates will be updated entirely so this will act as a temporally fix until this work is complete

### DIFF
--- a/app/assets/stylesheets/helpers/_organisation-links.scss
+++ b/app/assets/stylesheets/helpers/_organisation-links.scss
@@ -5,7 +5,7 @@
 @mixin organisation-links {
   @each $organisation in map-keys($govuk-colours-organisations) {
     .#{$organisation}-brand-colour {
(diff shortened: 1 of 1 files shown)


maxgds (approved):

maxgds:
Looks good to me, but I just realised you still have it in draft mode. Let me know if it wants merging.

chris-gds:
@maxgds Ace moved out of draft - just waiting for deployment to check.
https://government-f-fix-button-dfv89g.herokuapp.com/government/organisations/department-for-work-pensions/about/accessible-documents-policy

PR #2197, opened 2021-08-06
Fix tracking Travel Advice links twice

kevindew:
There is a bug where these navigation links are sending two GA events
per click. This is occurring because the contents-list component has an
undocumented [1] feature of tracking links which was presumably not
known about when this code was changed to use this component [2].

The resolution for this is to remove the usage of track links on the nav
element and just use the components one.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/147a825fc7608eb3dfdaab0d418c5b2fdaaf08aa/app/views/govuk_publishing_components/components/docs/contents_list.yml
[2]: https://github.com/alphagov/government-frontend/pull/1877

### DIFF
--- a/app/views/content_items/travel_advice.html.erb
+++ b/app/views/content_items/travel_advice.html.erb
@@ -30,7 +30,7 @@
     </div>
 
     <aside class="part-navigation-container" role="complementary">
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Makes sense to me

PR #2199, opened 2021-08-10
Return response headers for Explore AB test

sihugh:
At present we're only returning Vary response headers when we get a B variant. We want to do this all the time or it's confusing for Fastly to know how to cache.

I've added a test that ensures (using the `with_variant` helper) that the header is being set correctly, and also checks for analytics set up.

It's trickier to assert things about the template because slimmer's testing is harder to figure out right now, but it'd be useful to add these at some point too. :-)


injms (approved):
Thank you!

PR #2206, opened 2021-08-16
Swap to `gem_layout`

alex-ju:
### What

Update `government-frontend` to use the Design System-based page layout

### Why

Major benefits include performance improvements, accessibility improvements, and consistency with other applications.

### Visual changes
Example page


app/controllers/development_controller.rb
```diff
@@ -1,4 +1,6 @@
 class DevelopmentController < ApplicationController
+  include Slimmer::Template
+  slimmer_template "gem_layout"
```
injms, line 3: I think that these two lines need to be in `application_controller.rb`, rather than in `developerment_controller.rb`:

```ruby
  include Slimmer::Template
  slimmer_template "gem_layout"
```

injms (changes requested):

app/views/components/_published-dates.html.erb
```diff
@@ -22,7 +22,7 @@
       data-expanded="false"
       data-toggled-text="&#45;&nbsp;<%= t('components.published_dates.hide_all_updates', locale: :en) %>">&#43;&nbsp;<%= t('components.published_dates.show_all_updates', locale: :en) %></a>
       <div class="app-c-published-dates__change-history js-hidden" id="full-history">
-        <ol>
+        <ol class="app-c-published-dates__list">
```
injms, line 25: If all the class `app-c-published-dates__list` is doing is removing the padding, should that be done with `govuk-!-padding-0`?

chao-xian:
I'm seeing an issue locally with the header so just deploying onto Heroku to see if the issue is also there.

chao-xian:
Yeah this is what I'm seeing locally and on Heroku:

government-f-swap-to-ge-zxvfvb herokuapp com_government_case-studies_using-rewards-encouraging-good-behaviour

alex-ju:
> Yeah this is what I'm seeing locally and on Heroku:
> 
> government-f-swap-to-ge-zxvfvb herokuapp com_government_case-studies_using-rewards-encouraging-good-behaviour

Oh.. ouch. It seems like the layout broke after rebasing, indeed. I will investigate this further. Converting it into a draft in the meantime.

app/controllers/development_controller.rb
alex-ju: It seems that I've lost some code after rebasing as https://github.com/alphagov/government-frontend/pull/2235/files#diff-8f720467cd4cfd6b9a611598b9491f0ffa8582cd2d260b733a36e16c0179bfd0L62 got merged and removed slimmer_template declarations where `gem_layout` was used. Will update.

app/views/components/_published-dates.html.erb
alex-ju: I think the use of `!important` declarations should be avoided (unless we're looking for a quick-and-dirty fix). This component follows our [convention to use BEM notations](https://github.com/alphagov/govuk_publishing_components/blob/9fc1eae99b6e3eccb8a51dab4458038b7ea63eca/docs/component_conventions.md#bem) with `app-c-published-dates__list` fitting nicely as an element. Today the element has a padding 0, tomorrow may have a border or a background – the markup shouldn't need to be changed to reflect design adjustments.

alex-ju:
> I'm seeing an issue locally with the header so just deploying onto Heroku to see if the issue is also there.

Thanks for spotting this @chao-xian! I have now adapted my code to the recent changes.

chao-xian (approved):
WHOOHOO! Thanks Alex! 🦸

PR #2215, opened 2021-09-02
Add tracking to the travel advice pages

andysellick:

- uses the new feature of the track click script to track link clicks only inside the grey callout box (the second callout box on the page)
- this is inside govspeak content so we can't directly apply tracking to it, hence the new feature of the track click script to limit tracking to a specific element

Trello card: https://trello.com/c/JREgQ2iv/579-spike-into-tracking-enhancements

### DIFF
--- a/app/views/content_items/travel_advice.html.erb
+++ b/app/views/content_items/travel_advice.html.erb
@@ -52,11 +52,17 @@
       <%= render 'shared/travel_advice_summary', content_item: @content_item %>
     <% end %>
 
-    <%= render 'govuk_publishing_components/components/govspeak', {
-      direction: page_text_direction,
-    } do %>
-      <%= raw(@content_item.current_part_body) %>
-    <% end %>
+    <div data-module="gem-track-click"
+      data-track-category="SummaryTravelAdviceWarning"
+      data-track-action="callOutBoxClicked"
+      data-track-links-only
+      data-limit-to-element-class="call-to-action">
+      <%= render 'govuk_publishing_components/components/govspeak', {
+        direction: page_text_direction,
+      } do %>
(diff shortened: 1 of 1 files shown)


andysellick:
@leenagupte you're right, this tracking wraps the majority of the page content and not only the grey box (the thing we want to track). 

This is why I've added a feature to the track clicking script `data-limit-to-element-class`, which enables the tracking to be targeted to (links in) a specific element only. This PR will fire tracking events only when links in the `.call-to-action` box are clicked.

leenagupte:
> @leenagupte you're right, this tracking wraps the majority of the page content and not only the grey box (the thing we want to track).
> 
> This is why I've added a feature to the track clicking script `data-limit-to-element-class`, which enables the tracking to be targeted to (links in) a specific element only. This PR will fire tracking events only when links in the `.call-to-action` box are clicked.

Ah, OK, so it's looking for `<div class="call-to-action">` within `current_part_body` and just adds the tracking to that bit?

andysellick:
> Ah, OK, so it's looking for `<div class="call-to-action">` within `current_part_body` and just adds the tracking to that bit?
> On the off-chance that the name of the div changes, I think this tracking will stop working unexpectedly. Is there anyway to mitigate that?

Correct. And yes, if the class name of the box changes then the tracking will stop working, so it's not ideal, but it's the only way to put tracking into this area. However, the class name is a govspeak class, which is very unlikely to change - we don't often make changes to govspeak like that. And if the tracking does stop working and it's still required, I'm sure the PA community will notice.

leenagupte (approved):
Thanks for the explanations 💯 
I've tested this locally, and it LGTM 👍

PR #2220, opened 2021-09-09
Use devolved nations component in place of important metadata component on publication page

jon-kirwan:
## What
https://trello.com/c/WVrZ35zt/793-dev-implementation-of-devolved-nations-task

Replace uses of 'Metadata block component' for new 'Devolved Nations component' for linking to guidance for other nations.

This change takes effect on the following page types -
- Publications e.g. https://www.gov.uk/government/publications/theory-and-practical-driving-tests-taken-at-wigan-and-atherton-with-a-voiceover-or-interpreter
- Statistics e.g. https://www.gov.uk/government/statistics/affordable-housing-supply-in-england-2011-to-2012
- Consultations e.g. https://www.gov.uk/government/consultations/setting-the-grade-standards-of-new-gcses-in-england-2017-2018
- Guidance e.g. https://www.gov.uk/guidance/report-lost-or-stolen-cattle

When no advice is available the component is not rendered e.g. https://www.gov.uk/government/publications/foi-release-technology-and-outserviced-contracts

## Why
- The list of countries are being read out without providing any context that individual guidance is available
- This task is to implement a fix to aid screen-reader users

## Visual changes
| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

## Anything else
- Prerequisite now merged https://github.com/alphagov/government-frontend/commit/2a3d5d4c0fd03980bf60bc68f3659caaab0ddd5b

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -14,6 +14,7 @@ $govuk-new-link-styles: true;
 @import 'govuk_publishing_components/components/contents-list';
 @import 'govuk_publishing_components/components/contextual-sidebar';
 @import 'govuk_publishing_components/components/details';
+@import 'govuk_publishing_components/components/devolved-nations';
 @import 'govuk_publishing_components/components/document-list';
 @import 'govuk_publishing_components/components/error-message';
 @import 'govuk_publishing_components/components/error-summary';
--- a/app/presenters/content_item/national_applicability.rb
+++ b/app/presenters/content_item/national_applicability.rb
@@ -2,43 +2,8 @@ module ContentItem
   module NationalApplicability
     include Metadata
 
-    def applies_to
-      return nil unless national_applicability
-
-      all_nations = national_applicability.values
-      applicable_nations = all_nations.select { |n| n["applicable"] }
-      inapplicable_nations = all_nations - applicable_nations
-
-      applies_to = applicable_nations.map { |n| n["label"] }.to_sentence
-
-      if inapplicable_nations.any?
-        nations_with_alt_urls = inapplicable_nations.select { |n| n["alternative_url"].present? }
-        if nations_with_alt_urls.any?
-          alternate_links = nations_with_alt_urls
-            .map { |n| view_context.link_to(n["label"], n["alternative_url"], rel: :external, class: "govuk-link govuk-link--inverse") }
-            .to_sentence
-
-          applies_to += " (see #{translated_schema_name(nations_with_alt_urls.count)} for #{alternate_links})"
-        end
-      end
-
-      applies_to
-    end
-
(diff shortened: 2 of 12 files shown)


jon-kirwan:
> Can you explain to me how you identified the formats that are using the important metadata for rendering devolved nations info? I can see that it's used across a number of formats and while I can see some are clearly using it for a different purpose (eg the Fatality notices: https://www.gov.uk/government/fatalities/sergeant-eddie-collins-killed-in-iraq) I can't see how/where it's used on the likes of document collections: https://www.gov.uk/government/collections/severn-tidal-power-feasibility-study-conclusions

Yep, so I've gone about this in two ways -
- I've updated formats that have associated tests for devolved nations content and 
- I've checked against a CSV file (attached here [national-applicability.csv](https://github.com/alphagov/government-frontend/files/7163810/national-applicability.csv) and which _I think_ I sourced from the Trello card) that lists uses of the component (with devolved nations content). The listed pages seem to be of the following type -
- - https://www.gov.uk/guidance
- - https://www.gov.uk/government/statistics
- - https://www.gov.uk/government/publications
- - https://www.gov.uk/government/consultations

jon-kirwan:
> I'm intrigued by the "temp" commit that seems to be doing the sort of stuff I'd expect to be handled by dependabot PRs. I think I saw a reference to it needing to be deleted at one point on this PR?

Ah I'd added this as a temporary way to test the new component (prior to the dependabot merge) and forgot to go back and remove it once the version had been bumped! I've now removed the extra commit.

jon-kirwan:
> 1. I don't think the metadata component itself is especially inaccessible, it was more that the way we were using it for this particular data didn't work well as the links were unhelpfully labelled.

I've deleted from the _'Why'_ section - _'Currently, this "component" isn't accessible to screen-reader users.'_ so it's hopefully now clearer that the fail is related to the hyperlink context rather than the actual component itself.

> 2. We should link our PRs back to the Trello card so it's easy to trace further info. Sometimes putting the trello card url on the PR auto-links it for us and saves us that job, but the magic that drives that hasn't been very reliable recently.

Trello link now included in the description.

maxgds:
> Yep, so I've gone about this in two ways -

Great, that makes sense and I can't see any gaps after looking through the list. If it turns out there are it'll be easy enough to update further template types.

maxgds:
I notice you've added a json lockfile in the root that wasn't there before. I assume that's an error, can you remove it (or justify its presence)?

jon-kirwan:
> I notice you've added a json lockfile in the root that wasn't there before. I assume that's an error, can you remove it (or justify its presence)?

Apologies that's a mistake - i've now removed it.

maxgds (approved):

PR #2254, opened 2021-10-25
Make intervention component future-proof

maxf:
The next version of the intervention component (starting from
govuk_publishing_components v27.8.2) doesn't render at all
when no parameter is passed. So in order to make the current tests
still pass in the future, we add a dummy parameter now

### DIFF
--- a/app/views/content_items/_body_with_related_links.html.erb
+++ b/app/views/content_items/_body_with_related_links.html.erb
@@ -9,7 +9,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <div class="responsive-bottom-margin">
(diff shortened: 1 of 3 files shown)


app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -9,7 +9,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <div class="responsive-bottom-margin">
-      <%= render 'govuk_publishing_components/components/intervention' if should_show_sab_intervention? %>
+      <%= render 'govuk_publishing_components/components/intervention', { suggestion_text: "OK" } if should_show_sab_intervention? %>
```
chao-xian, line 12: If the `OK` is shown to users, where did the content come from?

maxf:
This PR is not obsolete

PR #2256, opened 2021-10-26
Remove jQuery from javascripts/modules/sticky-element-container.js

gclssvglx:
## What

This change removes jQuery from the `sticky-element-container.js` module and replaces it with vanilla JS.

## Why

We're removing jQuery from GOV.UK.

[Trello](https://trello.com/c/10eQvjT8/275-from-jquery-from-government-frontend-sticky-element-containerjs-reasonable-js-knowledge-required)

### DIFF
--- a/app/assets/javascripts/modules/sticky-element-container.js
+++ b/app/assets/javascripts/modules/sticky-element-container.js
@@ -13,65 +13,65 @@
   Modules.StickyElementContainer = function () {
     var self = this
 
-    self._getWindowDimensions = function _getWindowDimensions () {
+    self.getWindowDimensions = function () {
       return {
         height: root.innerHeight,
         width: root.innerWidth
       }
     }
 
-    self._getWindowPositions = function _getWindowPositions () {
+    self.getWindowPositions = function () {
       return {
         scrollTop: root.scrollY
       }
     }
 
-    self.start = function start ($el) {
-      $el = $el[0]
-      var $element = $el.querySelector('[data-sticky-element]')
+    self.start = function ($el) {
+      var el = $el[0]
+      var element = el.querySelector('[data-sticky-element]')
 
-      var _hasResized = true
-      var _hasScrolled = true
-      var _interval = 50
-      var _windowVerticalPosition = 1
-      var _startPosition, _stopPosition
+      var hasResized = true
+      var hasScrolled = true
+      var interval = 50
+      var windowVerticalPosition = 1
+      var startPosition, stopPosition
 
       initialise()
 
       function initialise () {
         root.onresize = onResize
         root.onscroll = onScroll
-        setInterval(checkResize, _interval)
-        setInterval(checkScroll, _interval)
+        setInterval(checkResize, interval)
+        setInterval(checkScroll, interval)
         checkResize()
         checkScroll()
-        $element.classList.add('sticky-element--enabled')
+        element.classList.add('sticky-element--enabled')
       }
 
       function onResize () {
-        _hasResized = true
+        hasResized = true
       }
 
       function onScroll () {
-        _hasScrolled = true
+        hasScrolled = true
       }
 
       function checkResize () {
-        if (_hasResized) {
-          _hasResized = false
-          _hasScrolled = true
+        if (hasResized) {
+          hasResized = false
+          hasScrolled = true
 
-          var windowDimensions = self._getWindowDimensions()
-          _startPosition = $el.offsetTop
-          _stopPosition = $el.offsetTop + $el.offsetHeight - windowDimensions.height
+          var windowDimensions = self.getWindowDimensions()
+          startPosition = el.offsetTop
+          stopPosition = el.offsetTop + el.offsetHeight - windowDimensions.height
         }
       }
 
       function checkScroll () {
-        if (_hasScrolled) {
-          _hasScrolled = false
+        if (hasScrolled) {
+          hasScrolled = false
 
-          _windowVerticalPosition = self._getWindowPositions().scrollTop
+          windowVerticalPosition = self.getWindowPositions().scrollTop
 
           updateVisibility()
           updatePosition()
@@ -79,7 +79,7 @@
       }
 
       function updateVisibility () {
-        var isPastStart = _startPosition < _windowVerticalPosition
+        var isPastStart = startPosition < windowVerticalPosition
         if (isPastStart) {
           show()
         } else {
@@ -88,7 +88,7 @@
       }
 
       function updatePosition () {
-        var isPastEnd = _stopPosition < _windowVerticalPosition
+        var isPastEnd = stopPosition < windowVerticalPosition
         if (isPastEnd) {
           stickToParent()
         } else {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -10,77 +10,77 @@
...
-    self.start = function start ($el) {
-      var $element = $el.find('[data-sticky-element]')
-      var _hasResized = true
-      var _hasScrolled = true
-      var _interval = 50
-      var _windowVerticalPosition = 1
-      var _startPosition, _stopPosition
+    self.start = function ($el) {
+      var el = $el[0]
+      var element = el.querySelector('[data-sticky-element]')
```
andysellick, line 31: Could these two lines be combined e.g. `var element = $el[0].querySelector('[data-sticky-element]')`?

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -10,77 +10,77 @@
...
   'use strict'
 
-  var $ = root.$
-  var $window = $(root)
-
   Modules.StickyElementContainer = function () {
     var self = this
 
-    self._getWindowDimensions = function _getWindowDimensions () {
+    self.getWindowDimensions = function () {
```
andysellick, line 16: I think it would be good to investigating converting this to using the more standard module format used by the rest of our JS, e.g.

```
window.GOVUK = window.GOVUK || {}
window.GOVUK.Modules = window.GOVUK.Modules || {};

(function (Modules) {
  function StickyElementContainer ($module) {
    this.module = $module
  }

  StickyElementContainer.prototype.init = function () {
    // etc.
  }

Modules.StickyElementContainer = StickyElementContainer
})(window.GOVUK.Modules)
```

However I don't think this should block this PR in its current form - we should raise a separate PR to do this afterwards.

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -10,77 +10,77 @@
...
+
+      var hasResized = true
+      var hasScrolled = true
+      var interval = 50
+      var windowVerticalPosition = 1
+      var startPosition, stopPosition
 
       initialise()
 
       function initialise () {
```
andysellick, line 41: This wasn't your change, but I'm curious to know why this is a separate function, when it's called from the line immediately before it and nowhere else. I've tested removing the function and just having the code called immediately, and it doesn't seem to break anything.

app/assets/javascripts/modules/sticky-element-container.js
andysellick: We could also look at improving the readability of this code (not related to your changes). For example, renaming `window` to `root` caused me some initial confusion, as well as what the initialise function is actually doing and the names of the elements involved (`el` and `element` could be better as `wrapper` and `stickyElement`, for example).

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -10,77 +10,77 @@
...
     }
 
-    self.start = function start ($el) {
-      var $element = $el.find('[data-sticky-element]')
-      var _hasResized = true
-      var _hasScrolled = true
-      var _interval = 50
-      var _windowVerticalPosition = 1
-      var _startPosition, _stopPosition
+    self.start = function ($el) {
```
andysellick, line 29: The sticky functionality no longer appears to work in IE10 and IE11 (probably something to do with how those older browsers calculate height). I don't think fixing that should be a blocking issue for this PR but it would be worth noting this somewhere so that we can address it separately.

andysellick (commented):
Looks like all the jQuery's gone, just a few comments, mostly non-blocking but happy to discuss.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: It can be a single line, but there are a couple of things that I thought made this approach better:

- we need a reference in `checkResize()` to the de-jQuery'd element - repeating `$el[0]` didn't seem quite right,
- my thinking was de-reference the jQuery object once to make it clear that from now on there is no jQuery, and 
- I'm guessing that at some point we'd want to pass into the `start()` a regular DOM element - having a variable set would make that change simpler.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: I tried to make this into a module 😢 - my skills are lacking and I hit a blocker maintaining a reference to the window object in the event handlers. Happy to give it a go, but might need some assist with that aspect.

Yeah, the naming is not great and totally agree with your suggestions.  I'll change the naming commit to include them.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: Yes, curious!  I tried exactly the same, but left it to keep as much of the code the same pre- and post-jQuery removal as possible. Refactoring to a module should resolve this I think.  I'll raise a separate ticket to convert.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: 😱 oh no!  There appears to be a lot of complexities in calculating the height of elements. Again, I'll raise separate PR to address this.

app/assets/javascripts/modules/sticky-element-container.js
andysellick: Ah of course, sorry hadn't noticed that when I made this comment.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: Raised [this card](https://trello.com/c/W73JrGKm/315-fix-sticky-element-in-government-frontend-not-working-in-ie-10-11) to address this issue.

app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -7,28 +7,28 @@
   Use 'data-module="sticky-element-container"' to instantiate, and add
   `[data-sticky-element]` to the child you want to position.
 */
-(function (Modules, root) {
+(function (Modules, window) {
```
andysellick, line 10: Nice renaming. One thought - since we're passing `window` into this code and referring to it as `window`, do we need to pass it anymore? It should be globally accessible from anywhere, right?

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: Raised [this card](https://trello.com/c/XwULsSw8/316-convert-stick-element-in-government-frontend-to-follow-the-module-pattern) to address this issue.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: Ah yes, absolutely correct - thanks for picking that up 👍 Fixed up in last commit.

andysellick (approved):
Good work 👍

PR #2260, opened 2021-10-28
Remove jq from set ga client

JonathanHallam:
## What

Remove jQuery from the set ga client id file.

## Why

We're using an old and unsupported version of jQuery for browser support reasons. Rather than upgrade, it's far better to remove our dependence.

jQuery makes writing JavaScript easier, but it doesn't do anything that you can't do with vanilla JavaScript, because it's all written in JavaScript.

Once it's removed, we no longer have to worry about upgrading it, and users don't have to download the jQuery library when they visit GOV.UK.

## Trello

https://trello.com/c/bqdisKbm/277-from-jquery-from-government-frontend-set-ga-client-id-on-formjs-and-applicationjs-moderate-js-knowledge-required

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,10 +12,8 @@
 //= require_tree ./modules
 //= require set-ga-client-id-on-form
 
-jQuery(function ($) {
-  var $form = $('.js-service-sign-in-form')
+var form = document.querySelector('.js-service-sign-in-form')
 
-  if ($form.length) {
-    new GOVUK.SetGaClientIdOnForm({ $form: $form }) // eslint-disable-line no-new
-  }
-})
+if (form) {
+  new GOVUK.SetGaClientIdOnForm(form) // eslint-disable-line no-new
+}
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js
```diff
@@ -12,10 +12,8 @@
 //= require_tree ./modules
 //= require set-ga-client-id-on-form
 
-jQuery(function ($) {
-  var $form = $('.js-service-sign-in-form')
+var form = document.getElementsByClassName('js-service-sign-in-form')
```
andysellick, line 15: The tests are passing so this seems to be working but I'm slightly confused about something. Both the original jQuery and the replacement (`getElementsByClassName`) return an array-like object of elements to act on, but then this is treated as a single entity when the attributes are read and set.

I think this works in the test because the test bypasses this bit of code to pass a single form element directly to the module. Have you tested this works as expected on a locally running page in `government-frontend`? I think there might also be value in having a test that simulates more than one form on a page, to check it applies the right attribute to all forms.

JonathanHallam: Blimey, yeah I think this would have broken in prod. Just a product of the disconnect between how the thing's called in tests and production. I'll see if I can work out a way to test it better

andysellick (approved):
🚀

PR #2270, opened 2021-11-05
Remove jQuery from track-radio-group.js

gclssvglx:
## What

This change removes jQuery from the `track-radio-group.js` module and replaces it with vanilla JS

## Why

We're removing jQuery from GOV.UK

[Trello](https://trello.com/c/IoQlSiAk/276-remove-jquery-from-government-frontend-track-radio-groupjs-reasonable-js-knowledge-required)

### DIFF
--- a/app/assets/javascripts/modules/track-radio-group.js
+++ b/app/assets/javascripts/modules/track-radio-group.js
@@ -1,95 +1,101 @@
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
-(function (global, GOVUK) {
-  'use strict'
+(function (Modules) {
+  function TrackRadioGroup (element) {
+    this.element = element.querySelector('form') || element
+  }
 
-  var $ = global.jQuery
+  TrackRadioGroup.prototype.init = function () {
+    this.track(this.element)
 
-  GOVUK.Modules.TrackRadioGroup = function () {
-    this.start = function (element) {
-      track(element)
+    if (this.crossDomainTrackingEnabled(this.element)) {
+      this.addCrossDomainTracking(this.element)
+    }
 
-      if (crossDomainTrackingEnabled(element)) {
-        addCrossDomainTracking(element)
+    this.checkVerifyUser(this.element)
+  }
+
+  TrackRadioGroup.prototype.track = function (element, withHint) {
+    this.element.addEventListener('submit', function (event) {
+      var options = { transport: 'beacon' }
+      var submittedForm = event.target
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/track-radio-group.js
```diff
@@ -1,95 +1,104 @@
...
 
-        var eventValue = eventTrackingValue($checkedOption, withHint)
+      if (this.crossDomainTrackingEnabled(submittedForm)) {
+        this.trackCrossDomainEvent(submittedForm, eventValue, options)
+      }
+    }.bind(this))
+  }
 
-        GOVUK.analytics.trackEvent('Radio button chosen', eventValue, options)
+  TrackRadioGroup.prototype.checkVerifyUser = function (element) {
```
andysellick, line 42: Maybe a little comment here would be helpful to explain what this function is doing. Maybe

```
// the analytics data should include whether the user has previously signed in to Verify
// this simulates the behaviour of JSONP, returning a true or false value from Verify
```

app/assets/javascripts/modules/track-radio-group.js
```diff
@@ -1,95 +1,104 @@
...
-      $.ajax({
-        url: 'https://www.signin.service.gov.uk/hint',
-        cache: false,
-        dataType: 'jsonp',
-        timeout: 3000
-      }).then(function (data) {
-        reportVerifyUser(element, data)
-      }, function (e) { console.log('error', e) })
+  TrackRadioGroup.prototype.reportVerifyUser = function (element, data) {
+    if (data != null && data.value === true) {
```
andysellick, line 55: This is outside the scope of this PR, but I worry slightly about this line. Am I right in thinking that if Verify suddenly stopped returning anything (i.e. it returned `null`) then none of the analytics events in this module would fire anymore?

andysellick (approved):
Great work on this @gclssvglx 👏 

Couple of suggestions on making what this does a bit clearer, but otherwise looks good.

andysellick (approved):
Excellent changes and work on this @gclssvglx. Think it's ready now. We can revisit this afterwards to remove unneeded events but this looks good for now.

app/assets/javascripts/modules/track-radio-group.js
andysellick: Note for history. No, if it returned `null` then events would still fire, but the behaviour wasn't quite working as intended. 

Expected behaviour: of the two GA events in the code, only the normal one was firing on live (the cross domain one wasn't happening). When a user submits the form, one event should fire to record that, and a second should fire if the user has previously signed in to Verify. The code now does that.

PR #2271, opened 2021-11-05
Add email subscribe/unsubscribe flash messages 

alex9smith:
When a user completes the single page email subscribe or unsubscribe
journey they will be redirected to the page they were on with a message
in the account session flash.

Check for these messages in the flash and render the appropriate success
banner. The pages chosen for  phase 1 and phase 2 of the rollout are
either `publication` or `detailed_guide` content types, so only add the
partial to these templates. In the future when we roll out to more pages
we can add the partial to the rest of the content types.

Also set the vary header to indicate to Fastly that the success
messages are different variations and should be cached and served
separately. The `GovukPersonalisation::ControllerConcern` handles this
for us but sets the header to vary on the full account session. This
would result in pages being cached per-user which isn't very helpful in
this case.

Instead, override the `set_account_vary_header` method to vary on the
flash message instead. The message content won't vary per user so this
will mean we only cache three versions of the page - one without a flash
message, one with the subscribed message and one with the unsubscribed
message.

[Trello](https://trello.com/c/1Giyrqs3/1127-add-single-page-flash-messages-to-content-pages)

**Subscribe:**
Screenshot 2021-11-05 at 15 50 45

**Unsubscribe:**
Screenshot 2021-11-05 at 15 52 32


app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if @account_flash.include?("email-subscription-success") %>
+
+  <%= render "govuk_publishing_components/components/success_alert", {
+    message: "You've subscribed to emails about this page.",
+    description: sanitize("<p class=\"govuk-body\">Go to your GOV.UK account to <a class=\"govuk-link\" href=\"\">see and manage all your GOV.UK email subscriptions</a></p>")
```
alex9smith, line 5: Waiting for confirmation on where this link should go

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if @account_flash.include?("email-subscription-success") %>
+
+  <%= render "govuk_publishing_components/components/success_alert", {
+    message: "You've subscribed to emails about this page.",
```
alex9smith, line 4: I tried using the fancy apostrophe (`’`) here, but I got a `ActionView::Template::Error: "\xE2" from ASCII-8BIT to UTF-8
` error when trying to render the view. If there's an easy way to sort this, let me know otherwise I think the regular one is fine

barrucadu: Is it possible your editor saved the file with some encoding other than utf-8?  https://stackoverflow.com/a/5286382 suggests it could happen if the encoding is wrong.

Otherwise, I think `&apos;` is the HTML entity for the apostrophe we want.

alex9smith: My editor _says_ it's saving files with UTF-8. I'll try the HTML entity

alex9smith: `&rsquo;` (right single quotation mark) looks like the correct entity for this:

image

barrucadu (approved):

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= render "govuk_publishing_components/components/success_alert", {
+    message: sanitize("You&rsquo;ve subscribed to emails about this page.")  ,
+    description: sanitize("<p class=\"govuk-body\">Go to your GOV.UK account to <a class=\"govuk-link\" href=\"/email/manage\">see and manage all your GOV.UK email subscriptions</a></p>")
+  } %>
+
+<% elsif @account_flash.include?("email-unsubscribe-success") %>
+
+  <%= render "govuk_publishing_components/components/success_alert", {
+    message: sanitize("You&rsquo;ve unsubscribed from emails about this page."),
+    description: sanitize("<p class=\"govuk-body\">Go to your GOV.UK account to <a class=\"govuk-link\" href=\"/email/manage\">see and manage all your GOV.UK email subscriptions</a></p>")
```
danacotoran, line 12: For the link, could we please add the `govuk-notification-banner__link` class to make it green?
See this for reference: https://components.publishing.service.gov.uk/component-guide/success_alert/with_styled_link

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -0,0 +1,15 @@
+<% if @account_flash.include?("email-subscription-success") %>
+
+  <%= render "govuk_publishing_components/components/success_alert", {
+    message: sanitize("You&rsquo;ve subscribed to emails about this page.")  ,
```
danacotoran, line 4: I wonder if the message and description should be in a locale file?

alex9smith: Sure, I'll move them over. 

Out of interest, do we have any sort of style guide / guidance that says when page content should be in a locale file and when it's fine to be in the template? I've seen some templates where everything is in the locale files and some where it's a mixture.

alex9smith: Here's how it looks now: 
image

Also, moving to locale files has sorted out all the unicode errors I was seeing before so the content is much more readable now

barrucadu:
@alex9smith Can you rebase this so I can deploy the branch for testing?

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
danacotoran: Alex to be honest I'm not too sure there is much specific guidance around localisation on GOVUK unfortunately (I might be wrong about that though, I can do some digging). 
I tend to add to locale files anything that could potentially be translated in the future, and/or bits of text that are shared between different partials in the application.

alex9smith:
@barrucadu should be good now

alex9smith:
I've fixed most of the spacing, column width and content changes requested from design review. 

There is one outstanding which we are deciding to leave for now as it won't be a simple fix. On desktop only, there is a lot of extra spacing underneath the flash message banners: 

image


This does not appear on mobile. To fix, we'll probably need to modify the component so it can take a custom `margin_bottom` argument, release the components gem, bump the version here then add in a custom margin to reduce the spacing.

barrucadu (approved):
Looks good to me (well, other than the extra spacing, but that's a separate issue)

PR #2276, opened 2021-11-10
Convert set ga code to our module format

andysellick:

## What

Converts the set ga id on form code to use our standard JS modules format. This is used on https://www.gov.uk/check-income-tax-current-year/sign-in/prove-identity

## Why

We're trying to standardise our approach to JS and having it in our modules format means we also don't need to manually initialise this code in `application.js` anymore.

## Visual changes
None.

Trello card: https://trello.com/c/bqdisKbm/277-from-jquery-from-government-frontend-set-ga-client-id-on-formjs-and-applicationjs-moderate-js-knowledge-required


spec/javascripts/set-ga-client-id-on-form.spec.js
```diff
@@ -11,7 +11,8 @@ describe('SetGaClientIdOnForm', function () {
     form = $(
       '<form class="js-service-sign-in-form" action="/endpoint"></form>'
     )
-    new GOVUK.SetGaClientIdOnForm(form[0]) // eslint-disable-line no-new
+    var setId = new GOVUK.Modules.SetGaClientIdOnForm(form[0]) // eslint-disable-line no-new
```
kevindew, line 14: You don't need to disable this linting rule anymore 🎉

app/assets/javascripts/application.js
```diff
@@ -10,10 +10,3 @@
 //= require govuk_publishing_components/components/step-by-step-nav
 
 //= require_tree ./modules
-//= require set-ga-client-id-on-form
-
-var form = document.querySelector('.js-service-sign-in-form')
```
kevindew, line 15: I'm wondering if this class can be removed from the HTML now? From a quick search I think it was only for this feature.

kevindew (commented):
Looks great Andy ⭐  - just a couple of small comments.

app/assets/javascripts/application.js
andysellick: Of course, thanks 🤦

kevindew (approved):
🕺

Issue #2278, opened 2021-11-11
Update #000000 to #0B0C0C on "Published/last updated"


chao-xian:
Reported by @JeremyHHY 

### Live page

https://www.gov.uk/government/publications/common-travel-area-guidance#history

![image (7)](https://user-images.githubusercontent.com/424772/141298224-944bbf44-f279-4316-9ba8-af8942f33445.png)

### Highlighted

![Screenshot 2021-11-11 at 12 29 47 pm](https://user-images.githubusercontent.com/424772/141298266-48b25c4b-b183-4bcc-9d04-b063a366c297.png)

PR #2279, opened 2021-11-11
Fix font colour being black

chao-xian:
It should be #0b0c0c as per our designs.

Fixes https://github.com/alphagov/government-frontend/issues/2278

## After
Screenshot 2021-11-11 at 1 26 05 pm

### DIFF
--- a/app/assets/stylesheets/components/_published-dates.scss
+++ b/app/assets/stylesheets/components/_published-dates.scss
@@ -13,7 +13,7 @@
 }
 
 .app-c-published-dates__change-history {
-  margin: 20px 0;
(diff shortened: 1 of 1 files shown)


owenatgov (commented):
Nice! Without checking why your tests are failing, I can see you put a double semi-colon on line 4.

Also, it's not part of the PR but I can see in this component that on lines 16 and 25 we could be using `govuk-spacing` instead of px values.

I'm also having a think about no-js view for this and if that `js-enabled` is a bad way to do this but that's a problem for another day. Once the above 2 points are sorted I'm happy to approve this.

owenatgov (approved):
Boodaful

PR #2283, opened 2021-11-15
Convert sticky-element.js to use Module pattern

gclssvglx:
### What

Currently, the sticky-element in Government Frontend does not follow the JavaScript module pattern, for example...

```
window.GOVUK = window.GOVUK || {}
window.GOVUK.Modules = window.GOVUK.Modules || {};

(function (Modules) {
  function StickyElementContainer ($module) {
    this.module = $module
  }

  StickyElementContainer.prototype.init = function () {
    // etc.
  }

  Modules.StickyElementContainer = StickyElementContainer
})(window.GOVUK.Modules)
```

Convert the existing code to be consistent with this pattern.

### Why do this?

The module pattern is the standard approach and this JavaScript component does not follow the convention.

[Trello](https://trello.com/c/XwULsSw8/316-convert-sticky-element-in-government-frontend-to-follow-the-module-pattern)

Test URLs:

- https://government-f-convert-st-2govfa.herokuapp.com/government/publications/rates-and-allowances-national-insurance-contributions/rates-and-allowances-national-insurance-contributions
- https://government-f-convert-st-2govfa.herokuapp.com/guidance/red-list-of-countries-and-territories

### DIFF
--- a/app/assets/javascripts/modules/sticky-element-container.js
+++ b/app/assets/javascripts/modules/sticky-element-container.js
@@ -7,111 +7,112 @@
   Use 'data-module="sticky-element-container"' to instantiate, and add
   `[data-sticky-element]` to the child you want to position.
 */
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
 (function (Modules) {
-  'use strict'
+  function StickyElementContainer (element) {
+    this.wrapper = element
+    this.stickyElement = this.wrapper.querySelector('[data-sticky-element]')
+    this.hasResized = true
+    this.hasScrolled = true
+    this.interval = 50
+    this.windowVerticalPosition = 1
+    this.startPosition = 0
+    this.stopPosition = 0
+  }
 
-  Modules.StickyElementContainer = function () {
-    var self = this
+  StickyElementContainer.prototype.init = function () {
+    if (!this.stickyElement) return
 
-    self.getWindowDimensions = function () {
-      return {
-        height: window.innerHeight,
-        width: window.innerWidth
-      }
-    }
+    window.onresize = this.onResize
+    window.onscroll = this.onScroll
+    setInterval(this.checkResize.bind(this), this.interval)
+    setInterval(this.checkScroll.bind(this), this.interval)
+    this.checkResize()
+    this.checkScroll()
+    this.stickyElement.classList.add('sticky-element--enabled')
+  }
 
-    self.getWindowPositions = function () {
-      return {
-        scrollTop: window.scrollY
-      }
-    }
-
-    self.start = function ($el) {
-      var wrapper = $el[0]
-      var stickyElement = wrapper.querySelector('[data-sticky-element]')
-
-      var hasResized = true
-      var hasScrolled = true
-      var interval = 50
-      var windowVerticalPosition = 1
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/sticky-element-container.js
```diff
@@ -7,111 +7,110 @@
...
 */
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
 (function (Modules) {
-  'use strict'
+  function StickyElementContainer (element) {
+    this.wrapper = element
+    this.stickyElement = this.wrapper.querySelector('[data-sticky-element]')
```
andysellick, line 17: Maybe out of scope for this PR, but I think it might be worth putting a check for the existence of this element before proceeding in the `init` function e.g. `if (!this.stickyElement) return`

spec/javascripts/modules/sticky-element-container.spec.js
```diff
@@ -2,62 +2,72 @@ describe('A sticky-element-container module', function () {
...
+    beforeEach(function () {
+      $element = $(
+        '<div data-module="sticky-element-container" style="height: 9001px; margin-bottom: 1000px">' +
+          '<div data-sticky-element>' +
+            '<span>Content</span>' +
+          '</div>' +
+        '</div>'
+      )
+
+      instance = new GOVUK.Modules.StickyElementContainer($element[0])
```
andysellick, line 20: Interesting that these tests don't appear to need to call the `init` function, yet presumably are still passing? That seems odd.

andysellick: The original code called the `start` function (line 33). Can we add the init function into this test? I'm concerned that if we're not calling it we're not testing it properly somehow.

spec/javascripts/modules/sticky-element-container.spec.js
```diff
@@ -2,62 +2,72 @@ describe('A sticky-element-container module', function () {
...
+        return {
+          scrollTop: 0
         }
-      })
+      }
 
-      it('hides the element, when scrolled at the top', function () {
-        instance.start($element)
+      instance.checkResize()
+      instance.checkScroll()
```
andysellick, line 39: Why do these two functions now need to be called where they weren't previously? I can see this in a few of the tests - is this related to not calling the `init` function?

andysellick (commented):
Generally looks good, some questions about the tests though.

app/assets/javascripts/modules/sticky-element-container.js
gclssvglx: Added great spot - thanks Andy

spec/javascripts/modules/sticky-element-container.spec.js
gclssvglx: Yeah, this and your previous comment are related to the `init()` function.  What I found was that the tests do not test the actual scrolling behaviour - in essence they just test some basic maths and the ability to add/remove classes on an element.  I've tested both with and without the `init()` function in the `beforeEach()` and the behaviour is exactly the same.  Which leads me to believe it has no value being run in the tests.

Probably out of scope of this PR - but should we create some tests that actually do test the physical scrolling behaviour?

andysellick (approved):

PR #2284, opened 2021-11-15
Update get involved

chris-gds:
## What

Update [get involved](https://www.gov.uk/government/get-involved) page with redesign

## Why

[Apply publishing gem components](https://components.publishing.service.gov.uk/component-guide/) + [Design System](https://design-system.service.gov.uk/) to align architecture, improve consistency and in turn improve accessibility 

## Visuals

<details>
 <summary>Provided Design</summary>
Redesigned_Get_involved
</details>

| Mobile      | Desktop |
| ----------- | ----------- |
| government-frontend dev gov uk_government_get-involved(iPhone 6_7_8) (1) |  government-frontend dev gov uk_government_get-involved (1)  |

### Anything else

- The "blue pull out" on provided design doesn't exist - this has been replaced with [inset component
](https://components.publishing.service.gov.uk/component-guide/inset_text) as this component is most suitable based on context and properties
- Lead paragraph text maintains inline anchor links from original page.
- "Some active government blogs include" have a selection of links that appear to be a type of "document list" this was not appropriate so was replaced for a link list
- "Take Part" in the original design has introduction copy, this doesn't exist and has been removed.
- Two images in the image grid are missing `alt` copy "National Service" "Help Making your neighbour.." these have been requested to be updated via Content.
- The original design omits the image grid, this has been kept at `column-full-width` inline with other pages.
- [The image grid uses a solution that is related to an layout feature request on the design system](https://github.com/alphagov/govuk-frontend/issues/2328)
- The image grid section has been updated to be [more inline with other instances](https://www.gov.uk/world/uk-and-the-commonwealth)

### Render page locally

`content-store`, `search-api`, `static` have to be run in parallel
`government-frontend` has to be started in `-app` mode not `app-live` mode as local data is needed

Recent local data is needed for `content-store` and `search-api` - this data has to be pulled from AWS so its gets recent snapshot - eg:
`gds aws govuk-integration-poweruser ./bin/replicate-elasticsearch.sh`
`gds aws govuk-integration-poweruser ./bin/replicate-mongodb.sh content-store`

I faced an issue where data from `search-api` was failing to download correctly from AWS.

The workaround is to delete the parent folder `govuk-docker/replication/elasticsearch-6`
Once this is done each app has to be re made using `make` (as the existing containers will still reference old data)


app/views/content_items/get_involved.html.erb
```diff
@@ -1,155 +1,229 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds govuk-!-padding-bottom-8">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t('get_involved.recently_opened'),
+      padding: true,
+      border_top: 2,
+      font_size: "s",
+      margin_bottom: 4
+    } %>
+    <%= render "govuk_publishing_components/components/document_list", {
```
jon-kirwan, line 95: Not an issue but I remember that with the recent work on [topical events](https://www.gov.uk/government/topical-events/cop26) and [world location news](https://www.gov.uk/world/france/news) we have a similar pattern for the heading and document list components but with slightly different formatting -
```
<div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-7">
  <div id="recently-updated">
    <%= render "govuk_publishing_components/components/heading", {
      text: t('latest_feed.title'),
      padding: true,
      border_top: 2,
    } %>
```
In the above example, the spacing is set to `govuk-!-margin-bottom-7` instead of `govuk-!-margin-bottom-8` and the heading component (despite being a h3) has the default size and no `margin_bottom`. Just wondering if it's worth using the same style properties or not?

app/controllers/get_involved_controller.rb
```diff
@@ -88,4 +90,46 @@ def retrieve_consultation_outcomes
...
+          path: item["link"],
+          description: "#{close_status} #{date_microformat(item['end_date'])}",
+        },
+        metadata: {
+          public_updated_at: Time.zone.parse(org_time),
+          document_type: org_acronym,
+        },
+      }
+    end
+  end
```
jon-kirwan, line 134: Again just a suggestion but you could refactor the filtered_links to move the `.map` operation stuff in to it's own private method e.g.
```
  def filtered_links(array, close_status)
    array.map do |item|
      {
        link: {
          text: item["title"],
          path: item["link"],
          description: "#{close_status} #{date_microformat(item['end_date'])}",
        },
        metadata: {
          public_updated_at: Time.zone.parse(map_property(org["public_timestamp"])),
          document_type: map_property(org["acronym"]),
        },
      }
    end
  end

private

  def map_property(property)
    item["organisations"].map { |org|
      org[property]
    }.join(", ")
  end
end
```

jon-kirwan:
👍 Looking good to me. I've just mentioned a couple of minor things and added one or two questions / suggestions.

app/views/content_items/get_involved.html.erb
chris-gds: > Just wondering if it's worth using the same style properties or not?

Agreed, this should match as close as possible.  I've updated the Design to [match similar to this page.](https://www.gov.uk/world/uk-and-the-commonwealth)

----

Side thought:

These image grid patterns are making me re-think override use eg `govuk-!-margin-bottom-7`.   How can we make sure all these instances match across repos?  

Maybe this `image_card` is slowly starting to qualify as a component variation: 

```
.gem-c-image-card__image-grid {
 @include govuk-responsive-padding(6);
}
```

app/controllers/get_involved_controller.rb
chris-gds: Like it - updated

app/views/content_items/get_involved.html.erb
jon-kirwan: Yep sounds good. The utility styles are great for one off overrides but if looks like we starting to use this pattern in quite a few places it might be time to add that variant.

app/assets/stylesheets/application.scss
```diff
@@ -25,6 +26,7 @@ $govuk-new-link-styles: true;
 @import 'govuk_publishing_components/components/govspeak-html-publication';
 @import 'govuk_publishing_components/components/heading';
 @import 'govuk_publishing_components/components/hint';
+@import 'govuk_publishing_components/components/inset-text';
 @import 'govuk_publishing_components/components/image-card';
 @import 'govuk_publishing_components/components/input';
 @import 'govuk_publishing_components/components/intervention';
```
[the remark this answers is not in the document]
jon-kirwan: Sorry Chris, it looks like the order is still the same as previously?

jon-kirwan (approved):
Yep great! Just one small comment but otherwise all fine so have approved.

app/assets/stylesheets/application.scss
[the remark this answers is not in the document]
chris-gds: My fault - sorted!

app/views/content_items/get_involved.html.erb
chris-gds: 🤔   - i'll capture this one on the gem to revisit

maxgds:
I think the decision you've made is fine as is, but I think the blue box is an older iteration of the notice component:
https://components.publishing.service.gov.uk/component-guide/notice Let me know if you want to switch that in instead, otherwise I'll push this tomorrow.

chris-gds:
> I think the blue box is an older iteration of the notice component:
https://components.publishing.service.gov.uk/component-guide/notice Let me know if you want to switch that in instead, otherwise I'll push this tomorrow.

So, I considered this, however the when I was looking at the [Design System guidance](https://design-system.service.gov.uk/components/notification-banner/):

> Use a notification banner to tell the user about something they need to know about, but that’s **not** directly related to the page content.

I felt this _was_ directly related to the page in this context.  I didn't quite think this was _quite_ suitable even though visually it appeared to be close to the Design.

Also the guidance says

>If the information is directly relevant to the thing the user is doing on that page, put the information in the main page content instead. Use inset text or warning text if it needs to stand out.

The warning text didn't seem suitable however, looking at the [inset guidance](https://design-system.service.gov.uk/components/inset-text/) (below) 
this felt more apt in this context

> Use the inset text component to differentiate a block of text from the content that surrounds it, for example:
quotes
examples
additional information about the page

maxgds:
I think the page has so much going on you could make the argument that going to find out about a closing thing is sufficiently different to merit it - but I can see your argument and I'm not going to push for the change. I'll merge now.

brucebolt:
@chris-gds It appears these changes have caused smokey to fail (see [the latest run](https://deploy.blue.production.govuk.digital/job/smokey/10042/console)).  Are you able to apply a fix to the smokey tests?

(Also paging @Tetrino who added the tests and may be able to give some more context)

chris-gds:
Hey @brucebolt thanks (I don't have access to prod) currently reviewing the smokey repo. Will liaise w/ @Tetrino

chris-gds:
Related `smokey` testing update: https://github.com/alphagov/smokey/pull/871

PR #2286, opened 2021-11-16
Add the single page notification button to publication template

danacotoran:
This adds the [Single page notification button](https://components.publishing.service.gov.uk/component-guide/single_page_notification_button) to the `publication` template, after the metadata blocks at the top and bottom of the page.

We only want to launch with a couple of pages in the initial stages, therefore the button will initially show up only on the pages listed on `in_single_page_notifications_trial?` under `content_items_controller.rb`

| Top  | Bottom |
| ------------- | ------------- |
|Screenshot 2021-11-18 at 12 19 40 | Screenshot 2021-11-18 at 12 19 51 |
| Screenshot 2021-11-18 at 12 19 08 |Screenshot 2021-11-18 at 12 19 25 |

------

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -134,6 +134,10 @@ private
       return
     end
 
+    # use these and `@content_item.base_path` in the template
+    @notification_button_visible = in_single_page_notifications_trial?
+    @include_single_page_notification_button_js = account_session_header.present?
+
     request.variant = :print if params[:variant] == "print"
 
     respond_to do |format|
@@ -205,6 +209,13 @@ private
   def set_account_vary_header
     # Override the default from GovukPersonalisation::ControllerConcern so pages are cached on each flash message
     # variation, rather than caching pages per user
-    response.headers["Vary"] = [response.headers["Vary"], "GOVUK-Account-Session-Flash"].compact.join(", ")
+    response.headers["Vary"] = [response.headers["Vary"], "GOVUK-Account-Session-Exists", "GOVUK-Account-Session-Flash"].compact.join(", ")
+  end
+
+  def in_single_page_notifications_trial?
+    %w[
+      /government/publications/open-standards-for-government
+      /government/publications/identity-proofing-and-verification-of-an-individual
+    ].include? @content_item.base_path
   end
 end
--- a/app/views/components/_published-dates.html.erb
+++ b/app/views/components/_published-dates.html.erb
@@ -4,10 +4,13 @@
   history = Array(history)
   last_updated ||= false
   link_to_history ||= false
-  history_class = "app-c-published-dates--history" if history.any?
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  classes = %w(app-c-published-dates)
+  classes << "app-c-published-dates--history" if history.any?
+  classes << shared_helper.get_margin_bottom if local_assigns[:margin_bottom]
 %>
 <% if published || last_updated %>
-<div class="app-c-published-dates <%= history_class %>" <% if history.any? %>id="history" data-module="gem-toggle"<% end %> lang="en">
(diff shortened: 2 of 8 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -134,6 +134,10 @@ def render_template
       return
     end
 
+    # use these and `@content_item.base_path` in the template
+    @notification_button_visible = in_single_page_notifications_trial?
```
huwd, line 138: These could work as helpers, but I think this is good for now?

huwd (approved):

app/views/components/_published-dates.html.erb
```diff
@@ -4,10 +4,13 @@
   history = Array(history)
   last_updated ||= false
   link_to_history ||= false
-  history_class = "app-c-published-dates--history" if history.any?
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
```
andysellick, line 7: This is interesting - I'd never imagined the shared helper being used outside of the components gem. My initial reaction was that this isn't a good idea, but actually I can't think of a reason why. Might be interesting to have a chat about this.

app/views/components/docs/published-dates.yml
```diff
@@ -40,3 +40,9 @@ examples:
       - display_time: 14th October 2000
         note: Updated information on pupil premium reviews and what information schools need to publish on their websites.
         timestamp: 2000-10-14T15:42:37.000+00:00
+  with_custom_margin_bottom:
+    description: |
+      The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale).
```
andysellick, line 45: What margin does the component default to?

app/views/content_items/publication.html.erb
```diff
@@ -53,8 +59,15 @@
       <%= render 'components/published-dates', {
         published: @content_item.published,
         last_updated: @content_item.updated,
-        history: @content_item.history
+        history: @content_item.history,
+        margin_bottom: @notification_button_visible ? 3 : nil
```
andysellick, line 63: Am I right in that this is the section at the bottom of the page? I've just been playing with the margin bottom and it seems that this check isn't necessary - if the notification button isn't present, having this margin bottom or not on the published dates component doesn't change the layout anyway, as the margin is the same as the margin already applied by the wrapping `responsive-bottom-margin` element.

app/views/components/docs/published-dates.yml
danacotoran: It defaults to having no `margin-bottom`.

app/views/content_items/publication.html.erb
danacotoran: Yeah you're correct about that, good spot. Will update

app/views/components/_published-dates.html.erb
danacotoran: We talked about it on Slack but just to briefly document what was said: 
Your concern was based on isolation - the applications usually import the components (like I'm using the shared helper) but they don't do it in a way that depends upon their internal structure.

I mentioned there is a precedent to the `shared_helper` being used outside the components gem in other frontend apps such as [finder-frontend](https://github.com/alphagov/finder-frontend/blob/919a38385d5ce526bb4b33449e6cf591ce55db38/app/views/components/_expander.html.erb#L7) and [collections](https://github.com/alphagov/collections/blob/0ef7d799302dfdd6a603f2ddcef130b9f10e2398/app/views/components/_signup-link.html.erb#L10).


We agreed it seems like something that should be raised for discussion at the GOVUK frontend community meeting.

app/views/components/docs/published-dates.yml
danacotoran: Have updated the doc to specify that

andysellick (approved):

barrucadu (approved):

PR #2238, opened 2021-10-05
Add "overview:" to titles where an attachment has the same name

maxgds:
## What
We currently have live pages that have attachments that share the exact same title for a number of publications and consultations. This can make for confusing user journeys and in the case of HTML attachments can make it confusing for users of assistive technology navigating between the pages and can disorientate those with cognitive problems. Our solution to the issue to to adjust the parent page title in cases where this happens and add the word "overview". This will affect both the HTML page title (as seen in the browser tab), and the H1 for affected pages. In unaffected pages the H1 contains the name of the document with supporting document type context above it. Where a page is affected that contextual text has "overview:" appended and it is moved inside the H1 tag.  

https://trello.com/c/Z3D4CElD/841-update-heading-and-title-pattern-for-publication-and-consultation-pages

### Before
image

### After
image

## Examples
Some affected sample urls used during testing:

https://www.gov.uk/government/publications/warm-home-discount-better-targeted-support-from-2022-cfp-response-to-the-beis-consultation
https://www.gov.uk/government/government/publications/impact-assessment-of-the-mental-capacity-amendment-act-2019
https://www.gov.uk/government/publications/warm-home-discount-better-targeted-support-from-2022-cfp-response-to-the-beis-consultation
https://www.gov.uk/government/government/publications/impact-assessment-of-the-mental-capacity-amendment-act-2019
https://www.gov.uk/government/publications/economic-crime-research-strategy-home-office-research-priorities
https://www.gov.uk/government/publications/civil-procedure-rule-committee-annual-report-2011
https://www.gov.uk/government/publications/national-probation-service-england-and-wales-divisions-map
https://www.gov.uk/government/publications/electric-vehicle-homecharge-scheme-authorised-installers
https://www.gov.uk/government/publications/grade-and-salary-bands
https://www.gov.uk/government/publications/armed-forces-pay-review-body-fiftieth-report-2021
https://www.gov.uk/government/publications/agreement-establishing-an-association-between-the-uk-and-central-america-ms-no322019
https://www.gov.uk/government/publications/notice-143-a-guide-for-international-post-users
https://www.gov.uk/government/publications/general-medical-council
https://www.gov.uk/government/publications/government-functional-standard-govs-001-government-functions
https://www.gov.uk/government/publications/spending-review-2021-launch-letter
https://www.gov.uk/government/publications/staying-legal-heavy-goods-vehicle-drivers
https://www.gov.uk/government/publications/draft-online-safety-bill

[...]

PR #2292, opened 2021-11-18
Fix variable scope of context boolean

maxgds:
In https://github.com/alphagov/government-frontend/pull/2238 we added the word overview to format titles where an attachment shared the name of the page. The intent was for this to appear inside the H1, using the context_inside parameter. Unfortunately a refactor after review meant that the variable was scoped too tightly and never available to the component call. This resolves that by initially setting the variable outside of the each loop, and adds extra testing to ensure it is there when necessary.

Sample affected url: https://www.gov.uk/government/consultations/call-for-evidence-to-inform-orbital-liability-and-insurance-policy

### Before:
image

### After:
image


jon-kirwan (approved):
Yep, works fine for me. Double checked against the original list of sample addresses in #2238 and all ok there too. 👍

PR #2302, opened 2021-11-25
Update publication pages

chris-gds:
## What

Apply the `context_inside` option to the [title component](https://components.publishing.service.gov.uk/component-guide/title) to all publications, remove the word "overview" yet keep translations.

## Why

Recently some work made the 'grey text' of publication pages `h1`.  This provided various benefits including for screen readers and tabs.  However, we only did this for pages where the title of the page and attachment are the same. For consistency this change applies to all publication pages.  So that we don't have some pages that operate in one way, and others that do it differently. Also it adds benefits.

## Visuals

image

## Anything else

Checked against:
http://government-frontend.dev.gov.uk/government/consultations/appraisal-periods-consultation

Add context title
`Consultation outcome: Appraisal periods consultation`

http://government-frontend.dev.gov.uk/government/publications/information-superiority-jdn-213

Adds withdrawn and keep context title
`[Withdrawn] Guidance: Information superiority: more than managing information`

http://government-frontend.dev.gov.uk/government/organisations/department-of-health-and-social-care/about

No context title
`About us - Department of Health and Social Care`

http://government-frontend.dev.gov.uk/government/statistical-data-sets/cw010-proportion-of-residents-walking-or-cycling-at-least-once-a-month

Adds withdrawn and no context title (as not a publication or consultation)
`[Withdrawn] Proportion of residents who do any walking or cycling (at local authority level) (CW010)`

### Related PRs

https://github.com/alphagov/government-frontend/pull/2238/
https://github.com/alphagov/government-frontend/pull/2292

### DIFF
--- a/app/presenters/content_item/withdrawable.rb
+++ b/app/presenters/content_item/withdrawable.rb
@@ -5,7 +5,15 @@ module ContentItem
     end
 
     def page_title
-      withdrawn? ? "[Withdrawn] #{title}" : title
+      if withdrawn? && context_title? && publication_overview?
+        "[Withdrawn] #{context_title}: #{title}"
+      elsif withdrawn?
+        "[Withdrawn] #{title}"
+      elsif context_title? && publication_overview?
+        "#{context_title}: #{title}"
+      else
+        title
+      end
     end
 
     def withdrawal_notice_component
@@ -21,12 +29,31 @@ module ContentItem
 
   private
 
+    def context_title?
+      context_title.present?
+    end
+
+    def publication_overview?
+      publication_overview.present?
+    end
+
+    def publication_overview
+      overview = (I18n.exists?("content_item.#{schema_name}") == "publication" || "statistical_data_set")
+      overview.presence
+    end
+
     def withdrawal_notice
       content_item["withdrawn_notice"]
     end
 
     def withdrawal_notice_title
       "This #{withdrawal_notice_context.downcase} was withdrawn on #{withdrawal_notice_time}".html_safe
+    end
+
+    def context_title
+      if I18n.exists?("content_item.schema_name.#{document_type}", count: 1, locale: :en)
+        I18n.t("content_item.schema_name.#{document_type}", count: 1, locale: :en)
+      end
     end
 
     def withdrawal_notice_context
--- a/app/views/content_items/_context_and_title.html.erb
+++ b/app/views/content_items/_context_and_title.html.erb
@@ -1,26 +1,7 @@
-<%
-  context_string =  t("content_item.schema_name.#{@content_item.document_type}", count: 1);
-  context_inside = false;
- %>
-
-<% @content_item&.featured_attachments.each do |fa| %>
-  <% return if !@content_item.attachment_details(fa).present? %>
-  <% if @content_item.attachment_details(fa)['title'] == @content_item.title %>
-      <% content_for :title do %>
-        <%= t("content_item.schema_name.#{@content_item.document_type}.overview", count: 1) %>: <%= @content_item.title %>
-      <% end %>
-      <%
-         context_string =  t("content_item.schema_name.#{@content_item.document_type}.overview", count: 1) << ":"
-         context_inside = true
-      %>
-      <% break %>
-    <% end %>
-<% end %>
-
 <%= render 'govuk_publishing_components/components/title',
-   context: context_string,
-   context_locale: t_locale_fallback("content_item.schema_name.#{@content_item.document_type}", count: 1),
-   context_inside: context_inside,
+   context: t("content_item.schema_name.#{@content_item.document_type}", count: 1) << ":",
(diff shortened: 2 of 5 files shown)


app/views/content_items/_context_and_title.html.erb
```diff
@@ -1,26 +1,7 @@
...
-      <% end %>
-      <%
-         context_string =  t("content_item.schema_name.#{@content_item.document_type}.overview", count: 1) << ":"
-         context_inside = true
-      %>
-      <% break %>
-    <% end %>
-<% end %>
-
 <%= render 'govuk_publishing_components/components/title',
```
maxgds, line 1: I think we've lost something we want to keep here - line 9-11 adds both the document type and "overview" to the html title tag. While we're dropping the word overview, the title tag wants to line up with the H1 and include the document type. eg https://www.gov.uk/government/publications/national-probation-service-england-and-wales-divisions-map would want a title of:
Map: Probation Service England and Wales regional maps - GOV.UK

app/presenters/content_item/withdrawable.rb
```diff
@@ -5,7 +5,7 @@ def withdrawn?
     end
 
     def page_title
-      withdrawn? ? "[Withdrawn] #{title}" : title
+      withdrawn? ? "[Withdrawn] #{withdrawal_context_title}: #{title}" : title
```
maxgds, line 8: Good spot in recognising we need to fix the withdrawn content. Unfortunately we have a couple of issues around this change - 
1) It's caused a failing test on fatality notices, usually you'd want to update that test BUT see 2.
2) This is being applied to all content types and government-frontend serves content that is a different format from "publications" and we're scoping this title change to just publications and consultations. Publication subtypes are all the ones updated in the en.yml to have an "overview" entry, or you can see them in whitehall publisher. There is probably something we can query in the content item to determine whether it's a publication - we should make that check here and only update the title if it's one we want to affect. I think this is because only publications and consultations can have attachments, but Tobi will have more background otherwise.

app/views/content_items/_context_and_title.html.erb
```diff
@@ -1,26 +1,12 @@
...
-      <% end %>
-      <%
-         context_string =  t("content_item.schema_name.#{@content_item.document_type}.overview", count: 1) << ":"
-         context_inside = true
-      %>
-      <% break %>
-    <% end %>
+<% content_for :title do %>
+  <%= t("content_item.schema_name.#{@content_item.document_type}", count: 1) %>: <%= @content_item.title %>
+   <% break %>
```
maxgds, line 3: The break was initially there to jump out of the each loop once we found a match, since we're no longer in a loop I don't think it's needed.

maxgds (changes requested):
Couple of things look to need addressing, one simple and one more challenging.

app/presenters/content_item/withdrawable.rb
chris-gds: > This is being applied to all content types and government-frontend serves content that is a different format from "publications" and we're scoping this title change to just publications and consultations. Publication subtypes are all the ones updated in the en.yml to have an "overview" entry, or you can see them in whitehall publisher. 

So, I was trying to work out what this was scoped to define "publications and consultations" and you've mentioned this defined by the existence of the `overview` entry.  So the logic has been based of this.


> There is probably something we can query in the content item to determine whether it's a publication - we should make that check here and only update the title if it's one we want to affect. I think this is because only publications and consultations can have attachments, but Tobi will have more background otherwise.

I've used the `overview` property in this instance. Hopefully that's correct.

maxgds: That should work but it feels a bit dirty/manual it feels ok to check for a translation and display it based on whether it exists, but less so to base other template decisions on it. The content item seems to have an entry of
```schema_name: "publication",```
regardless of the sub type, if you can hook into that it would be cleaner.

Comment (chris-gds) [reply app/presenters/content_item/withdrawable.rb]: 🙌🏼 Ace, that works - updated.

File: app/presenters/content_item/withdrawable.rb [hunk 4]
```diff
@@ -21,6 +29,19 @@ def withdrawal_notice_component
...
+    def context_title?
+      context_title.present?
+    end
+
+    def publication_overview?
+      publication_overview.present?
+    end
+
+    def publication_overview
+      overview = (I18n.exists?("content_item.#{schema_name}") == "publication" || "statistical_data_set")
```
maxgds, line 41: Could you expand a little on what's going on here to help me understand? I have two questions:

1) Why is statistical_data_set  relevant? It's not a sub type of publication so we never expect to apply this overview/context title behaviour to it.
2) Why are you looking up whether there is a translation for the schema_name ? Don't we want to always apply overview if it's a publication?

It feels like this could be as simple as line 32 - 34 being

```
    def context_title?
      content_item.#{schema_name} == "publication"
    end
    ```
and lines 36-44 being removed, but I expect I'm missing something.

hannako:
Closing as this PR has been open 4 years!

PR #2309, opened 2021-12-08
Revert PRs #2238 and #2292

maxgds:
## What
https://trello.com/c/qZP9fwF2/958-revert-overview-work
This reverts work where we added the word "overview" to titles of publications and consultations if they had an attachment with a matching title.

## Why
We plan to revisit this work at a later date as the aim remains to find a way to disambiguate landing pages and attachment pages both to support assistive tech, but also listings such as search engine or Content Data listings. The applied solution has the following problems:

- It means landing pages for publications and consultations are not consistent which has been negatively received by publishers.
- The word "overview" is disliked so we need a better term
- We did not apply this change to withdrawn content
- The titles with the context string pre-pended are getting extremely long in some cases


jon-kirwan (approved):
Just one very small thing to mention.

The key for 'overview' still exists in the other locale files e.g. `fr.yml` (which was added here: https://github.com/alphagov/government-frontend/commit/12288567347ddfa53abc9e338f758d6e7f9d8d93).

I've approved as I think it's fine to leave as it is, and remove it later on when the `missing_keys` task is run but just wanted to double check?

maxgds:
@jon-kirwan good spot on the translation files. I figured out how to use the i18n gem and did an update to remove unused translations. Could you have a quick look over the last commit to sense check it? Tip - use the cog at the top of the diff view to ignore whitespace changes as it's tidied up a lot.

jon-kirwan:
> @jon-kirwan good spot on the translation files. I figured out how to use the i18n gem and did an update to remove unused translations. Could you have a quick look over the last commit to sense check it? Tip - use the cog at the top of the diff view to ignore whitespace changes as it's tidied up a lot.

All looks fine to me. I've scanned through all the changes and I see lots of occurrences of 'overview' removed, and no other changes that should / or shouldn't be there, so it's a 👍 from me (good tip on the whitespace option by the way. Makes it much easier to scan through the diffs)

PR #2324, opened 2022-01-05
Turn scroll tracking on for various pages

andysellick:

## What

Turn on scroll tracking for various pages.

**NOTE** this change depends on [this change](https://github.com/alphagov/govuk_publishing_components/pull/2548) in the components gem being deployed to `static` at the same time, otherwise duplicate or no scroll tracking events may occur.

## Why
We're retiring the old scroll tracker and migrating all pages that use it to the new one.

## Visual changes
None.

Trello card: https://trello.com/c/sdugtMbX/37-migrate-to-new-scroll-tracker

### DIFF
--- a/app/views/content_items/guide.html.erb
+++ b/app/views/content_items/guide.html.erb
@@ -4,6 +4,21 @@
     canonical_url: @content_item.canonical_url,
     body: @content_item.has_parts? ? @content_item.current_part_body : nil
   ) %>
+
+  <%
+    scroll_track_headings_paths = [
+      "/council-housing",
+      "/driving-abroad",
+      "/get-a-passport-urgently",
+      "/driving-abroad/international-driving-permit",
+      "/taking-goods-out-uk-temporarily/get-an-ata-carnet",
+    ]
+
+    full_url = [@content_item.base_path, @content_item.part_slug].compact.join('/')
+  %>
+  <% if scroll_track_headings_paths.include?(full_url) %>
+    <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings"/>
+  <% end %>
 <% end %>
 
 <% unless @content_item.requesting_a_part? %>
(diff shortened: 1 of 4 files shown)


app/views/content_items/guide.html.erb
```diff
@@ -4,6 +4,23 @@
...
+      "/settled-status-eu-citizens-families",
+      "/council-housing",
+      "/family-permit",
+      "/driving-abroad",
+      "/get-a-passport-urgently",
+      "/driving-abroad/international-driving-permit",
+      "/taking-goods-out-uk-temporarily/get-an-ata-carnet",
+    ]
+
+    full_url = [@content_item.base_path, @content_item.part_slug].join('/')
```
kevindew, line 19: I'm not sure this will operate correctly, given the paths you have listed above. 

I assume we're expecting part_slug to sometimes be nil in order to have paths like "/council-housing", however if we do `["/council-housing", nil].join("/")` we get `/council-housing/`. Perhaps you want to do `[@content_item.base_path, @content_item.part_slug].compact.join('/')` ?

kevindew (commented):
Nice one! looks a big improvement on the masses of configuration in: https://github.com/alphagov/govuk_publishing_components/pull/2548/files

I think I've caught a small issue with guides, otherwise I've done a quick skim the paths match the source PR but I've not been too thorough.

app/views/content_items/guide.html.erb
andysellick: Great spot, thanks! Have updated.

andysellick: The trello card links to a spreadsheet of all these URLs and some more details about them, including what the original tracking was and what changes have been made. Each one falls into one of these categories:

- not needed and scroll tracking has (probably already) been removed
- was tracking headings but content has changed so no tracking was occurring, so have removed tracking
- was tracking headings or percentages so has been migrated to the new scroll tracker

kevindew (approved):

PR #2327, opened 2022-01-06
Revert "Force English for metadata"

callumknights:
This reverts commit 20171564e8e23470ad203021ecdf4ae87e13316a.

Relevant keys have now been translated to this change can be reverted.

Trello: https://trello.com/c/4rRSo056/2794-re-enable-translated-dates-3

image

image

### DIFF
--- a/app/assets/stylesheets/components/_published-dates.scss
+++ b/app/assets/stylesheets/components/_published-dates.scss
@@ -1,5 +1,4 @@
 .app-c-published-dates {
-  direction: ltr;
   line-height: 1.45em;
   color: govuk-colour("black");
 }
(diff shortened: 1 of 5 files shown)


ChrisBAshton (commented):
Looks good so far, @callumknights 👍 just one merge conflict that needs looking at before I can approve.

maxgds (changes requested):
Hey, so looking at the screenshot, I assume that's for with the fix in place? It's not good enough because we're falling back to English for the values against the keys, which then shifts them out of the flow. A user reading right to left is going to encounter "British Embassy Beirut" ahead of whatever the key says. And the date looks messy too - I've a feeling that effectively reads as "92 [month] 9102" and also comes ahead of the key that probably says "updated". I suggest you raise an issue on this repo saying that we have the key translations now and this needs further work, and maybe copy this comment across. But I'd suggest this isn't a good thing to merge.

PR #2328, opened 2022-01-07
Track only specific headings

andysellick:

## What

Applies to https://www.gov.uk/guidance/travel-abroad-from-england-during-coronavirus-covid-19 and https://www.gov.uk/guidance/travel-to-england-from-another-country-during-coronavirus-covid-19

- for the given URLs, only scroll track some of the headings, not all of the headings
- this will need to be updated in future if the headings change

## Why
We only want to scroll track these headings. Anything else is just adding unneeded data.

## Visual changes
None.

Trello card: https://trello.com/c/sdugtMbX/37-migrate-to-new-scroll-tracker

### DIFF
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -12,8 +12,6 @@
       "/guidance/healthcare-for-eu-and-efta-citizens-visiting-the-uk",
       "/guidance/guidance-for-suppliers-of-cattle-sheep-and-goat-ear-tags",
       "/guidance/ecmt-international-road-haulage-permits",
-      "/guidance/travel-abroad-from-england-during-coronavirus-covid-19",
-      "/guidance/travel-to-england-from-another-country-during-coronavirus-covid-19",
     ]
     scroll_track_percent_paths = [
       "/guidance/brexit-guidance-for-businesses",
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Changes look good - have you tried it out on integration? just want to confirm that the HTML output all wires up correctly.

It'd be useful to have some context in the commit message (and PR in general) about why we need to do this complex and fragile approach for these pages? I'm assuming there's a problem of some sort in what you had before and that this is probably a non-ideal compromise?

andysellick:
Thanks @kevindew have updated the commit message to add some more detail.

I've just tried it on integration and it seems fine. Not sure what you meant by the HTML output wiring up correctly, but here's the view source, seems to be all in one line and fairly clean.

Screenshot 2022-01-07 at 13 48 25

kevindew:
> I've just tried it on integration and it seems fine. Not sure what you meant by the HTML output wiring up correctly, but here's the view source, seems to be all in one line and fairly clean.

Ah I just meant that it all worked as expected.

Thanks for adding the extra detail. Hmm has someone complained about this? It doesn't seem too big a deal to me to be sending unnecessary data to GA - certainly better that than we add extra fragile code where tracking breaks when the content does

Looking at: https://www.gov.uk/guidance/travel-abroad-from-england-during-coronavirus-covid-19 there are 17 headers in main and we're tracking 9 so it's not much of a difference for the overhead. The other one seems to have 30 headers - but still seems a reasonable quantity to track.

andysellick:
I was asked to track only those specific headings because I think although the difference between those tracked and the total number is small, it amounts to quite a lot of unneeded data when you multiply up all the visits - and these are some of our most visited pages at the moment. Agreed though, this is a fragile approach and one I'll try to discourage.

kevindew (approved):
Cool - well sounds reasonable and I know this is only caught in the midst of a migration and not a new change.

PR #2350, opened 2022-01-28
Add single page notification button to `consultation` template

danacotoran:
Add support for the single page notification button to the `consultation`
template. The button will not appear until the feature is enabled on
pages with this content type.

Seeing as the button is always grouped together with the published dates
component I created a new  `published_dates_with_notification_button`
partial to be shared between `publication`, `detailed_guide` and
`consultation`.

https://trello.com/c/iqhyMnMP

### DIFF
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -196,11 +196,7 @@
         %>
       </div>
 
-    <%= render 'components/published-dates', {
-        published: @content_item.published,
-        last_updated: @content_item.updated,
-        history: @content_item.history
-      } %>
+      <%= render 'shared/published_dates_with_notification_button' %>
     </div>
   </div>
   <%= render 'shared/sidebar_navigation' %>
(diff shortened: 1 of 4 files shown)


huwd (commented):
nice!
Non blocking but a thought, we're using instance variables here in a partial.
Do we / should we prefer to pass local variables here instead?

Pro: as instance variables are set in the controller, they'll fail if we ever try to include this partial into a view that's spun up by a different controller. So using local variables in partials allows us to make these truely re-usable between any view, and not locked to the implimentation in a controller. Perhaps especially as we've put this partial in _shared/ which implies it could be re-used broadly?

Con: It will make your lovely newly slimmed down calls in publication, detailed guide and consultation chunky again. 🤔 

I'm not sure what the right thing is here, no what's blocking?
I wonder if there's a preference in the frontend community?

huwd:
We caught up and discussed the pros / cons of locales, we decided because of this case sticking with what we've got is all cool.
Nice work.

huwd (approved):

PR #2356, opened 2022-02-04
Change single page notification button allow list to an exemption list

danacotoran:
The next stage for the single page notification rollout is to put the button on all pages with a `publication`, `detailed_guide` or `consultation` type. 
The exemption mechanism will be a denylist. Publishers can request that the button not be included on certain individual pages. The way the exemption is applied is by adding the `content_id` of the page to the single page notification button exclusion list.

### Implement notification button exemption list:
Change single page notification button allow list to an exemption list.

Switch from using `base_path`s to `content_id`s for long term stability (content ids persist, base paths can change)

Add some basic tests to verify that pages whose content item is on the exemption list do not display with the single page notification button.

### Put account flash message on `consultation`:
This was missed in the previous PR which added the button to the `consultation` template. The success message is necessary to indicate when subscribing/unsubscribing to notifications has been completed.

### Fix failing test:
`.app-c-published-dates:last-of-type` is failing because there is now another element after the published dates component. 
The `:last-of-type` selector doesn't work the way one might assume on first glance, as in it doesn't apply to classes the same way that it applies to element selectors. 
https://stackoverflow.com/questions/13211453/css-how-to-say-classlast-of-type-classes-not-elements 

-----

https://trello.com/c/4OWUb0no

-----

### DIFF
--- a/app/views/content_items/consultation.html.erb
+++ b/app/views/content_items/consultation.html.erb
@@ -4,6 +4,8 @@
   ) %>
 <% end %>
 
+<%= render 'shared/email_subscribe_unsubscribe_flash' %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render 'govuk_publishing_components/components/title', @content_item.title_and_context %>
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -135,7 +135,7 @@ private
     end
 
     # use these and `@content_item.base_path` in the template
-    @notification_button_visible = in_single_page_notifications_trial? && !@content_item.brexit_child_taxon
+    @notification_button_visible = @content_item.has_single_page_notifications? && !@content_item.brexit_child_taxon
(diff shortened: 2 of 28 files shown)


app/views/content_items/consultation.html.erb
```diff
@@ -4,6 +4,8 @@
   ) %>
 <% end %>
 
+<%= render 'shared/email_subscribe_unsubscribe_flash' %>
```
huwd, line 7: Ah great catch.

huwd: Do we need a test for this?

huwd (changes requested):
Great work, really nice neat Ruby I think - and good catch on the flash and the test.
I wonder if we should add a test to catch the absence of the flash banner?

But generally this is a ✅ 
Gonna abuse the "request changes" feature to block any accidential merge of this until the 28th, when we've heard back from our last stakeholders.

huwd (approved):

huwd (approved):
Nice work on the new tests

PR #2357, opened 2022-02-04
Adding new Dockerfile, puma.rb and ci.yaml files

alexbakervr:


.github/ci.yaml
```diff
@@ -0,0 +1,19 @@
...
+    paths-ignore:
+      - "Jenkinsfile"
+      - ".git**"
+
+jobs:
+  build-publish-image-to-ecr:
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-ecr.yaml@main
+    secrets:
+      AWS_GOVUK_ECR_ACCESS_KEY_ID: ${{ secrets.AWS_GOVUK_ECR_ACCESS_KEY_ID }}
+      AWS_GOVUK_ECR_SECRET_ACCESS_KEY: ${{ secrets.AWS_GOVUK_ECR_SECRET_ACCESS_KEY }}
```
sengi, line 19: tiny nit: missing line ending before end of file here (editor config issue maybe? Same issue in the Dockerfile)

Dockerfile
```diff
@@ -1,24 +1,46 @@
...
-ENV APP_HOME /app
-RUN mkdir $APP_HOME
+ENV RAILS_ENV=production
 
-WORKDIR $APP_HOME
-ADD Gemfile* $APP_HOME/
-ADD .ruby-version $APP_HOME/
-RUN bundle install
+RUN apt-get update -qy && \
+    apt-get upgrade -y && \
```
sengi, line 8: We shouldn't really be running `apt-get upgrade` in Dockerfiles but we're having to do so as a workaround until we have our base images sorted out. Could you pop in the TODOs from (e.g.) https://github.com/alphagov/frontend/blob/main/Dockerfile which signpost this, please? Thanks!

Dockerfile
```diff
@@ -1,24 +1,46 @@
...
+
+RUN bundle config set deployment 'true' && \
+    bundle config set without 'development test' && \
+    bundle install --jobs 4 --retry=2
+
+COPY . /app
+
+RUN GOVUK_APP_DOMAIN=www.gov.uk \
+    GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+    bundle exec rails assets:precompile
```
sengi, line 28: Precompiling assets into the image is another temporary workaround which we should signpost with a TODO (see https://github.com/alphagov/frontend/blob/main/Dockerfile). The reason it's important to comment those things is that a lot of the devs will be new to containers so it'll be helpful to distinguish for them between the things we're doing that are good practice and the things which are sketchy workarounds 😅

sengi (approved):

Dockerfile
```diff
@@ -1,24 +1,46 @@
-FROM ruby:2.7.2
-RUN apt-get update -qq && apt-get upgrade -y
-RUN apt-get install -y build-essential nodejs && apt-get clean
-RUN gem install foreman
+ARG base_image=ruby:2.7.2
```
sengi, line 1: Consider using ruby:2.7.2-slim (it's way smaller and less attack surface)

bilbof (approved):

PR #2366, opened 2022-02-15
Move single page notification button to `government-frontend`

danacotoran:
The single page notification button component was originally created in `govuk_publishing_components` as the initial understanding was that it might be used on different frontend apps.

However, the button is only being used on `government-frontend` currently and it is highly unlikely that there will be a purpose for this component outside of the scope of this application.

This commit copies over the `single_page_notification_button` from `govuk_publishing_components` into `government-frontend`'s app-level component library. 
Some notable adjustments have been made to the test file (`government-frontend` uses Minitest while `govuk_publishing_components` relies on Rspec) and also to the single page notification helper file as instance variables are not allowed in helpers in `government-frontend`.

The component's JS test file relies on Jasmine Ajax so support for that has also been added as part of this work.

The next step in this piece of work is to remove the component from `govuk_publishing_components`.

A few pages which use the component:
https://government-f-move-singl-nrz30x.herokuapp.com/government/publications/open-standards-for-government 
https://government-f-move-singl-nrz30x.herokuapp.com//guidance/check-which-plastic-packaging-is-exempt-from-plastic-packaging-tax
https://government-f-move-singl-nrz30x.herokuapp.com/guidance/uk-trade-agreements-with-non-eu-countries#full-publication-update-history

------

https://trello.com/c/HWDXgTdp

------

### DIFF
--- a/app/views/components/_single_page_notification_button.html.erb
+++ b/app/views/components/_single_page_notification_button.html.erb
@@ -1,0 +1,21 @@
+<%
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
+  wrapper_classes = %w(govuk-!-display-none-print)
+  wrapper_classes << shared_helper.get_margin_bottom
+  component_classes = %w[app-c-single-page-notification-button]
+  component_classes << "js-personalisation-enhancement" if local_assigns[:js_enhancement]
+  base_path = local_assigns[:base_path] || nil
+%>
+<% button_text = capture do %>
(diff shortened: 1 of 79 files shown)


app/views/components/_single_page_notification_button.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
+  wrapper_classes = %w(govuk-!-display-none-print)
```
injms, line 4: We should use square brackets when creating a list this way:

```suggestion
  wrapper_classes = %w[govuk-!-display-none-print]
```

danacotoran:
Closing this because the button needs to stay in the components gem after all, as it is being referenced [here](https://github.com/alphagov/account-api/blob/df050dfaf12e42c9703f7cd15cbb8382a31c6c20/app/controllers/personalisation/check_email_subscription_controller.rb)

PR #2372, opened 2022-02-25
Add accessible form link

1pretz1:
These are needed to render the new accessible form link from GOVUK
Publishing Components. This is dependant on the component being updated
first [1].

This change will apply to publications and consultations, which use the
attachments partial. Initially, the new flow of requesting accessible
formats will be a pilot and only available for documents, owned by a few
organisations. This code will not need to be changed when the change is
released to all orgs.

> App tests will not pass until [1] has been merged

### Running code locally against [1]
Screenshot 2022-02-25 at 16 21 11

[1]: https://github.com/alphagov/govuk_publishing_components/pull/2636

Trello: https://trello.com/c/FdQvPVhD/1094-accessible-format-request-government-frontend-add-parent-document-contentid-to-the-attachment-model

### DIFF
--- a/app/presenters/content_item/attachments.rb
+++ b/app/presenters/content_item/attachments.rb
@@ -1,9 +1,15 @@
 module ContentItem
   module Attachments
     def attachment_details(attachment_id)
-      content_item.dig("details", "attachments")&.find do |attachment|
+      found_attachment = content_item.dig("details", "attachments")&.find do |attachment|
         attachment["id"] == attachment_id
       end
+
+      return unless found_attachment
+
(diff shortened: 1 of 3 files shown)


edwardkerry (commented):
this look OK to me, thanks Pete! Will wait until publishing components is finalised and tests passing before approving though just in case.

1pretz1:
Rebased the PR which uses the latest version of GOVUK Publishing Components so probably could do with a re-review @edwardkerry if you would be so kind.

edwardkerry (approved):

PR #2358, opened 2022-02-07
Update HMPO live chat form to new URL

Tetrino:
As requested by HMPO, this commit changes the old provider URL to
the new provider, as well as the associated availability URL.

We need to decide how this can be tested correctly by HMPO.

https://trello.com/c/wg6cckR1/770-set-up-a-test-page-for-hmpo-to-check-new-webchat-update

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2379, opened 2022-03-04
Remove jQuery from Webchat

leenagupte:
Trello: https://trello.com/c/jymNmjHN

# What's changed?

Remove jQuery from [webchat/library.js](https://github.com/alphagov/government-frontend/blob/main/app/assets/javascripts/webchat/library.js) and [webchat.js](https://github.com/alphagov/government-frontend/blob/main/app/assets/javascripts/webchat.js).

This change is very large, so the initialisation of the code has not been changed, neither has Webchat been converted into a GOVUK module. The focus of this change is just to remove the reliance on jQuery in the Webchat scripts which is a large change by itself.

An anomaly was discovered during the testing of this change.

There is only one webchat page that this script runs on:
[/government/organisations/hm-passport-office/contact/hm-passport-office-webchat](https://www.gov.uk/government/organisations/hm-passport-office/contact/hm-passport-office-webchat)

The [availabilityURL] always returns a response of:

```
{
  "status": "success",
  "response": "UNAVAILABLE"
}
```

with an http status of 200. This seems incorrect as the [openUrl] states that all advisors are busy, so we would expect the availabilityURL to return a status of "BUSY". 

There is an [open ticket] to change the URLs, so hopefully that will fix the problem. The removal of jQuery in this change has not negatively affected the workings of Webchat. The behaviour hasn't been affected, even if that behaviour is incorrect.

The response stubbed from the [availabilityURL] has also been changed in the tests as the switch from using Ajax to calling XMLHttpRequest requires more information from request object (http status code and readyState) to determine when the request has been completed.

## Results from testing
Screenshot 2022-03-04 at 14 33 48

Screenshot 2022-03-04 at 14 34 42

[availabilityURL]: https://hmpowebchat.klick2contact.com/v03/providers/HMPO/api/availability.php
[openUrl]: [https://hmpowebchat.klick2contact.com/v03/launcherV3.php?p=HMPO&d=717&ch=CH&psk=c](https://hmpowebchat.klick2contact.com/v03/launcherV3.php?p=HMPO&d=717&ch=CH&psk=chat_a1&iid=STC&srbp=0&fcl=0&r=Chat&s=https://hmpowebchat.klick2contact.com/v03&u=&wo=&uh=&pid=2&iif=0)
[open ticket]: https://github.com/alphagov/government-frontend/pull/2358

### DIFF
--- a/app/assets/javascripts/webchat.js
+++ b/app/assets/javascripts/webchat.js
@@ -1,14 +1,10 @@
 //= require ./webchat/library.js
 
-var $ = window.$
-
-$(document).ready(function () {
-  var GOVUK = window.GOVUK
-  if (GOVUK.Webchat) {
-    var webchats = document.querySelectorAll('.js-webchat')
-    for (var i = 0; i < webchats.length; i++) {
-      /* eslint-disable no-new */
-      new GOVUK.Webchat(webchats[i])
-    }
+var GOVUK = window.GOVUK
+if (GOVUK.Webchat) {
+  var webchats = document.querySelectorAll('.js-webchat')
+  for (var i = 0; i < webchats.length; i++) {
+    /* eslint-disable no-new */
+    new GOVUK.Webchat(webchats[i])
   }
-})
+}
--- a/app/assets/javascripts/webchat.js
+++ b/app/assets/javascripts/webchat.js
@@ -5,10 +5,10 @@ var $ = window.$
 $(document).ready(function () {
   var GOVUK = window.GOVUK
   if (GOVUK.Webchat) {
-    $('.js-webchat').map(function () {
-      return new GOVUK.Webchat({
-        $el: $(this)
-      })
-    })
+    var webchats = document.querySelectorAll('.js-webchat')
+    for (var i = 0; i < webchats.length; i++) {
+      /* eslint-disable no-new */
+      new GOVUK.Webchat(webchats[i])
+    }
   }
 })
--- a/app/assets/javascripts/webchat/library.js
+++ b/app/assets/javascripts/webchat/library.js
@@ -90,9 +90,13 @@
 
     function advisorStateChange (state) {
       state = state.toLowerCase()
-      var currentState = $el.find('.' + webchatStateClass + state)
-      $el.find('[class^="' + webchatStateClass + '"]').addClass('govuk-!-display-none')
-      currentState.removeClass('govuk-!-display-none')
+      var currentState = el.querySelector('[class^="' + webchatStateClass + state + '"]')
+      var allStates = el.querySelectorAll('[class^="' + webchatStateClass + '"]')
+
+      for (var index = 0; index < allStates.length; index++) {
+        allStates[index].classList.add('govuk-!-display-none')
+      }
+      currentState.classList.remove('govuk-!-display-none')
       trackEvent(state)
     }
 
--- a/app/assets/javascripts/webchat/library.js
+++ b/app/assets/javascripts/webchat/library.js
@@ -39,14 +39,21 @@
     }
 
     function checkAvailability () {
-      var ajaxConfig = {
-        url: availabilityUrl,
-        type: 'GET',
-        timeout: AJAX_TIMEOUT,
-        success: apiSuccess,
-        error: apiError
+      var done = function () {
+        if (request.readyState === 4 && request.status === 200) {
+          apiSuccess(JSON.parse(request.response))
+        } else {
+          apiError()
+        }
       }
-      $.ajax(ajaxConfig)
+
+      var request = new XMLHttpRequest()
+      request.open('GET', availabilityUrl, true)
+      request.setRequestHeader('Content-type', 'application/x-www-form-urlencoded')
+      request.addEventListener('load', done.bind(this))
+      request.timeout = AJAX_TIMEOUT
+
+      request.send()
     }
 
     function apiSuccess (result) {
--- a/app/assets/javascripts/webchat/library.js
+++ b/app/assets/javascripts/webchat/library.js
@@ -1,10 +1,7 @@
 (function (global) {
-  'use strict'
-  var $ = global.jQuery
-  if (typeof global.GOVUK === 'undefined') { global.GOVUK = {} }
-  var GOVUK = global.GOVUK
+  var GOVUK = global.GOVUK || {}
 
-  function Webchat (options) {
+  function Webchat (el) {
     var POLL_INTERVAL = 5 * 1000
     var AJAX_TIMEOUT = 5 * 1000
     var API_STATES = [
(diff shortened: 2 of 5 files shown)


andysellick:
Mentioning this for the record. The URL for HMPO is being updated in this PR: https://github.com/alphagov/government-frontend/pull/2358

app/assets/javascripts/webchat.js
```diff
@@ -1,8 +1,6 @@
 //= require ./webchat/library.js
 
-var $ = window.$
-
-$(document).ready(function () {
+document.addEventListener('DOMContentLoaded', function () {
```
andysellick, line 3: Recently had to do this exact change in another script and Kevin pointed out it would work just fine by removing the `document.ready` entirely without replacing it. Can you test whether this would work?

app/assets/javascripts/webchat/library.js
```diff
@@ -27,14 +23,19 @@
       if (!availabilityUrl || !openUrl) {
         throw Error.new('urls for webchat not defined')
       }
-      $openButton.on('click', handleOpenChat)
+
+      if (redirectUrl === true) {
```
andysellick, line 27: Just curious why this `if` statement wraps the button add listener? I don't see the equivalent in the original code.

app/assets/javascripts/webchat/library.js
```diff
@@ -27,14 +23,19 @@
...
+        var openButton = document.getElementsByClassName('.js-webchat-open-button')[0]
+        openButton.addEventListener('click', handleOpenChat)
```
andysellick, line 29: Suggestion (slightly cleaner and a little more robust - jQuery silently fails if you attach a listener to a non-existent element, JS doesn't)

```
var openButton = document.querySelector('.js-webchat-open-button')
if (openButton) {
  openButton.addEventListener('click', handleOpenChat)
}
```

app/assets/javascripts/webchat/library.js
```diff
@@ -27,14 +23,19 @@
       if (!availabilityUrl || !openUrl) {
         throw Error.new('urls for webchat not defined')
```
andysellick, line 24: Not your code, but I wonder if we should take this opportunity to improve this error handling. Kevin suggested something like this on a bit of work I did recently:

```
throw Error.new('urls for webchat not defined', window.location)
```

app/assets/javascripts/webchat/library.js
```diff
@@ -27,14 +23,19 @@
...
+      }
       intervalID = setInterval(checkAvailability, POLL_INTERVAL)
       checkAvailability()
     }
 
     function handleOpenChat (evt) {
       evt.preventDefault()
-      this.dataset.redirect === 'true' ? window.location.href = openUrl : global.open(openUrl, 'newwin', 'width=366,height=516')
+      var redirect = this.getAttribute('data-redirect')
+      redirect === 'true' ? window.location.href = openUrl : window.open(openUrl, 'newwin', 'width=366,height=516')
```
andysellick, line 38: I'm not certain, but I think this line isn't the equivalent of the original. In the original, `this.dataset.redirect` is set to something based on the ternary operator. In the new code, only the `redirect` variable is updated.

I think this might need a line adding after it to set the attribute back onto the element, e.g.

```
redirect === 'true' ? window.location.href = openUrl : window.open(openUrl, 'newwin', 'width=366,height=516')
this.setAttribute('data-redirect', redirect)
```

Having said that, I'm not sure how changing this attribute would change anything, since the only instance of it being called is once when the JS is initialised. So, confused.

app/assets/javascripts/webchat/library.js
```diff
@@ -91,9 +91,13 @@
 
     function advisorStateChange (state) {
       state = state.toLowerCase()
-      var currentState = $el.find('.' + webchatStateClass + state)
-      $el.find('[class^="' + webchatStateClass + '"]').addClass('govuk-!-display-none')
-      currentState.removeClass('govuk-!-display-none')
+      var currentState = el.querySelectorAll('[class^="' + webchatStateClass + state + '"]')[0]
```
andysellick, line 101: In jQuery, `.find()` finds everything that matches. In JavaScript, it's weirdly more complex

- `querySelector` finds the first element that matches
- `querySelectorAll` finds all elements that match

If you do `querySelectorAll()[0]` that returns the first one found - it's equivalent to doing `querySelector`. Which one do you want to do here? I suspect it's `querySelector`?

andysellick (commented):
Looks good! Well done for jumping into this tangle. Just a few questions and suggestions.

app/assets/javascripts/webchat.js
DilwoarH: You might also be able to to do the following:

```js
(function () {
.... your code
})()
```

andysellick: I think a self executing function in this context would be the same as having the code execute at this point without a document.ready.

app/assets/javascripts/webchat/library.js
leenagupte: Would passing `window.location.href` be more useful than using `window.location`? The `window.location` object seems to have a lot of stuff in it:

Screenshot 2022-03-08 at 12 00 59

andysellick: Either's good.

app/assets/javascripts/webchat/library.js
leenagupte: Ah, I think I was being overly cautious, because the "open button" only exists when `data-redirect` is true. However, you are right it in that it's not needed. I'll swap it for you selection below. Thanks!

app/assets/javascripts/webchat/library.js
leenagupte: Oh thanks! I created the `redirect` variable because otherwise the line length made the statement unreadable. But yes, it makes sense that the original element needs to be updated, thanks!

And yes, I'm confused too!

app/assets/javascripts/webchat/library.js
leenagupte: 🤦‍♀️  Yes, I just need `querySelector`

app/assets/javascripts/webchat/library.js
leenagupte: I'll add this change to a new commit.

leenagupte:
@andysellick Thanks for your review. I've made the recommended changes and fixed up.

leenagupte:
When https://github.com/alphagov/government-frontend/pull/2358 is merged tomorrow, Webchat will start working again. At that point I will rebase and make sure that this change hasn't broken anything.

app/assets/javascripts/webchat.js
```diff
@@ -1,14 +1,10 @@
...
-$(document).ready(function () {
-  var GOVUK = window.GOVUK
-  if (GOVUK.Webchat) {
-    $('.js-webchat').map(function () {
-      return new GOVUK.Webchat({
-        $el: $(this)
-      })
-    })
+var GOVUK = window.GOVUK
+if (GOVUK.Webchat) {
```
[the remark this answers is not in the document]
andysellick: Ah excellent, full marks 👍

leenagupte:
It's currently showing "Advisors are busy" which matches live. I would like to see the pop-up open when advisors are available before merging.

Screenshot 2022-03-10 at 10 23 47

leenagupte:
Oh dear. It's seems that the "Advisors are available" message shows. But the link doesn't link to anywhere 😱

leenagupte:
Success! I was defining `openButton` incorrectly. Rather than:
```javascript
var openButton = document.getElementsByClassName('.js-webchat-open-button')[0]
```

it should have been:
```javascript
var openButton = document.querySelector('.js-webchat-open-button')
```

Screenshot 2022-03-10 at 13 40 00

I'll update and fix up.

PR #2383, opened 2022-03-08
Update email subscription banner content

danacotoran:
Update the single page notification success banner content based on
input from our designers and reduce some repetition in the
`email_subscribe_unsubscribe_flash` partial

-----

https://trello.com/c/bPoMTXr0

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -13,6 +13,8 @@ class ContentItemsController < ApplicationController
 
   attr_accessor :content_item, :taxonomy_navigation
 
+  helper_method :show_email_subscription_success_banner?
+
   def show
     load_content_item
 
@@ -44,6 +46,10 @@ class ContentItemsController < ApplicationController
 
       redirect_to service_url(selected[:url])
     end
+  end
+
+  def show_email_subscription_success_banner?
+    @account_flash.include?("email-subscription-success") || @account_flash.include?("email-unsubscribe-success") || @account_flash.include?("email-subscription-already-subscribed")
   end
 
 private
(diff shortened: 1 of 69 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -46,6 +48,10 @@ def service_sign_in_options
     end
   end
 
+  def show_email_subscription_success_banner?
```
alex9smith, line 51: When we have a helper method that's needed in the view but not the controller, we usually put it in its own helper file - see [this comment from Huw](https://github.com/alphagov/email-alert-frontend/pull/1197#discussion_r751131532) explaining it on one of my PRs.

So in this case we'd have a `app/helpers/content_items_helper.rb` file with a `ContentItemsHelper` class and this method.

What do you think?

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -1,29 +1,25 @@
-<% if @account_flash.include?("email-subscription-success") %>
+<% banner_heading = capture do %>
```
alex9smith, line 1: I think this should go into a helper function as well to keep the view smaller - either in the helper class [suggested here](https://github.com/alphagov/government-frontend/pull/2383/files#r822405822), or as a controller method if you decide not to go with that

app/controllers/content_items_controller.rb
danacotoran: Thank you @alex9smith, that makes sense, I get fuzzy on what is best practice sometimes, it's good to be reminded! 

I've made the changes as suggested, let me know if that's okay.

alex9smith (approved):
Looks great 👍🏻

PR #2386, opened 2022-03-09
Change heading on foreign-travel-advice

chris-gds:
## What 

Change to component, alter from `h1` to `h2` on page: https://www.gov.uk/foreign-travel-advice/afghanistan

## Why

> The problem is that in part-separated content we have two `h1` headings, one for the guide name, then one for the part name, and then the content is displayed with `h2` headings.  Make the guide heading a `h2` even though it comes before the part heading. This is not ideal as it's a slightly confusing structure but does not violate any standards and it leaves the `h1` as the most relevant title on the page.

## Visual changes

None

## Anything else

Ideally the Content under the `h2` should now be updated to `h3` however this lives within `govspeak` controlled by Content teams. Been made aware there are 226 different guides to update.
https://www.gov.uk/foreign-travel-advice/afghanistan

### DIFF
--- a/app/assets/stylesheets/helpers/_parts.scss
+++ b/app/assets/stylesheets/helpers/_parts.scss
@@ -20,15 +20,4 @@
       margin-left: govuk-spacing(1);
     }
   }
-
-  .part-title {
-    @include govuk-font(27, $weight: bold);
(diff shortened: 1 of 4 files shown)


hannako:
Closing as this PR has been open 3 years

PR #2424, opened 2022-04-20
Migrate manuals from Manuals Frontend to Government Frontend

1pretz1:
This PR migrates manuals, manual sections and updates pages to Government Frontend, from Manuals Frontend. HMRC manuals have also been migrated. I've treated each page as its own content item (with own views / presenters) as this follows convention within the application, even though the `/updates` pages aren't actually represented as a content items in content store. I've also tried to reduce duplication by using presenter modules and view partials. 

I've left some to-do's which should be resolved at a later date. They mainly concern extra requests to Content Store for manual section pages, as they need data that is only available on the parent manual page. We should add the extra data needed to the manual section when it is being published, which would solve the issue. 

Examples of these pages can be found below.

### Manual

* https://www.gov.uk/guidance/content-design (manual)
* https://www.gov.uk/guidance/content-design/what-is-content-design (manual section)
* https://www.gov.uk/guidance/content-design/updates (manual updates page)

### HMRC Manual

* https://www.gov.uk/hmrc-internal-manuals/admin-law-manual (HMRC manual)
* https://www.gov.uk/hmrc-internal-manuals/admin-law-manual/adml1000 (HMRC section)
* https://www.gov.uk/hmrc-internal-manuals/admin-law-manual/updates (HMRC updates page).

### Testing this PR

You can test the PR yourself either locally or on integration. Local testing is simpler and you can do this by running the app your self against imported or live data (e.g `./startup --live` script then localhost:3090/guidance/some-manual-path) or using [docker][1].

[1]: https://github.com/alphagov/govuk-docker/blob/9f2ce14097583111d18bf7ee12b3393929a517a8/projects/government-frontend/docker-compose.yml#L40

For testing on integration you will need to deploy two branches and then publish a manual from Manuals Publisher. You can only test a normal manual, unless you publish a manual with your Signon organisation set to HMRC (instead of GDS).

1) Deploy to integration: https://github.com/alphagov/manuals-publisher/pull/1844

2) Deploy to integration: https://github.com/alphagov/government-frontend/pull/2424

3) Publish new manual: https://manuals-publisher.integration.publishing.service.gov.uk/ 

4) View on: https://www.integration.publishing.service.gov.uk/

### Dependant on

- [x] https://github.com/alphagov/govuk-content-schemas/pull/1090

Trello:
https://trello.com/c/YmmLFXu5/1127-manuals-migrate-manuals-frontend-code-to-government-frontend

### DIFF
--- a/app/presenters/manual_section_presenter.rb
+++ b/app/presenters/manual_section_presenter.rb
@@ -1,0 +1,59 @@
+class ManualSectionPresenter < ContentItemPresenter
+  include ContentItem::Metadata
+  include ContentItem::Manual
+  include ContentItem::ManualSection
+
+  def base_path
+    manual["base_path"]
+  end
+
+  def intro
+    return nil unless details["body"]
+
+    intro = Nokogiri::HTML::DocumentFragment.parse(details["body"])
+
+    # Strip all content following and including h2
+    intro.css("h2").xpath("following-sibling::*").remove
+    intro.css("h2").remove
+
+    intro.text.squeeze == "\n" ? "" : intro
+  end
+
+  def visually_expanded?
+    details.fetch("visually_expanded", false)
+  end
+
+  def main
+    return nil unless details["body"]
+
+    document = Nokogiri::HTML::DocumentFragment.parse(details["body"])
+
+    # Identifies all h2's and creates an array of objects from the heading and
+    # its proceeding content up to the next heading. This is so that it can be
+    # consumed by accordion components in the template.
+    document.css("h2").map do |heading|
+      content = []
+      heading.xpath("following-sibling::*").each do |element|
+        if element.name == "h2"
+          break
+        else
+          content << element.to_html
+        end
+      end
+
+      {
+        heading: {
+          text: heading.text,
+          id: heading[:id],
+        },
+        content: content.join,
+      }
+    end
+  end
+
+private
+
+  def manual
+    content_item["links"]["manual"].first
+  end
+end
--- a/app/views/content_items/hmrc_manual.html.erb
+++ b/app/views/content_items/hmrc_manual.html.erb
@@ -1,0 +1,28 @@
+<div id="manuals-frontend" class="manuals-frontend-body">
+
+<%= machine_readable_metadata(schema: :article) %>
+
+<%= render "content_items/manuals/hmrc_header", {
+  content_item: @content_item, heading_level: 1, margin_bottom: 3
+} %>
+
+<div class="manual-body">
+  <% if @content_item.description.present? %>
+    <%= render "govuk_publishing_components/components/lead_paragraph", {
+      text: @content_item.description
+    } %>
+  <% end %>
+  <% if @content_item.body.present? %>
+    <%= render "govuk_publishing_components/components/govspeak", {} do %>
+      <%= raw(@content_item.body) %>
+    <% end %>
+  <% end %>
+
+  <% @content_item.section_groups.each do |group| %>
+    <div class="subsection-collection">
+      <%= render "content_items/manuals/hmrc_sections", group: group %>
+    </div>
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/print_link" %>
+</div>
--- a/app/views/content_items/manuals/_updates.html.erb
+++ b/app/views/content_items/manuals/_updates.html.erb
@@ -1,0 +1,51 @@
+<div class="govuk-grid-row">
+  <div class="manual-body" id="content">
+    <article aria-labelledby="section-title">
+      <div class="govuk-grid-column-full">
+        <%= render "govuk_publishing_components/components/heading", {
+          font_size: "l",
+          heading_level: 1,
+          id: "section-title",
+          margin_bottom: 4,
+          text: t("manuals.updates_title", title: content_item.title),
+        } %>
+      </div>
+
+      <% content_item.presented_change_notes.each do |year, updates_by_year| %>
+        <div class="govuk-grid-column-two-thirds">
+          <%= render "govuk_publishing_components/components/heading", {
+            text: year,
+            font_size: "l"
+          } %>
+
+          <%= render "govuk_publishing_components/components/accordion", {
+            heading_level: 3,
+            items: updates_by_year.map do |day, updated_documents|
+              accordion_content = capture do %>
+                <div class="govuk-!-margin-top-3">
+                  <% updated_documents.each do |_, change_notes| %>
+                    <% change_notes.each do |change_note| %>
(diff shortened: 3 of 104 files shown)


jon-kirwan (changes requested):

test/integration/manual_test.rb
```diff
@@ -0,0 +1,65 @@
...
+    setup_and_visit_content_item("content-design")
+
+    assert_has_component_title(@content_item["title"])
+    assert page.has_text?(@content_item["description"])
+  end
+
+  test "renders metadata" do
+    setup_and_visit_content_item("content-design")
+
+    assert_has_metadata(
```
edwardkerry, line 14: I often wonder if we should somehow reference or link in a comment to the [govuk-content-schema](https://github.com/alphagov/govuk-content-schemas/blob/b3c10be62e86ef6ee4fb482f75dbd4c4863ebae7/examples/manual/frontend/content-design.json) this data comes from. It can sometimes be a little opaque to trace from this test => gds-api-adapters => govuk-content-schemas. More a thought than a suggestion!

app/presenters/manual_section_presenter.rb
```diff
@@ -0,0 +1,59 @@
...
+  def intro
+    return nil unless details["body"]
+
+    intro = Nokogiri::HTML::DocumentFragment.parse(details["body"])
+
+    # Strip all content following and including h2
+    intro.css("h2").xpath("following-sibling::*").remove
+    intro.css("h2").remove
+
+    intro.text.squeeze == "\n" ? "" : intro
```
edwardkerry, line 19: Is it possible for an intro to have more than one new line at this point i.e could we end up with `"\n \n"`?
Would it be possible to write a test for this new functionality?

edwardkerry:
Excellent work @1pretz1, really appreciate the clear commit messages and history. I just have a few non-blocking admin questions and one suggestion for a possible test, but otherwise this large migration looks good.

edwardkerry (commented):
not a blocker but could this commit be squashed into `Add page furniture locales`?

1pretz1:
Thanks for the reviews @jon-kirwan and @edwardkerry! I believe I've addressed all the comments. 

I couldn't see what the last comment by Ed referred to and so I assumed it was a suggestion of merging https://github.com/alphagov/government-frontend/pull/2424/commits/76730306dfec30ba5d20534f896a639a9d9c66e6 into https://github.com/alphagov/government-frontend/pull/2424/commits/ebf79f526c8a135d19951299d8340d77e87c671c. Instead, I've moved the commits so they're next to each other, as I felt combining them cluttered the only locales we actually have values for (English).

test/integration/manual_test.rb
edwardkerry: Yeah, it's a general faff rather than anything for this PR in particular to solve!

edwardkerry (approved):

edwardkerry (approved):
Thanks for this @1pretz1!

app/views/content_items/manuals/_updates.html.erb
```diff
@@ -0,0 +1,52 @@
...
+            text: year,
+            font_size: "l"
+          } %>
+
+          <%= render "govuk_publishing_components/components/accordion", {
+            heading_level: 3,
+            items: updates_by_year.map do |day, updated_documents|
+              accordion_content = capture do %>
+                <div class="updates-list">
+                  <% updated_documents.each do |_, change_notes| %>
```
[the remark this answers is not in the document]
jon-kirwan: Thanks @1pretz1, looks great! Sorry to be really fussy but could we also remove the `<br />` below the `simple_format` line. I think it's no longer necessary as the the `govuk-body` already has margin-bottom.

app/views/content_items/hmrc_manual.html.erb
```diff
@@ -0,0 +1,30 @@
+<div class="govuk-width-container">
+  <div id="manuals-frontend" class="manuals-frontend-body">
```
jon-kirwan, line 2: We can remove the parent `<div class="govuk-width-container">` here since we already have the same class on the main wrapper element `<div id="wrapper" class="direction-ltr govuk-width-container">` and the effect is to have double spacing.

<table role="table">
<thead>
<tr>
<th>Live</th>
<th>Dev</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/168244255-780f5c1e-f65e-4f57-94c8-d2c36297fb2f.png"></a></td>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/168244627-03917120-7ffd-4ce4-b034-cebecb46d55b.png"></a></td>
</tr>
</tbody>
</table>

app/views/content_items/manuals/_updates.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          <%= render "govuk_publishing_components/components/heading", {
+            text: year,
+            font_size: "l"
+          } %>
+
+          <%= render "govuk_publishing_components/components/accordion", {
+            heading_level: 3,
+            items: updates_by_year.map do |day, updated_documents|
+              accordion_content = capture do %>
+                <div class="updates-list">
```
jon-kirwan, line 25: One more change here: Can we please replace `updates-list` (no longer needed) with `govuk-!-margin-top-3` to fix an issue with no spacing between the list of updates and the accordion header / button

jon-kirwan (changes requested):

app/views/content_items/manuals/_updates.html.erb
1pretz1: Ah, good spot. Have removed `updates-list` and add `govuk-!-margin-top-3`.

app/views/content_items/hmrc_manual.html.erb
1pretz1: Nice find! I didn't notice the duplicate `govuk-width-container` as I didn't realise this class was injected into the HTML as a result of the Slimmer middleware. Have learnt more about our templating works though with a little quest bouncing from Slimmer to Static to Govuk Publishing Components 😄

app/views/content_items/manuals/_updates.html.erb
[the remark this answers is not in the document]
1pretz1: No worries at all - I welcome your comments! Have removed the `<br />`.

1pretz1:
Hey @jon-kirwan, thank you for the comments, really useful! I believe I've addressed them all now if you'd like to take another gander.

jon-kirwan (approved):
Thanks for the changes Pete. Great work! 🚀

PR #2467, opened 2022-06-21
Fix redirecting to incorrect assets domain

kevindew:
Trello: https://trello.com/c/nvG4055w/510-asset-redirects-are-broken-in-draft-draft-draft-assets

The previous code had a bug where a redirect to the draft assets host
was incorrect. It was redirecting to a double prefixed asset host as the
"draft-" was implicit. Illustrated by:

```
➜  ~ govuk-connect app-console -e production draft_frontend/government-frontend
irb(main):001:0> Plek.new.external_url_for("draft-assets")
=> "https://draft-draft-assets.publishing.service.gov.uk"
irb(main):002:0> Plek.new.external_url_for("assets")
=> "https://draft-assets.publishing.service.gov.uk"
```

This resolves the problem by using the external_url_for method for both
scenarios. As when we are on the draft stack there will be an env var
that prefixes hostnames accordingly.

I added climate_control to test this, since a regression test seemed
appropriate. I removed the preview test as there was no logic written
for this and the other tests cover that code path.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,7 @@ end
 
 group :test do
   gem "capybara"
+  gem "climate_control"
   gem "faker"
   gem "i18n-coverage"
   gem "minitest-reporters"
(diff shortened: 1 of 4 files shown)


ollietreend (approved):
Great stuff ✨ thanks for this

PR #2056, opened 2021-03-23
Hides print button if content longer than 10 items on HTML

DilwoarH:
This change hides the print button under the content if the content items exceed 10 or more.

This fixes the following issue:
https://github.com/alphagov/government-frontend/issues/1907

It means that there is only one print button showing at one time.

## Before

before

## After

Mar-23-2021 12-38-26

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #2493, opened 2022-07-20
Use CSS sticky on HTML publications instead of custom script

andysellick:

## What
Stop using `sticky-element-container.js` on [HTML publication pages](https://www.gov.uk/government/publications/heatwave-plan-for-england/beat-the-heat-staying-safe-in-hot-weather) and instead use native CSS sticky.

The page layout has a contents list in the left hand column and the main content in the right. The script was used to keep a jump link to the contents (and a print this page button) visible at all times when scrolling, so users can jump back to the top at any time.

This change switches from using the script to the native CSS position: sticky attribute. It has been applied to the contents list, so now the entire contents list is visible in the left column when the page is scrolled. It also stops in the right position automatically when reaching the bottom.

Internet Explorer doesn't support the CSS sticky property, so this doesn't work in that browser. It fails gracefully though, in that nothing bad happens.

Behaviour on mobile remains the same - contents stack at top, link at bottom jumps back to top, no sticky.

## Why
`sticky-element-container.js` isn't perfect - it has a bug under certain conditions that makes the sticky element overlap the footer. It's also very complicated, and something we'd like to retire, but it's used on another page still so it's still needed.

I've kept the 'Contents' jump link at the foot of the page but renamed it to 'Back to top', as that is more of an accurate description. However now that the whole contents menu scrolls with the page it's possible that we don't need this link at all (except possibly for Internet Explorer users).

## Visual changes

| Before | After |
------- | -------
Initial state at the top of the page (unchanged) | Initial state at the top of the page (unchanged)
Screenshot 2022-07-20 at 11 50 05 | Screenshot 2022-07-20 at 12 43 53
When scrolling down - just contents link and print button visible | When scrolling down - whole contents menu and print button visible
Screenshot 2022-07-20 at 11 50 17 | Screenshot 2022-07-20 at 12 44 02
At the bottom - print button and contents link stop | At the bottom - contents stop, back to top link replaces contents link (but does the same thing)
Screenshot 2022-07-20 at 11 50 28 | Screenshot 2022-07-20 at 12 44 15

### DIFF
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -25,6 +25,9 @@
     @include govuk-media-query($from: desktop) {
       position: sticky;
       top: govuk-spacing(2);
+      max-height: 100vh;
+      overflow-y: auto;
+      margin-bottom: govuk-spacing(4);
     }
 
     .direction-rtl & {
--- a/app/assets/stylesheets/views/_html-publication.scss
+++ b/app/assets/stylesheets/views/_html-publication.scss
@@ -22,6 +22,11 @@
   }
 
   .contents-list-container {
+    @include govuk-media-query($from: desktop) {
+      position: sticky;
+      top: govuk-spacing(2);
+    }
+
     .direction-rtl & {
       float: right;
     }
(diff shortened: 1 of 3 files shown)


injms (approved):
Nice update - thank you!

DilwoarH:
@andysellick - does this fix #2056 as well?

andysellick:
> @andysellick - does this fix #2056 as well?

@DilwoarH yes, I believe so.

PR #2509, opened 2022-08-03
Visa service recruitment banner

hannako:
Add recruitment banner to pages related to the UK visa service.

See [trello](https://trello.com/c/4oYXIjMN/1173-add-a-research-recruitment-banner-for-the-visa-journeys-team-s) card for list of pages.

Screenshot 2022-08-03 at 14 29 39

### DIFF
--- a/app/presenters/content_item/recruitment_banner.rb
+++ b/app/presenters/content_item/recruitment_banner.rb
@@ -4,13 +4,35 @@ module ContentItem
     TREE_TEST_PAGE = "/browse/working".freeze
     TREE_TEST_MAPPINGS = { TREE_TEST_PAGE => TREE_TEST_URL }.freeze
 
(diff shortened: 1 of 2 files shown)


AgaDufrat (approved):
LGTM. Testing all pages would warn us if we are trying to recruit users for more than one test on the same page. 
If it happens that one of the Visa tests pages is tagged to "Working..." browse, the test URL would be for the Tree test. We should probably mention this the UR. If that would be an issue we should probably ask which test should take priority.

hannako:
@AgaDufrat Yes that's correct - I've set it so tree test trumps visa test. I didn't check every content item, but I checked the first 5 or so and none of those were tagged to our browse category of interest so I don't THINK it will be an issue. But I'll make a note on the trello card for Lorna 👍

PR #2515, opened 2022-08-09
Update Tree test survey link

AgaDufrat:
as per tracking sheet: https://docs.google.com/spreadsheets/d/104ird2xevvcA77Lrhtrelkz4DpkoLJ30jdnDaq6bhpg/edit


patrickpatrickpatrick (approved):
needs to pass that test but it's a simple link swap so should be OK

PR #2516, opened 2022-08-09
Implement the content list component on `manual_section` pages for Ministry of Justice only

jon-kirwan:
## What
https://trello.com/c/564ra84M/1132-manuals-implement-list-of-contents-for-manuals

Implement the content list component on `manual_section` pages and for **Ministry of Justice** only

## Why

Feature request as sections are easier to navigate with a list of contents.

## Visual changes
Sample pages created on Integration.

## Manual (Ministry of Justice)
Sample page based on https://www.justice.gov.uk/courts/procedure-rules/civil/rules

Screenshot 2022-08-15 at 14 33 08

## Manual section (Ministry of Justice)
Sample page based on https://www.justice.gov.uk/courts/procedure-rules/civil/rules/raprnotes

Screenshot 2022-08-15 at 14 33 34

## Manual section (Ministry of Justice)
Sample page based on https://www.justice.gov.uk/courts/procedure-rules/civil/rules/part01

Screenshot 2022-08-15 at 14 33 55

## Manual updates (Ministry of Justice)

Screenshot 2022-08-15 at 14 35 23

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -19,6 +19,7 @@ module ContentItem
     end
 
     def show_contents_list?
+      return false if exclude_contents_list_for_manual_section?
       return false if contents_items.count < 2
       return true if contents_items.count > 2
       return false if no_first_item?
@@ -109,6 +110,13 @@ module ContentItem
 
     def no_first_item?
       first_item.nil?
+    end
+
+    def exclude_contents_list_for_manual_section?
+      # MOJ require contents lists for manual section pages.
+
+      moj_content = content_item.dig("links", "organisations", 0, "content_id") == "dcc907d6-433c-42df-9ffb-d9c68be5dc4d"
+      document_type == "manual_section" && !moj_content
     end
   end
 end
--- a/app/presenters/content_item/manual.rb
+++ b/app/presenters/content_item/manual.rb
@@ -16,20 +16,8 @@ module ContentItem
     alias_method :manual_page_title, :page_title
 
     def breadcrumbs
-      crumbs = []
-
-      if view_context.request.path == base_path
-        crumbs.push({
-          title: I18n.t("manuals.breadcrumb_contents"),
-        })
-      else
-        crumbs.push({
-          title: I18n.t("manuals.breadcrumb_contents"),
-          url: base_path,
-        })
-      end
+      [{ title: I18n.t("manuals.breadcrumb_contents") }]
     end
-    alias_method :manual_breadcrumbs, :breadcrumbs
 
     def section_groups
       content_item.dig("details", "child_section_groups") || []
--- a/app/presenters/content_item/manual_section.rb
+++ b/app/presenters/content_item/manual_section.rb
@@ -15,6 +15,24 @@ module ContentItem
       document_heading << content_item["title"] if content_item["title"]
     end
 
+    def breadcrumbs
+      if show_contents_list?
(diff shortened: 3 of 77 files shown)


test/integration/hmrc_manual_updates_test.rb
```diff
@@ -39,7 +39,7 @@ class HmrcManualUpdatesTest < ActionDispatch::IntegrationTest
 
     manual_specific_breadcrumbs = page.all(".gem-c-breadcrumbs")[1]
     within manual_specific_breadcrumbs do
-      assert page.has_link?(I18n.t("manuals.breadcrumb_contents"), href: @content_item["base_path"])
+      assert page.has_text?(I18n.t("manuals.breadcrumb_contents"))
```
1pretz1, line 42: We'll want to keep this behaviour ('Contents' link for HMRC and normal manuals updates pages to the parent manual). 

As part of this I separated out the breadcrumbs method as it was getting a bit chaotic in one place, but missed updating the [updates module](https://github.com/alphagov/government-frontend/blob/main/app/presenters/content_item/manual_updates.rb).

We could probably just add:

```
def breadcrumbs
  [
    {
      title: I18n.t("manuals.breadcrumb_contents"),
      url: base_path,
    },
  ]
end
```

app/presenters/content_item/contents_list.rb
```diff
@@ -114,7 +114,9 @@ def no_first_item?
 
     def exclude_contents_list_for_manual_section?
       # MOJ require contents lists for manual section pages.
-      moj_content = content_item["links"]["organisations"].first["content_id"] == "dcc907d6-433c-42df-9ffb-d9c68be5dc4d"
+
+      # moj_content = content_item["links"]["organisations"]&.try(:content_id)&.first == "dcc907d6-433c-42df-9ffb-d9c68be5dc4d"
+      moj_content = content_item["links"]["organisations"] && content_item["links"]["organisations"].first["content_id"] == "dcc907d6-433c-42df-9ffb-d9c68be5dc4d"
```
1pretz1, line 119: I'm assuming we hit some content items that don't have organisations links? In which case, nice catch!

1pretz1 (commented):
Thanks for picking this up @jon-kirwan! I've left two comments, with only one potential modification. The commits could probably be tidied up too (as we still have WIP).

One other thing to note is that the contents list component in it's current state if still subject to the other conditionals (https://github.com/alphagov/government-frontend/blob/ef8ba143a2b6178b5682d10ffb420ca6289d657d/app/presenters/content_item/contents_list.rb#L23-L25), which means an MOJ manual section could still not display a contents list  if, for example, the number of content items is less than two. I think these general rules are fine, as they ensure a list is only rendered if it makes sense.

test/integration/hmrc_manual_updates_test.rb
jon-kirwan: Thanks, updated and removed the related test changes.

app/presenters/content_item/contents_list.rb
jon-kirwan: Thanks Pete. I had various test failures related to the previous changes and so have amended like so (which seems to be okay) >

https://github.com/alphagov/government-frontend/blob/7f53b82f9884fdc45a8c51a7523ca93f961fb3e6/app/presenters/content_item/contents_list.rb#L118

1pretz1 (approved):
Looks good to me!

PR #2529, opened 2022-08-25
Remove code related to world location news article

BeckaL:
This content type is no longer in use, having been replaced by the 'news article' schema.  For example, if you look at some of the urls that are removed here in `config/govuk_examples.yml`, you can see that all of the `world_location_news_article` examples now redirect to a .../news/... url, and have a schema of `news_article`.

Whitehall does not publish anything with this schema anymore, and the few articles that were left behind with this schema were 'orphaned' content items which should have been previously redirected (but we suspect a bulk redirect task probably failed in a few cases, hence leaving behind 5 content items with this schema). This redirect has now been completed for all content items, meaning we have no remaining content items with this schema.

This means that code to render it can now be removed. This will also enable us to clean up the schema from govuk-content-schemas - PR [here](https://github.com/alphagov/govuk-content-schemas/pull/1118) (failing because of tests removed in the current PR here).

Query on non draft content store showing no remaining items:

```
criteria = ContentItem.where(schema_name: "world_location_news_article")
criteria.map{ | doc | doc.id }

=> []
```

There are two content items in the _draft_ content store - these also look like similar cases to the above, as they look 'orphaned' and the editions in whitehall link, and publish, to the version that does not use the world-location-news-article schema. Since these are both old content items, and on the draft stack, and are duplicates of the correct content items with the in-use schema (so if they are ever published, which is unlikely because they're from 2017, they will get published with the new schema and to the correct url), we've made the decision to leave these since we don't think they'll cause issues.

[Trello](https://trello.com/c/FtEA9kPn/187-remove-worldlocationnewsarticle-schema-and-related-code)


BeckaL:
Thanks so much Aga, really appreciate you having a look! And good spot on those remaining references 🕵️‍♀️, I'll remove those now!

PR #2535, opened 2022-08-31
Add single page notification button to document collections

hannako:
## What

Add the single page notification button to document collections. Pass in the `skip_account` flag to the component so that users sign up via the non GOV.UK account route. 

Review app: https://government-frontend-pr-2535.herokuapp.com/government/collections/technology-case-studies

| Before | After |
| --------| ---------- |
| Screenshot 2023-02-03 at 16 32 08 | Screenshot 2023-06-30 at 09 36 31|

Related work:

- https://github.com/alphagov/email-alert-frontend/pull/1489
- https://github.com/alphagov/govuk_publishing_components/pull/3229

Trello: https://trello.com/c/fXCk2Vck/1614-turn-on-email-notifications-for-document-collections-m

### DIFF
--- a/app/views/content_items/document_collection.html.erb
+++ b/app/views/content_items/document_collection.html.erb
@@ -22,6 +22,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
+<%= render 'shared/single_page_notification_button', content_item: @content_item %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/app/views/content_items/document_collection.html.erb
+++ b/app/views/content_items/document_collection.html.erb
@@ -22,7 +22,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
-<%= render 'shared/single_page_notification_button', content_item: @content_item %>
+<%= render 'shared/single_page_notification_button', content_item: @content_item, skip_account: "true" %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/app/views/shared/_single_page_notification_button.html.erb
+++ b/app/views/shared/_single_page_notification_button.html.erb
@@ -14,6 +14,7 @@
 %>
 
 <% ga4_data_attributes = ga4_data_attributes || default_ga4_data_attributes %>
+<% skip_account = skip_account || "false" %>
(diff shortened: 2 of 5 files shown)


Tetrino (approved):

app/views/content_items/document_collection.html.erb
```diff
@@ -22,7 +22,7 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
-<%= render 'shared/single_page_notification_button', content_item: @content_item %>
+<%= render 'shared/single_page_notification_button', content_item: @content_item, skip_account: "true" %>
```
BeckaL, line 25: Could you add the info you put into the PR description about what skipping the account actually does into this PR? I think it would help, it wasn't obvious to me but your explanation was v useful.

app/views/shared/_single_page_notification_button.html.erb
```diff
@@ -1,6 +1,9 @@
+<% skip_account = skip_account || nil %>
```
BeckaL, line 1: Is nil the right default here? Would we want to default to false?

BeckaL (commented):
LGTM, one question and one suggestion for a commit message addition

app/views/content_items/document_collection.html.erb
hannako: Yes good suggestion thank you, will do that 👍

app/views/shared/_single_page_notification_button.html.erb
hannako: I assumed it would be nil or "true" when I shipped [this](https://github.com/alphagov/govuk_publishing_components/pull/3229/files), but without any validation that those are the only two options, that is a bug waiting to happen isn't it!!! Because putting "false" as the `skip account` value passed to the component, will make [this](https://github.com/alphagov/govuk_publishing_components/blob/main/lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb#L64) line select the wrong form action. 

So I will update this line here to default to "false"
And I will update the component so it expects either "true" or "false".

hannako: https://github.com/alphagov/govuk_publishing_components/pull/3471

BeckaL (approved):
LGTM! 🚀

PR #2559, opened 2022-09-26
Update links in Webchat README

leenagupte:
After talking to HMPO and their new service provider, it became clear that they were looking at an old version of the code for the Webchat implementation and quoting incorrect polling times etc.

Removing the link to a specific commit means that we are always linking to the current version of the code.
There is a risk that the file could move and the link could break, but if that seems a more acceptable risk than having to also update these docs if any of the hardcoded values change.

Also removes the jQuery example code and replaces it with vanilla Javascript as jQuery has been removed from all of the frontend GOV.UK applications and we don't want to encourage its return.

### DIFF
--- a/docs/webchat.md
+++ b/docs/webchat.md
@@ -8,25 +8,23 @@ See the [reference implementation](https://github.com/alphagov/reference-webchat
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


sihugh (approved):

DilwoarH:
🥇

PR #2563, opened 2022-10-03
Remove unnecessary `mkdir /app` from Dockerfile.

sengi:
`WORKDIR` creates its directory anyway, and the mkdir fails now that we're creating `/app` in govuk-ruby-base. Thought I'd fixed all these before merging the change to govuk-ruby-base, but turns out I forgot to actually raise a bunch of the PRs that I'd prepared.


nsabri1 (approved):
LGTM

PR #2621, opened 2022-12-06
Bump Rails Translation Manager and use new plural file

1pretz1:
We're now using the extra plural rules defined by RTM as a prod
dependency [1]. This will replace the list defined in govuk_app_config
and so the reference has also been removed. RTM's version loads
automatically [2] and so we don't have to explicitly include the file in
the app.

Testing locally on a production rails server:

Fetching from plural file [1]:
```
[6] pry(#<ContentItemsController>)> plural_form = I18n.with_locale("cy") { I18n.t!("i18n.plural.keys") }.sort
=> [:few, :many, :one, :other, :two, :zero]
[7] pry(#<ContentItemsController>)> plural_form = I18n.with_locale("zh") { I18n.t!("i18n.plural.keys") }.sort
=> [:other]
[8] pry(#<ContentItemsController>)> plural_form = I18n.with_locale("dr") { I18n.t!("i18n.plural.keys") }.sort
=> [:one, :other]
```

Fetching from the default store (rails-i18n) [3]:
```
[11] pry(#<ContentItemsController>)> plural_form = I18n.with_locale("ar") { I18n.t!("i18n.plural.keys") }.sort
=> [:few, :many, :one, :other, :two, :zero]
```

[1]: https://github.com/alphagov/rails_translation_manager/blob/main/config/locales/plurals.rb
[2]: https://github.com/alphagov/rails_translation_manager/blob/f18fd2e178719dd0a103981160f898771957fa43/lib/rails_translation_manager.rb#L24
[3]: https://github.com/svenfuchs/rails-i18n/blob/ad26c50e1fea0eac2c9b85815f2c051f33a78045/rails/pluralization/ar.rb

### DIFF
--- a/config/locales/zh-hk.yml
+++ b/config/locales/zh-hk.yml
@@ -57,226 +57,152 @@ zh-hk:
       updated: 已更新
     schema_name:
       aaib_report:
-        one: 空中意外事故調查報告
         other: 其他空中意外事故調查報告
       announcement:
-        one: 公告
         other: 其他公告
       asylum_support_decision:
-        one: 庇護支援裁判決定
         other: 其他庇護支援裁判決定
       authored_article:
-        one: 撰寫文章
         other: 其他撰寫文章
       business_finance_support_scheme:
-        one: 商業融資支援計劃
         other: 其他商業融資支援計劃
       case_study:
-        one: 案例研究
         other: 其他案例研究
       closed_consultation:
-        one: 專屬諮詢
         other: 其他專屬諮詢
       cma_case:
-        one: 競爭及市場管理案例
         other: 其他競爭及市場管理案例
       coming_soon:
-        one: 即將推出之項目
         other: 其他即將推出之項目
       consultation:
-        one: 諮詢
         other: 其他諮詢
       consultation_outcome:
-        one: 諮詢結果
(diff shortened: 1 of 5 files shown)


config/locales/zh-hk.yml
```diff
@@ -57,226 +57,152 @@ zh-hk:
       updated: 已更新
     schema_name:
       aaib_report:
-        one: 空中意外事故調查報告
```
brucebolt, line 60: Hey @1pretz1, I just noticed that the `one` value for Chinese is the one we should've been keeping (translated to `Air Accident Investigation Reports`).  The value for `other` was always wrong (it translates to `Other Air Accident Investigation Reports`).  I think what we should've done was swap the value for `one` into the place of `other`, then delete `one`.

This seems to be the case for all Chinese plural forms across all the apps.

1pretz1: Below PRs should resolve the issue:
* https://github.com/alphagov/whitehall/pull/7160
* https://github.com/alphagov/collections/pull/3088
* https://github.com/alphagov/government-frontend/pull/2635

brucebolt: Thanks!  All approved.

PR #2626, opened 2022-12-08
Use bootsnap to improve startup time.

sengi:
We seem to be using [bootsnap](https://github.com/Shopify/bootsnap) on most of the publisher apps but not on the frontends.

We don't care much about startup time per se, but reducing the CPU spike on pod startup is helpful for capacity planning and helps to keep the overall performance of the cluster more predictable.

Tested: works on the integration k8s cluster. Benchmarking locally with `docker run` showed [\>40% improvement](https://docs.google.com/spreadsheets/d/13-acYdjnMcqV_JzrHR-fWFK8vYsM2aCXJuw6X_cRoAY/edit#gid=0) in startup time and savings of roughly 2 core-seconds per pod startup. Regression tested on EC2 by deploying to the integration environment, checking that smoke tests still pass there and checking there were no errors/warnings in app logs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,24 +3,22 @@ ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:3.1.2
 
 FROM $builder_image AS builder
 
-WORKDIR /app
-
+WORKDIR $APP_HOME
 COPY Gemfile* .ruby-version ./
 RUN bundle install
-
-COPY . /app
-RUN bundle exec rails assets:precompile && rm -rf /app/log
+COPY . ./
+RUN bundle exec bootsnap precompile --gemfile .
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -3,24 +3,22 @@ ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:3.1.2
 
 FROM $builder_image AS builder
 
-WORKDIR /app
-
+WORKDIR $APP_HOME
```
nsabri1, line 6: Maybe i am mistaken - but dont you need to define this `ARG APP_HOME /app`

theseanything (approved):

sengi:
Thanks! Since it's Friday afternoon and this thing's push-on-green I'll wait until Monday before merging :D

PR #2643, opened 2022-12-28
Handle webchat CSP modifications in application 

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

Webchat integrations on GOV.UK require modifications to the Content
Security Policy (CSP), these previously have been handled in
govup_app_config [1]. Overtime we have ended up with these falling out
of date, with more CSP entries than used cases. It is also a pain to
deploy these as you need to update the gem first.

This PR moves the CSP configuration into this application, which
resolves these pain points and follows the convention in our CSP
documentation [2] to not apply config that is just for a single
application.

This applies an approach where the CSP modification is only applied on
the webchat page, so it won't be broadcast across all other GOV.UK
pages. I set this up to meet the needs of the one webchat integration we
have which has a single entry to connect-src. In future we may well also
need to add a script-src integration as well. I also considered setting
up the ability that connect-src could contain an array, but didn't
implement it as there wasn't precedence of a webchat with this. It
shouldn't be hard to configure though if that situation occurs.

A slight difference to the integration in the govuk_app_config CSP is
the prefix of `https://` [3]. Which requires any connections to
omni.eckoh.uk use TLS. I'm assuming we just didn't know we could specify
this when we first set-up our CSP and the precedence remained.

Setting up an integration test for this functionality was a little bit
of a pain as the controller tests don't execute the CSP code and thus
don't allow testing. I imagine this would be resolved by switching to
Rails request testing [4]. I therefore set the testing up for this in a
Capybara integration test.

I noticed while developing this that the omni.eckoh.uk provider seems to
have an integration problem (the available resource returns 404) which I
deemed as out-of-scope for this change.

[1]: https://github.com/alphagov/govuk_app_config/blob/7f060692720df50a27f6845f052b04eae2246226/lib/govuk_app_config/govuk_content_security_policy.rb#L77-L84
[2]: https://docs.publishing.service.gov.uk/manual/content-security-policy.html
[3]: https://github.com/alphagov/govuk_app_config/blob/7f060692720df50a27f6845f052b04eae2246226/lib/govuk_app_config/govuk_content_security_policy.rb#L83
[4]: https://guides.rubyonrails.org/testing.html#integration-testing

---

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -13,6 +13,10 @@ class ContentItemsController < ApplicationController
 
   attr_accessor :content_item, :taxonomy_navigation
 
+  content_security_policy do |p|
+    p.connect_src(*p.connect_src, -> { csp_connect_src })
+  end
+
   def show
     load_content_item
 
@@ -211,5 +215,11 @@ private
     # Override the default from GovukPersonalisation::ControllerConcern so pages are cached on each flash message
     # variation, rather than caching pages per user
     response.headers["Vary"] = [response.headers["Vary"], "GOVUK-Account-Session-Exists", "GOVUK-Account-Session-Flash"].compact.join(", ")
+  end
+
+  def csp_connect_src
+    return if !@content_item || !@content_item.respond_to?(:csp_connect_src)
+
+    @content_item.csp_connect_src
   end
 end
--- a/app/models/webchat.rb
+++ b/app/models/webchat.rb
@@ -5,13 +5,14 @@ class Webchat
 
   validates :base_path, :open_url, :availability_url, presence: true
 
-  attr_reader :base_path, :open_url, :availability_url, :open_url_redirect
+  attr_reader :base_path, :open_url, :availability_url, :open_url_redirect, :csp_connect_src
 
   def initialize(attrs)
     @base_path = attrs["base_path"] if attrs["base_path"].present?
     @open_url = attrs["open_url"] if attrs["open_url"].present?
     @availability_url = attrs["availability_url"] if attrs["availability_url"].present?
     @open_url_redirect = attrs.fetch("open_url_redirect", nil)
+    @csp_connect_src = attrs["csp_connect_src"]
     validate!
   end
 
--- a/app/models/webchat.rb
+++ b/app/models/webchat.rb
@@ -8,10 +8,10 @@ class Webchat
   attr_reader :base_path, :open_url, :availability_url, :open_url_redirect, :csp_connect_src
 
   def initialize(attrs)
-    @base_path = attrs["base_path"] if attrs["base_path"].present?
-    @open_url = attrs["open_url"] if attrs["open_url"].present?
-    @availability_url = attrs["availability_url"] if attrs["availability_url"].present?
-    @open_url_redirect = attrs.fetch("open_url_redirect", nil)
+    @base_path = attrs["base_path"]
+    @open_url = attrs["open_url"]
+    @availability_url = attrs["availability_url"]
+    @open_url_redirect = attrs["open_url_redirect"]
     @csp_connect_src = attrs["csp_connect_src"]
     validate!
   end
--- a/docs/webchat.md
+++ b/docs/webchat.md
@@ -39,14 +39,18 @@ Once this shim is removed we can move this component into Static as a GOV.UK Pub
 - base_path: /government/contact/my-amazing-service
   open_url: https://www.my-amazing-webchat.com/007/open-chat
   availability_url: https://www.my-amazing-webchat.com/007/check-availability
+  csp_connect_src: https://www.my-amazing-webchat.com
 ```
 
 3. Deploy changes
 4. Go to https://www.gov.uk/government/contact/my-amazing-service
 5. Finished
 
-## CORS considerations
-To avoid CORS and CSP issues a new provider would need to be added to the [Content Security Policy](https://docs.publishing.service.gov.uk/manual/content-security-policy.html)
+## Content Security Policy considerations
+
+For a webchat provider to integrate with GOV.UK it needs permissions from the [Content Security Policy](https://docs.publishing.service.gov.uk/manual/content-security-policy.html).
+
+This will be set-up for a provider by the `csp_connect_src` option which configures the `connect-src` directive, however other providers may need additional configuration, such as `script-src`. This configuration should be done in the same manner as `csp_connect_src` to only affect resources that embed webchat.
 
 ## Required configuration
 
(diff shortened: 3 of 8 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -13,6 +13,10 @@ class ContentItemsController < ApplicationController
 
   attr_accessor :content_item, :taxonomy_navigation
 
+  content_security_policy do |p|
+    p.connect_src(*p.connect_src, -> { csp_connect_src })
```
Rosa-Fox, line 17: Just to check if I have understood correctly... in the past i've defined the CSP for the application in the initializers somewhere. In this case we want to change the CSP specific to different parts of the application (webchat), so this is instead setting the CSP for content items but it is overriden by this method in [contact_presenter](https://github.com/alphagov/government-frontend/blob/76071c744e4de0b5b5fed38de3758142d419fce5/app/presenters/contact_presenter.rb#L96) for webchat to allow the extra value? Otherwise `connect_src` is set to nil.

app/models/webchat.rb
```diff
@@ -5,13 +5,14 @@ class Webchat
 
   validates :base_path, :open_url, :availability_url, presence: true
 
-  attr_reader :base_path, :open_url, :availability_url, :open_url_redirect
+  attr_reader :base_path, :open_url, :availability_url, :open_url_redirect, :csp_connect_src
```
Rosa-Fox, line 8: 👍

app/models/webchat.rb
```diff
@@ -5,13 +5,14 @@ class Webchat
...
   def initialize(attrs)
-    @base_path = attrs["base_path"] if attrs["base_path"].present?
-    @open_url = attrs["open_url"] if attrs["open_url"].present?
-    @availability_url = attrs["availability_url"] if attrs["availability_url"].present?
-    @open_url_redirect = attrs.fetch("open_url_redirect", nil)
+    @base_path = attrs["base_path"]
+    @open_url = attrs["open_url"]
+    @availability_url = attrs["availability_url"]
+    @open_url_redirect = attrs["open_url_redirect"]
+    @csp_connect_src = attrs["csp_connect_src"]
```
Rosa-Fox, line 15: def looks better!

docs/webchat.md
```diff
@@ -84,3 +57,11 @@ The default response from the api as used by HMRC webchat provider is JSONP. To
...
   availability_payload_format: json
 ```
+
+### CSP connect-src
+
+Updates the Content Security Policy for pages that embed webchat to grant permission to make requests to the host specified. This should be in the form of a hostname, ideally with a scheme. For more information see, [connect-src](https://content-security-policy.com/connect-src/).
+
+```yaml
+  csp_connect_src: https://webchat.host
+```
```
Rosa-Fox, line 67: 👍

lib/webchat.yaml
```diff
@@ -1,4 +1,5 @@
 - base_path: /government/organisations/hm-passport-office/contact/hm-passport-office-webchat
   open_url: https://omni.eckoh.uk/v03.5/launcherV3.php?p=HMPO&d=717&ch=CH&psk=chat_a1&iid=STC&srbp=0&fcl=0&r=Chat&s=https://omni.eckoh.uk/v03.5&u=&wo=&uh=&pid=2&iif=0
   availability_url: https://omni.eckoh.uk/v03.5/providers/HMPO/api/availability.php
+  csp_connect_src: https://omni.eckoh.uk
```
Rosa-Fox, line 4: 👍

test/integration/contact_test.rb
```diff
@@ -53,4 +53,32 @@ class ContactTest < ActionDispatch::IntegrationTest
...
+    parsed_csp = page.response_headers["Content-Security-Policy"]
+                     .split(";")
+                     .map { |directive| directive.strip.split(" ") }
+                     .each_with_object({}) { |directive, memo| memo[directive.first] = directive[1..] }
+
+    assert_includes parsed_csp["connect-src"], "https://webchat.host"
+
+    # reset back to default driver
+    Capybara.use_default_driver
+  end
```
Rosa-Fox, line 83: 👍

test/models/webchat_test.rb
```diff
@@ -14,6 +15,7 @@ class WebchatTest < ActiveSupport::TestCase
     assert_equal(instance.base_path, webchat_config["base_path"])
     assert_equal(instance.open_url, webchat_config["open_url"])
     assert_equal(instance.availability_url, webchat_config["availability_url"])
+    assert_equal(instance.csp_connect_src, webchat_config["csp_connect_src"])
```
Rosa-Fox, line 18: 👍

Rosa-Fox (approved):
Hi Kevin,

Apologies for the delay, I spent some time understanding webchat and reading through the various docs linked on Trello (which were really helpful). I think it looks good. I tested it out locally:

With the data in the YAML:
Screenshot 2022-12-29 at 12 03 39

Updating the YAML to https://www.my-amazing-webchat.com:

Screenshot 2022-12-29 at 12 05 44

A non-webchat page:

Screenshot 2022-12-29 at 12 06 23

app/controllers/content_items_controller.rb
kevindew: Yup exactly that. 

We configure a global CSP for the app via [the initializer](https://github.com/alphagov/government-frontend/blob/1175a1a8d5ed3401bcf48dacd404ca7cf9753f7d/config/initializers/content_security_policy.rb#L1) and then we can refine it a controller or action level. 

This allows us to have three layers of CSP:
1) In govuk_app_config, directives that cover all GOV.UK apps. These are a pain to change and test because they require a new gem release, but are good to apply one rule consistently everywhere. Best used for a change that applies to multiple apps e.g. analytics or govspeak
2) Across an app, via the initializer. This offers us a chance to apply configuration that applies to just one app and every route in it. Much faster to change than the gem. Best used for a case where an app uses a library that other apps don't. There are no cases of this used yet, but I'm expecting it to be used so that apps without jQuery can be configured more securely than the apps that still use jQuery (licence-finder and Whitehall).
3) In a controller, which can be specified to an action level, this allows just a single resource to set up extra rules for the CSP which is best for when there's an exception to the rules for a couple of pages, i.e this scenario. We have precedence for doing this in Smart Answers already: https://github.com/alphagov/smart-answers/blob/45a463407b84ed3e58b268f4522d7493dbdd40ba/app/controllers/smart_answers_controller.rb#L8-L12

Rosa-Fox: ok cool, just wanted to check I wasn't missing something. Thanks for the explanation, makes sense :)

PR #2661, opened 2023-01-13
Fix layout of /government/history and remove use of inline styles

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

The changes here have been made in order for this application to be compatible with the changes coming in https://github.com/alphagov/govuk_app_config/pull/279, where inline styles will no longer be permitted. The scope grew a little to do a bit of a clean-up in the history pages where inline styles were present.

This change:
- removes classes that have no effect
- fixes a layout bug
- replaces inline styles with a class

The layout bug is demonstrated below

Before:

Screenshot 2023-01-13 at 18 25 12

After:

Screenshot 2023-01-13 at 18 31 42

---

### DIFF
--- a/app/views/histories/10_downing_street.html.erb
+++ b/app/views/histories/10_downing_street.html.erb
@@ -21,7 +21,7 @@
 
 <div class="govuk-grid-row">
   <nav class="govuk-grid-column-one-third">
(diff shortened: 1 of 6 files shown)


patrickpatrickpatrick (approved):
Have compared all the pages that are changed here and there doesn't appear to be any styling discrepancies on desktop and mobile.

PR #2662, opened 2023-01-14
Dockerfile cleanup.

sengi:
- Use bootsnap to reduce startup resource usage spike.
- Parameterise Ruby version.
- Make better use of WORKDIR.
- Use env vars from the base image where appropriate, instead of hardcoding paths.
- Remove unnecessary use of `bundle exec`.
- Update .dockerignore.

Tested: app boots successfully with `docker run`.


marcpomfret (approved):

nsabri1 (approved):
lgtm

PR #2705, opened 2023-03-01
Adds top margin to title component inside inverse-header

davidtrussler:
[Github issue](https://github.com/alphagov/govuk_publishing_components/issues/3085)

This PR addresses a problem where the inverse header displays with uneven vertical spacing. This is caused by the way the component is rendered on the page rather than the component itself and requires that some extra padding is added to the top of the component so that it displays in a satisfactorily balanced way. 

|Current|Updated|
|-|-|
|heading_before|heading_after|

### DIFF
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -37,7 +37,8 @@
         title: @content_item.title,
         context: I18n.t("content_item.schema_name.#{@content_item.format_sub_type}", count: 1),
         inverse: true,
-        margin_bottom: 0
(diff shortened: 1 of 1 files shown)


andysellick (approved):
Good fix. The `margin_top` option doesn't appear to be documented in the title component, as a related piece of work, would you be able to add that?

davidtrussler:
> Good fix. The `margin_top` option doesn't appear to be documented in the title component, as a related piece of work, would you be able to add that?

Oh yes, that's true. Yes, I'll do that as well. 👍

PR #2712, opened 2023-03-09
Refactor Get Involved

brucebolt:
This page was created differently to all the other content types.

Refactoring so it is the same as everything else.

The test coverage isn't as good as it should be, but retains the same level of testing as before the refactor.

Tests depend on https://github.com/alphagov/publishing-api/pull/2312.


BeckaL (approved):
Looks good to me, nice improvement ⭐

brucebolt:
> It's a shame the review app doesn't work - is that something to do with search? - but I checked it out locally using the start up script (retro!) and it all looked great.

Yes, I think it's because this app's Heroku setup doesn't include the env var to provide the URL for live search.  This is the only document type in Government Frontend that uses search, so maybe it's not the right app for the document to be rendered by 🤔.

app/presenters/get_involved_presenter.rb
```diff
@@ -1,2 +1,111 @@
...
+    Services.search_api.search(query)["results"]
+  end
+
+  def filtered_links(array, close_status)
+    array.map do |item|
+      {
+        link: {
+          text: item["title"],
+          path: item["link"],
+          description: "#{close_status} #{item['end_date'].to_date.strftime('%d %B %z')}",
```
AgaDufrat, line 90: I think we may need to change the time formatting here
```suggestion
          description: "#{close_status} #{item['end_date'].to_date.strftime('%d %B %Y')}",
```

AgaDufrat: Oops, too late!

brucebolt: Good spot!  I don't know how that changed, since I just copy and pasted the stuff over 🤷‍♂️.

brucebolt: I'll raise a PR to fix this.

brucebolt: https://github.com/alphagov/government-frontend/pull/2714

hannako:
> > It's a shame the review app doesn't work - is that something to do with search? - but I checked it out locally using the start up script (retro!) and it all looked great.
> 
> Yes, I think it's because this app's Heroku setup doesn't include the env var to provide the URL for live search. This is the only document type in Government Frontend that uses search, so maybe it's not the right app for the document to be rendered by 🤔.

This fixes the review apps https://github.com/alphagov/government-frontend/pull/2715

PR #2721, opened 2023-03-22
Refactor manuals

patrickpatrickpatrick:
## What

Refactor the manual and HMRC manual partials so that
- they have common layout partials with the mark-up that is common to both of them
- they use the same header partial but with an optional attribute that can change the supertext and the background colour
- the header component now is outside the main container
- add tests to prevent regression of having multiple elements with id of 'content' on page

## Why

Pages in manuals (like the [content design manual](https://www.gov.uk/guidance/content-design/what-is-content-design) or the [HMRC manual](https://www.gov.uk/hmrc-internal-manuals/paye-manual/paye70215)) have a duplicate ID of “content“ and therefore a somewhat broken skip link. The skip link used to go to the second “content“ ID, the one after the second header and second breadcrumbs. And the main element was [fixed at some point](https://github.com/alphagov/manuals-frontend/pull/969) to start there too, which seems to have been reverted as well.

[Trello Card](https://trello.com/c/CuRc02VB/1697-bug-no-way-to-skip-to-main-manual-content)


patrickpatrickpatrick:
Replaced by these PRs

- https://github.com/alphagov/government-frontend/pull/2727
- https://github.com/alphagov/government-frontend/pull/2728
- https://github.com/alphagov/government-frontend/pull/2729

PR #2750, opened 2023-04-20
Replace consultation view styles with Design System styles

jon-kirwan:
## What

https://trello.com/c/fc41oyEK/1923-enable-individual-loading-of-stylesheets-in-government-frontend

Remove unused **consultation** view stylesheets as a pre-requisite to implementing the AssetHelper to load component and view style sheets only required on the page being viewed.

## Why

So there will be less document types to be reviewed and updated when implementing the helper.

## Visual changes

None

## Anything else

- https://github.com/alphagov/govuk_publishing_components/pull/3014
- https://github.com/alphagov/frontend/pull/3342
- https://github.com/alphagov/smart-answers/pull/6213
- [RFC #149](https://github.com/alphagov/govuk-rfcs/pull/152)


app/assets/stylesheets/views/_consultation.scss
```diff
@@ -1,15 +0,0 @@
...
-
-  .consultation-date {
-    @include govuk-font(19, $weight: bold);
-  }
-
-  .original-consultation {
-    border-top: 1px solid $govuk-border-colour;
-
-    @include govuk-media-query($from: tablet) {
-      padding-top: govuk-spacing(6);
```
MartinJJones, line 12: When testing locally, there is no white space at the top of the consultation component. I imagine we can fix this by adding a utility class to the HTML.

When testing on mobile, we still have white space at the top of the element, this is because [`mobile_top_margin` is set to true](https://github.com/alphagov/government-frontend/blob/54b3ea5fc89e07ae9401f81855f447060b52bebc/app/views/content_items/consultation.html.erb#L56).

So it appears we have 2 different approaches for adding some space to the top of the element, mobile is using margin-top and desktop is using padding-top.

jon-kirwan: Thanks, good spot. Should now be fixed by adding `govuk-!-margin-top-6`.

MartinJJones (approved):
Looks good to me 👍

PR #2793, opened 2023-05-23
Render Worldwide Offices

jkempster34:
https://trello.com/c/0r6EWhlK

## Visual/behaviour differences
- Now shows breadcrumbs
- The `location` links take you to the news page. E.g., `https://www.gov.uk/world/usa/news` instead of `https://www.gov.uk/world/usa`. This is because I used the existing `WorldLocationBasePath`, and I figured we would want to keep things consistent ([see example "Spain" link here](http://government-frontend.dev.gov.uk/government/case-studies/doing-business-in-spain.es))
- Some slight (imperceptible) changes to some padding (couple of pixels where Whitehall uses it's own `$gutter` constant, and I've replaced it with `$govuk-gutter`)
- Some slight changes to the address formatting. See the Iceland pic below, where an extra `<br>` has snuck in

## Screenshots
||Production|This PR|
|-|-|-|
|Standard logo and two short locations|image|image|
|No access details, no locations|image|image|
|Different logo and logo colour|image|image|
|Locations hidden through JavaScript|image|image|

### Add Worldwide Organisations and Offices to README
The README contains a list of content items and their production status.

Now that the migration of Worldwide Organisations and Offices is underway, we
can add them in.

### Add Rinku for auto linking
Worldwide office comments make use of the `auto_link` method. This used to be
part of Rails but was deprecated.

This adds the Rinku gem, which is a drop in replacement for `auto_link`. 

This matches the implementation of Organisations in the Collections
application. We may be able to refine this further in future by autolinking the
comments before publishing in Whitehall.

### Add locales for WorldwideOffices
This adds the necessary locales to render the contacts for WorldwideOffices.

### Render title and body for Worldwide Offices
This renders a basic Worldwide Office page including the browser title and the
main body of access and opening times.

We override the `show_contents_list?` method so that a contents list is always
shown as it is in Whitehall.

### Add Govspeak gem for formatting contacts
It's important that the addresses of Worldwide locations are formatted
correctly according to the country that they are in.

The `Govspeak` gem includes a presenter with this functionality.

### Render contacts for Worldwide Offices
WorldwideOffices and WorldwideOrganisations have linked Contacts.

Though there is an existing ContactPresenter, this is quite specific to the
contacts pages. This adds a presenter to represent the linked contacts of
WorldwideOffices and WorldwideOrganisations. It is namespaced under
WorldwideOrganisations as they are the parent of WorldwideOffices.

### DRY ContactPresenter and WorldwideOrganisation::LinkedContactPresenter
The `contact` schema is used for both contact pages, as well as the contacts
related to WorldwideOffices and WorldwideOrganisations.

However, the ContactPresenter and WorldwideOrganisation::LinkedContactPresenter
are quite separate in their implementations.

This adds a module to DRY up any common code.

We may want to consider using the AddressFormatter for Contacts as a further
refinement.

This adds a module to DRY up any common code.

### Render office title for Worldwide Offices
The h2 for this page comes from the title of the Worldwide Offices Contact.

### Render header section for Worldwide Offices
This adds rendering for the header section of Worldwide Offices.

The logo element displays the title and logo of the first sponsoring
organisation, as well as the name of the Worldwide Organisation.

The "location" and "part of" sections display the world locations and
sponsoring organisations associated with this office.

### Add JavaScript to hide extra world location links
The "hide other links" JavaScript from Whitehall is used to hide long lists of
world locations in the header of the Worldwide Office and Worldwide
Organisation pages.

This moves it across to government-frontend, removing any jQuery, unneeded
features and configurations options. I have rewritten the module in the style
of the other JavaScript in this application.

---


app/lib/address_formatter.rb
```diff
@@ -0,0 +1,66 @@
+class AddressFormatter
```
brucebolt, line 1: The govspeak gem has a feature called [`HCardPresenter`](https://github.com/alphagov/govspeak/blob/3027b9b3063f5495891e16ca830b909bbea43ce5/lib/govspeak/presenters/h_card_presenter.rb#L16), which seems to do much the same.

I wonder if it would be possible to use that, instead of copying over the formatter from Whitehall?

brucebolt (commented):
This looks good to me.

I've left a question about whether we can use the govspeak gem, instead of copying `AddressFormatter` from Whitehall.

Also probably worth checking with `#govuk-navigation-tech` when getting a review from them whether they want to keep the behaviour of hiding some of the world locations using JavaScript.

app/lib/address_formatter.rb
jkempster34: That looks like the same thing, I'll have a look now thanks! 👀

jkempster34: Would you mind having a second look at this? It looks like that worked ☝️. Only difference is what you can see on the image of the Iceland office which is probably inconsequential?

brucebolt: This all looks good to me and will make these consistent with other parts of GOV.UK that have used the govspeak gem.

brucebolt (approved):

app/assets/javascripts/modules/hide-other-links.js
```diff
@@ -0,0 +1,87 @@
...
+
+      if (node.nodeName.toLowerCase() === 'a') {
+        linksCount += 1
+      }
+    }, this)
+  }
+
+  HideOtherLinks.prototype.createShowLink = function () {
+    var hiddenCount = this.hiddenElementContainer.children.length
+    var linkText = `+ ${hiddenCount} other${hiddenCount > 1 ? 's' : ''}`
```
patrickpatrickpatrick, line 50: Sorry to nitpick but template strings aren't supported by IE 11 so you would need to use string concatenation here.

jkempster34: 👍 I'll update that thanks! Does the rest of this file look okay? I haven't written JS in this style before

app/assets/stylesheets/views/_worldwide-organisation.scss
```diff
@@ -1,10 +1,51 @@
+// stylelint-disable max-nesting-depth
 .worldwide-organisation-header {
```
patrickpatrickpatrick, line 2: This is a continuation of what was there before but as you are adding more to it perhaps it would be best if this was written to be more in-line with [guidance in the GDS Way](https://gds-way.cloudapps.digital/manuals/programming-languages/css.html#sass-nesting). Personally I think 'show-other-content', 'logo' and 'metadata' could be moved from out of worldwide-organisation-header and the BEM naming convention could be used here (worldwide-organisation-header__logo for instance).

jkempster34: Great,

I've made this change as a separate commit. I've left the `show-other-content` class as-is because it's related to JS, but I've nested it in a `js-enabled` so it's clear.

app/assets/javascripts/modules/hide-other-links.js
jkempster34: Change made

patrickpatrickpatrick (approved):
Looks good to me!

PR #2833, opened 2023-06-19
Text algin right table headers for RTL content items

davidgisbey:
## Description

We've had a zendesk ticket come in https://govuk.zendesk.com/agent/tickets/5368540 which has reported an issue with table headings in RTL languages. While the body of the table renders RTL the headers don't.

I've updated the stylesheets to ensure that TH elements use `text-align: start;` which resolves the issue.

### Before

image

### After

image


maxgds (approved):
Nice. There's probably some work to check other alignments throughout our codebases but that's not a "now" problem.

PR #2837, opened 2023-06-26
Change font size of second H1 on guide pages

patrickpatrickpatrick:
## What

Change the font size of the heading on guide pages by using a heading component with the font size `govuk-heading-l`.

## Why

This type of page features a title (rendered by the application) and a body of HTML that is saved to the content item in Whitehall and then rendered using Govspeak. This was leading to an accessibility issue where the page had two differently styled H1s. If we wanted to ensure that the page had one H1 then all the content that is rendered in the guides would need to be re-edited to update the heading levels. In addition we would need to ensure that future guides used this offset of headings. The solution that has the least amount of overhead (and would also ensure this issue does not crop up again for new pages) was to change the stying of the second H1 to resemble the first H1. Having two H1s on the page is acceptable if they both resemble H1s and do not differ in styling.

### Before

Screenshot 2023-06-26 at 11 17 48

### After

Screenshot 2023-06-26 at 11 17 57

[Related Trello Card
](https://trello.com/c/c1iqDv2I/1737-two-h1s-on-the-same-page-that-look-like-different-levels-2-m)

### DIFF
--- a/app/views/content_items/guide.html.erb
+++ b/app/views/content_items/guide.html.erb
@@ -56,9 +56,7 @@
   <div class="govuk-grid-column-two-thirds" id="guide-contents">
     <% if @content_item.has_parts? %>
       <% if @content_item.show_guide_navigation? %>
-        <h1 class="part-title">
-          <%= @content_item.current_part_title %>
-        </h1>
+        <%= render 'govuk_publishing_components/components/heading', heading_level: 1, font_size: 'l', margin_bottom: 6, text: @content_item.current_part_title %>
       <% end %>
       <%
         disable_youtube_expansions = true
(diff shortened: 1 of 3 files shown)


app/views/content_items/guide.html.erb
```diff
@@ -56,7 +56,7 @@
   <div class="govuk-grid-column-two-thirds" id="guide-contents">
     <% if @content_item.has_parts? %>
       <% if @content_item.show_guide_navigation? %>
-        <h1 class="part-title">
+        <h1 class="govuk-heading-l">
```
andysellick, line 59: Have you considered using the [heading component](https://components.publishing.service.gov.uk/component-guide/heading)?

patrickpatrickpatrick: ah yeah, that would make sense!

andysellick (approved):
Bit of a thorny one this, but this seems like the simplest solution for now. I wonder if there's a future alternative where we could use something like the [title component with context](https://components.publishing.service.gov.uk/component-guide/title/with_context) to combine the two H1s, but that's a whole other conversation.

PR #2845, opened 2023-06-30
Don't show "Part of:" tag when sponsoring organisations are absent

jkempster34:
Currently, when sponsoring organisations are absent for a worldwide organisation (this isn't true for any currently, but can be), we still show the "Part of:" description in the header.

This makes it conditional on the presence of the sponsoring organisations, and adds some additional testing around this.

|Before|After|
|-|-|
|image|image|


JonathanHallam (approved):
Thanks :)

PR #2864, opened 2023-07-14
Fix text for component

MuriloDalRi:
https://govuk.zendesk.com/agent/tickets/5406038

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -43,7 +43,7 @@ en:
     original_consultation: Original consultation
     ran_from: This consultation ran from
     respond_online: Respond online
-    visit_soon: Visit this page again soon to download the outcome to this public&nbsp;feedback.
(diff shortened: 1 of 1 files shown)


deborahchua (approved):

config/locales/en.yml
```diff
@@ -43,7 +43,7 @@ en:
     original_consultation: Original consultation
     ran_from: This consultation ran from
     respond_online: Respond online
-    visit_soon: Visit this page again soon to download the outcome to this public&nbsp;feedback.
+    visit_soon: Visit this page again soon to download the outcome to this public feedback.
```
davidtrussler, line 46: The `&nbsp;` character removed here is present in all the other locales (as far as I can see). Most of them will probably never be viewed but I would guess at least the `cy` one would be. So strictly speaking I feel we should change some/all of these as well.

PR #2899, opened 2023-08-15
Enable taxonomy email signup for document collection

catalinailie:
Allow users to sign up to a mapped taxonomy topic email subscription from some document collection pages if `taxonomy_topic_email_override` is present.

Trello card: https://trello.com/c/uvbb9UYR/1893-document-collection-template-uses-different-email-signup-component-if-taxonomytopicemailoverride-is-present-l

Without `taxonomy_topic_email_override` 
Screenshot 2023-08-23 at 15 25 33

With `taxonomy_topic_email_override`
Screenshot 2023-08-23 at 16 23 47

### DIFF
--- a/app/presenters/document_collection_presenter.rb
+++ b/app/presenters/document_collection_presenter.rb
@@ -63,6 +63,10 @@ class DocumentCollectionPresenter < ContentItemPresenter
     )
   end
 
+  def taxonomy_topic_email_override_base_path
+    content_item.dig("links", "taxonomy_topic_email_override", 0, "base_path")
+  end
+
 private
 
   def group_documents(group)
--- a/app/views/content_items/document_collection.html.erb
+++ b/app/views/content_items/document_collection.html.erb
@@ -22,7 +22,11 @@
 </div>
 
 <%= render 'shared/publisher_metadata_with_logo' %>
-<%= render 'shared/single_page_notification_button', content_item: @content_item, skip_account: "true" %>
+<% if @content_item.taxonomy_topic_email_override_base_path.present? %>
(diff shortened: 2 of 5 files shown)


app/presenters/document_collection_presenter.rb
```diff
@@ -5,6 +5,7 @@ class DocumentCollectionPresenter < ContentItemPresenter
   include ContentItem::TitleAndContext
   include ContentItem::ContentsList
   include ContentItem::SinglePageNotificationButton
+  include ContentItem::TaxonSignupLink
```
hannako, line 8: I can see you've followed the pattern here of requiring a module to implement the logic for the `TaxonSignupLink`. But it's meant you've had to make `taxonomy_topic_email_override` a method available on the parent `content_item_presenter` class. Which means all the subclasses can now access that (and we'd probably need a test in the ContentItemPresenterTest etc). 

A module would be needed if we had logic do decide whether or not to use the override (like the exemption list in the SinglePageNotificationButton module), or if we knew we'd need this feature for other document types in the future.

But as that's not the case, I think it might be safer, and also less confusing to future us, if you just add the code to the `document_collection_presenter` class itself. So it's only available in that one place, and nothing else needs to change.

```
def taxonomy_topic_email_override_base_path
  content_item.dig("links","taxonomy_topic_email_override","base_path")
end
```

And then in the view you could just check if the attribute is `present?` to decide which partial to show. 

Does that make sense and do you agree?

catalinailie: Yes, you are right. I first used that approach before realising we need to do this only for document collections.

AshGDS (approved):

hannako (approved):

PR #2900, opened 2023-08-17
Add tracking for email subscriptions completion

andysellick:

## What
Adds GA4 tracking for notifications about subscribing/unsubscribing to emails about some pages, specifically:
- [detailed guides](https://www.gov.uk/guidance/cost-of-living-payment)
- call for evidence
- consultation
- [publications](https://www.gov.uk/government/publications/form-pa14-medical-certificate-probate)

These events are generated using the auto tracker, with attributes in the notification, so it should only fire when the notification is visible.

Detailed guide with the notification looks like this:

Screenshot 2023-08-17 at 09 27 01

## Why
Part of the GA4 migration.

## Visual changes
None.

Trello card: https://trello.com/c/TXQOqtZ0/628-email-subscriptions-manage-subscriptions-sign-in-change-unsubscribe-forms

### DIFF
--- a/app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
+++ b/app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
@@ -5,11 +5,27 @@
 <% if show_email_subscription_success_banner?(@account_flash) %>
   <div class="govuk-grid-row govuk-!-margin-top-3">
     <div class="govuk-grid-column-two-thirds">
-      <%= render "govuk_publishing_components/components/success_alert", {
-        message: email_subscription_success_banner_heading(@account_flash),
-        description: banner_description,
-        margin_bottom: 0,
-      } %>
+      <%
+        ga4_data = {
+          event_name: "form_complete",
+          type: "email subscription",
+          text: email_subscription_success_banner_heading(@account_flash, :en),
+          section: local_assigns[:title],
+          action: "complete",
+          tool_name: "Get emails from GOV.UK"
+        }.to_json
+      %>
+      <%= content_tag(:div,
+        data: {
+          module: "ga4-auto-tracker",
+          ga4_auto: ga4_data
+        }) do %>
+        <%= render "govuk_publishing_components/components/success_alert", {
+          message: email_subscription_success_banner_heading(@account_flash),
+          description: banner_description,
+          margin_bottom: 0,
+        } %>
+      <% end %>
     </div>
   </div>
 <% end %>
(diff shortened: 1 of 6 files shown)


app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -5,11 +5,27 @@
...
-      <%= render "govuk_publishing_components/components/success_alert", {
-        message: email_subscription_success_banner_heading(@account_flash),
-        description: banner_description,
-        margin_bottom: 0,
-      } %>
+      <%
+        ga4_data = {
+          event_name: "form_complete",
+          type: "email subscription",
+          text: email_subscription_success_banner_heading(@account_flash),
```
AshGDS, line 12: It looks like `email_subscription_success_banner_heading` calls the translation function. It's probably worth ensuring the English translation is used for this text instead.

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
```diff
@@ -5,11 +5,27 @@
...
+          tool_name: "Get emails from GOV.UK"
+        }.to_json
+      %>
+      <%= content_tag(:div,
+        data: {
+          module: "ga4-auto-tracker",
+          ga4_auto: ga4_data
+        }) do %>
+        <%= render "govuk_publishing_components/components/success_alert", {
+          message: email_subscription_success_banner_heading(@account_flash),
```
AshGDS, line 24: Similar to the other comment regarding translation

AshGDS (changes requested):
Looks good :+1: Hopefully this isn't too much of a headache, but it looks like some of the text used might get translated so I've requested changes in case this can be resolved. Thanks

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
AshGDS: More specifically, it calls the `t()` function - is that different to I81n.t?

andysellick: I think the first one might be an alias for the second, but I'm not 100% certain.

Have altered the method to supply a specific locale if requested, good spot 👍

app/views/shared/_email_subscribe_unsubscribe_flash.html.erb
AshGDS: Whoops, my mistake!

AshGDS (approved):
lgtm :+1:

PR #2906, opened 2023-08-24
Filter people roles by organisation

jkempster34:
> [!WARNING]  
> 🚨 Do not merge! 🚨 
> Depends on https://github.com/alphagov/whitehall/pull/8162 being merged and all worldwide organisations being republished

https://trello.com/c/54I3mDxr

People linked to a worldwide organisation may have multiple roles, some of which are related to other organsaitions. Currently, we are incorrectly displaying these roles on the world wide organisation pages.

This copies the approach taken in alphagov/collections#1719. We link to all roles from the worldwide organisation, then filter out the incorrect ones in the rendering app.

See also https://github.com/alphagov/publishing-api/pull/2478 and https://github.com/alphagov/whitehall/pull/8162.

---

### DIFF
--- a/test/presenters/worldwide_organisation_presenter_test.rb
+++ b/test/presenters/worldwide_organisation_presenter_test.rb
@@ -6,13 +6,23 @@ class WorldwideOrganisationPresenterTest < PresenterTestCase
   end
 
   test "description of primary_role_person should have spaces between roles" do
-    presenter = create_presenter(WorldwideOrganisationPresenter, content_item: { "links" => { "primary_role_person" => [{ "details" => { "image" => {} }, "links" => { "role_appointments" => [{ "details" => { "current" => true }, "links" => { "role" => [{ "title" => "Example Role 1" }] } }, { "details" => { "current" => true }, "links" => { "role" => [{ "title" => "Example Role 2" }] } }] } }] } })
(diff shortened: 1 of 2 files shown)


test/presenters/worldwide_organisation_presenter_test.rb
```diff
@@ -6,15 +6,25 @@ def schema_name
   end
 
   test "description of primary_role_person should have spaces between roles" do
-    presenter = create_presenter(WorldwideOrganisationPresenter, content_item: { "links" => { "primary_role_person" => [{ "details" => { "image" => {} }, "links" => { "role_appointments" => [{ "details" => { "current" => true }, "links" => { "role" => [{ "title" => "Example Role 1" }] } }, { "details" => { "current" => true }, "links" => { "role" => [{ "title" => "Example Role 2" }] } }] } }] } })
+    presenter = create_presenter(WorldwideOrganisationPresenter, content_item: { "links" => { "primary_role_person" => [{ "details" => { "image" => {} }, "links" => { "role_appointments" => [{ "details" => { "current" => true }, "links" => { "role" => [{ "content_id" => "1", "title" => "Example Role 1" }] } }, { "details" => { "current" => true }, "links" => { "role" => [{ "content_id" => "2", "title" => "Example Role 2" }] } }] } }], "roles" => [{ "content_id" => "1" }, { "content_id" => "2" }] } })
```
CristinaRO, line 9: This is an existing ailment of the test, but would there be value in adding a test to check that when a person doesn't have any current roles for this org the presenter doesn't blow up?

jkempster34: If the persons role isn't in the filter list, they won't be in the content item at all, so this should be a non-problem. I'm wary of adding too many tests if they aren't needed.

I did wonder what happens if a role appointment doesn't have a role, but I don't think that is possible either, and that would have errored previously too

CristinaRO (approved):

PR #2922, opened 2023-09-12
Reflect travel advice alert status from travel advice publisher on the advice page

hannalaakso:
## What
Publishers can now set an alert status for a country programmatically instead of including it in the copy content (see screenshot below for the editor options). 

4_alert_status_options

This PR:
- Removes legacy code to render travel alerts as part of the govspeak content. To our knowledge this code has never been used as it wasn't possible to set `alert_status` in the editor until recently. Removes formatting from the presenter which was customised for rendering the alerts in govspeak: it used to flatten the array of alert statuses and wrap them each in a paragraph tag. These changes are in readiness for rendering the alerts with warning text.
- Renders alert statuses with warning text component
- Wraps the alert status in govspeak because the alert copy uses the abbreviation markup as recommended by the GOV.UK styleguide. Wrapping the alerts in govspeak applies the govspeak styling for acronyms.
- Runs automated tests to check that the alert statuses are rendering correctly

## Why
This will make it easier for publishers to update the alert status, keep it consistent in implementation, make it more visible for users and improve the semantics of the page.

## Further information
As these alerts will replace the alerts previously set in the body copy, publishers will need to remove any alerts from the body copy when they set the status using the travel advice publisher checkboxes. 

This change means that whilst travel alerts were previously rendered using the govspeak callout component, they will now be rendered using the warning text. The warning text contains visually hidden text to help screen reader users understand the content of the alerts.

Since the editor options were implemented as checkboxes, a publisher could set multiple statuses even though only one status will probably ever be set. The logic should therefore work with up to four options. We might change the options to radios in a further iteration.

## Before
Screenshot 2023-09-13 at 18 23 42

## After
Screenshot 2023-09-13 at 18 23 03

Fixes https://trello.com/c/Y4fruQto/1987-reflect-travel-advice-alert-status-from-travel-advice-publisher-on-the-advice-page-m

### DIFF
--- a/app/views/content_items/travel_advice.html.erb
+++ b/app/views/content_items/travel_advice.html.erb
@@ -16,6 +16,12 @@
   <div class="govuk-grid-column-two-thirds travel-advice__header">
     <%= render 'govuk_publishing_components/components/title', @content_item.title_and_context %>
 
+    <% @content_item.alert_status.each do |text| %>
+      <%= render "govuk_publishing_components/components/warning_text", 
+        text: text 
+      %> 
+    <% end %>
+
     <aside class="part-navigation-container" role="complementary">
       <%= render "govuk_publishing_components/components/contents_list", aria: { label: t("travel_advice.pages") }, contents: @content_item.part_link_elements, underline_links: true, ga4_tracking: true %>
 
--- a/app/views/content_items/travel_advice.html.erb
+++ b/app/views/content_items/travel_advice.html.erb
@@ -16,10 +16,13 @@
   <div class="govuk-grid-column-two-thirds travel-advice__header">
     <%= render 'govuk_publishing_components/components/title', @content_item.title_and_context %>
 
-    <% @content_item.alert_status.each do |text| %>
-      <%= render "govuk_publishing_components/components/warning_text", 
-        text: text 
-      %> 
+   <%= render "govuk_publishing_components/components/govspeak", {
+    } do %>
+       <% @content_item.alert_status.each do |text| %>
+        <%= render "govuk_publishing_components/components/warning_text", 
+          text: text 
+        %> 
+      <% end %>
     <% end %>
 
     <aside class="part-navigation-container" role="complementary">
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -5,5 +5,4 @@ data:
 ignore_unused:
   - 'content_item.schema_name.*.{one,other,zero,few,many,two}'
   - 'corporate_information_page.*'
-  - 'travel_advice.alert_status.*'
   - 'publication.documents.other'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -525,10 +525,10 @@ en:
       short_ordinal: "%e %B %Y"
   travel_advice:
     alert_status:
-      avoid_all_but_essential_travel_to_parts_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all but essential travel to parts of %{country}.
-      avoid_all_but_essential_travel_to_whole_country_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all but essential travel to %{country}.
-      avoid_all_travel_to_parts_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all travel to parts of %{country}.
-      avoid_all_travel_to_whole_country_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all travel to %{country}.
+      avoid_all_but_essential_travel_to_parts_html: <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advises against all but essential travel to parts of %{country}.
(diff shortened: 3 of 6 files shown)


config/locales/en.yml
```diff
@@ -525,10 +525,10 @@ en:
       short_ordinal: "%e %B %Y"
   travel_advice:
     alert_status:
-      avoid_all_but_essential_travel_to_parts_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all but essential travel to parts of the country.
-      avoid_all_but_essential_travel_to_whole_country_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all but essential travel to the whole country.
-      avoid_all_travel_to_parts_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all travel to parts of the country.
-      avoid_all_travel_to_whole_country_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all travel to the whole country.
+      avoid_all_but_essential_travel_to_parts_html: The <abbr title="Foreign, Commonwealth and Development Office">FCDO</abbr> advise against all but essential travel to  parts of %{country}.
```
leenagupte, line 528: Typo: There's an extra space between "to" and "parts" in

app/views/content_items/travel_advice.html.erb
```diff
@@ -16,13 +16,15 @@
...
-     <% @content_item.alert_status.each do |text| %>
-      <%= render "govuk_publishing_components/components/warning_text", {
-        text: text % {:country => @content_item.country_name
-      } 
-      } %>
+    <%= render "govuk_publishing_components/components/govspeak", {
+    } do %>
+       <% @content_item.alert_status.each do |text| %>
+        <%= render "govuk_publishing_components/components/warning_text", {
+          text: text % {:country => @content_item.country_name}
```
leenagupte, line 23: I've never seen the `%` operator used in this way. Is this supposed to pass the country name to the I18n file? I'm not sure if this is working.

config/i18n-tasks.yml
```diff
@@ -5,5 +5,5 @@ data:
 ignore_unused:
   - 'content_item.schema_name.*.{one,other,zero,few,many,two}'
   - 'corporate_information_page.*'
-  - 'travel_advice.alert_status.*'
+  # - 'travel_advice.alert_status.*'
```
leenagupte, line 8: Should this be deleted rather than commented out?

leenagupte (commented):
The PR description is wonderful! Thanks for explaining the reasoning behind all of the decisions taken.

I noticed that the presenter tests were failing, so I had a look. I don't think that the country name was being passed in to the translation file as expected. Perhaps the `%` makes it work on the page?

I thought it might be beneficial to check that the country name is being presented in the alert status correctly. However when I tried to test that the presenter was returning the country name, the test kept bombing.

It's an easy fix though. In the presenter, change the `alert_status` method to pass in the country:

```ruby
    alert_statuses.map do |alert|
      if allowed_statuses.include?(alert)
        I18n.t("travel_advice.alert_status.#{alert}_html", country: country_name).html_safe
      end
    end
```

And then update the presenter test to check for the country name:

```ruby
    test "presents alert statuses" do
      example = schema_item("full-country")
      example["details"]["alert_status"] = %w[avoid_all_but_essential_travel_to_parts avoid_all_travel_to_parts]

      presented = present_example(example)

      assert presented.alert_status.first.include?("advise against all but essential travel to parts of #{presented.country_name}")
      assert presented.alert_status.last.include?("advise against all travel to parts of #{presented.country_name}")
    end
```

I'm happy to chat in person about this.

app/views/content_items/travel_advice.html.erb
hannalaakso: I _think_ it was working. But I've changed it to the syntax we usually use 👍

app/views/content_items/travel_advice.html.erb
```diff
@@ -16,6 +16,13 @@
   <div class="govuk-grid-column-two-thirds travel-advice__header">
     <%= render 'govuk_publishing_components/components/title', @content_item.title_and_context %>
 
+    <% @content_item.alert_status.each do |text| %>
+      <%= render "govuk_publishing_components/components/warning_text", 
+        text: text, 
+        country: @content_item.country_name
```
leenagupte: As the country name is being passed into the presenter (i.e. the `@content_item` object), I don't think that it also needs to be passed into the component here. `text` will already contain the country name. 

I also don't believe that the [warning text component](https://components.publishing.service.gov.uk/component-guide/warning_text) takes a "country" parameter.

leenagupte (commented):
I've run it locally and it looks great! 🎉 

Screenshot 2023-09-20 at 17 23 56

Thanks for adding the test! 
There's one minor inline comment. Once that's fixed, this will be ready to go.

app/views/content_items/travel_advice.html.erb
hannalaakso: Thank you @leenagupte, I've now fixed this.

leenagupte (approved):
👍 🎉 

I just wanted to add that as well as the PR description, the atomic commits and informative commit messages made this change really easy to review, so thank you! 🏅

PR #2942, opened 2023-10-03
Search tracking for manual pages

andysellick:

## What
- adds GA4 tracking to search forms on manual pages
- including manuals e.g. https://www.gov.uk/guidance/understanding-your-driving-test-result/car-driving-test
- and manual section e.g. https://www.gov.uk/guidance/understanding-your-driving-test-result/car-driving-test
- and hmrc manual e.g. https://www.gov.uk/hmrc-internal-manuals/capital-gains-manual

## Why
Part of the GA4 migration.

## Visual changes
None.

Trello card: https://trello.com/c/zZGLn0tw/694-search-enhancement-add-search-form-tracking-to-manual-manualsection-and-hmrcmanualsection-pages

### DIFF
--- a/app/views/content_items/manuals/_header.html.erb
+++ b/app/views/content_items/manuals/_header.html.erb
@@ -8,7 +8,6 @@
 
 <%= content_tag(:header, aria: { "labelledby": "manual-title" }, class: classes) do %>
   <div class='govuk-grid-column-two-thirds'>
-
     <% if type %>
       <span class="manual-type"><%= type %></span>
     <% end %>
@@ -22,7 +21,6 @@
       margin_bottom: margin_bottom,
     } %>
 
-
     <%
       # The metadata component on this page receives ga4_tracking: true as it has a 'See all updates' link.
       metadata_with_ga4_tracking = { ga4_tracking: true }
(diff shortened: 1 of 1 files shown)


app/views/content_items/manuals/_header.html.erb
```diff
@@ -31,10 +29,8 @@
     <%= render 'govuk_publishing_components/components/metadata', metadata_with_ga4_tracking %>
 
     <div class="in-manual-search">
-      <form action="/search/all" >
-
+      <form action="/search/all" data-module="ga4-form-tracker" data-ga4-form='{ "event_name": "search", "type": "content", "section": "<%= content_item.title %>", "action": "search" }'>
```
AshGDS, line 32: It might be worth using `.to_json` to build this, as if the `content_item.title` contains a double quote, it looks like the tracking breaks.

E.g.

```HTML
<% test_title = "This has an \" escaped quote in it"%>
      <form action="/search/all" data-module="ga4-form-tracker" data-ga4-form='{ "event_name": "search", "type": "content", "section": "<%= test_title %>", "action": "search" }'>
        <input type='hidden' name="manual[]" value="<%= content_item.base_path %>">
```
Returns this on submit

```JS
GA4 configuration error: Expected ',' or '}' after property value in JSON at position 71 (line 1 column 72) Location {ancestorOrigins: DOMStringList, href: 'http://127.0.0.1:3090/hmrc-internal-manuals/compliance-handbook/ch160200', origin: 'http://127.0.0.1:3090', protocol: 'http:', host: '127.0.0.1:3090', …}
```

Also, the search term is coming through as `[redacted]` - is that correct?

Thanks

AshGDS (changes requested):
Looks good, though I think these need to be investigated :+1:

app/views/content_items/manuals/_header.html.erb
andysellick: Good shout, thanks for spotting that, have reworked.

Going to go back to the PAs on the search term - you're probably right, we should should capture it but process it for PII.

PR #2943, opened 2023-10-03
Update taxonomy topic email signup button

catalinailie:
When a document collection is tagged to a level one taxon, we want the email signup link to send the user to the confirmation page instead of the select subtopics page.
Example on the email signup link: [/email-signup/confirm?topic=/education/inspection-and-performance-of-schools](https://www.gov.uk/email-signup/confirm?topic=/education/inspection-and-performance-of-schools)

Trello card: https://trello.com/c/dJvwL0sU/2081-update-taxonomy-topic-email-signup-button-m


app/views/shared/_taxon_signup_link.html.erb
```diff
@@ -5,7 +5,7 @@
 >
   <%= render "govuk_publishing_components/components/signup_link", {
     link_text: "Get emails about this topic",
-    link_href: "/email-signup/?link=#{base_path}",
+    link_href: "/email-signup/confirm?topic=#{base_path}",
```
hannako, line 8: I think we should make it clearer that this is an override for a subset of document collection pages ONLY. Someone in the future might use this partial assuming it's just a normal taxon signup link - but actually it's not, it links to the confirmation page. 

I reckon the best way to do this is to replace the taxon_signup_link partial with a document collections email signup partial, and put all [this](https://github.com/alphagov/government-frontend/blob/main/app/views/content_items/document_collection.html.erb#L25-L29) logic into there. 

We should also have an integration test for document_collection_email_subscriptions, where we could test that the correct signup link is displayed for the variety of different content items that could be passed to this application. By way of documentation really, as this is a complicated thing and could easily be broken!

hannako (commented):
Thanks for picking this up Catalina!

app/views/shared/_taxon_signup_link.html.erb
catalinailie: Updated the code. Not sure what other test scenarios should I add

hannako: @JamesCGDS @AshGDS

Catalina and I were wondering if you could advise us on something here. 

Signing up for emails on https://www.gov.uk/money/dealing-with-hmrc-paying-hmrc-shares takes you to:
`https://www.gov.uk/email-signup?link=/money/dealing-with-hmrc-paying-hmrc-shares`

This PR overrides the form action of the single page notification button for some document collections, so that clicking on the button would take the user to:
`https://www.gov.uk/email-signup/confirm?topic=/money/dealing-with-hmrc-paying-hmrc-shares`

You can see that both those URL's look the same, and they both result in the same email subscription. So from out POV all good. But we wanted to check if there's any relevance of that different entry point to the tracking.

AshGDS: @hannako It looks ok to me :+1: It might be worth letting the analysts in our govuk-ga4 Slack channel know that the form action will change, as I think this change would only affect the data coming through to their dashboards rather than breaking anything us devs have written.

@andysellick Was the one who did our email subscription tracking, so he might be able to spot something code wise that I can't see!

hannako: @andysellick is this something you could take a look at, or something I should pick up with a PA? TLDR: Does it matter to analytics where the user is taken when they click on the email sign up link ie is the destination important. Or is only the click of interest.

AshGDS: @hannako - I believe he has asked the PAs here https://gds.slack.com/archives/C02HKNAEGAZ/p1698044960768499 - I think it should be OK to go ahead with, as the PAs said they haven't published the email tracking GA4 wise, so if the data coming through is incorrect it wouldn't impact their production dashboard.

hannako: Thanks for looping me in @AshGDS I've added a comment to that thread.

andysellick: Yes, what Ash said! Sorry, forgot about this thread.

hannako: Thanks all!

hannako (approved):

PR #2947, opened 2023-10-05
AB test - replace some content on UTR page

richardTowers:
https://trello.com/c/jKWA7Nu3/543-ab-test-content-for-find-your-utr-number
    
HMRC would like to see whether a link to a video performs better than
some descriptive help text. We've agreed to run an AB test to compare
these variants.

There's no elegant way to AB test bits of content in GOV.UK, so we've
had to resort to a workaround where we look for a particular replacement
string in the content (in this case `{{ab_test_find_utr_number_video_links}}`)
and replace it with a value from the translations file, depending on which variant
the user should see.

The code will only be executed on the /find-utr-number page, and only if the placeholder
string appears in the content item's body. This will allow us to deploy this code to production,
and treat the introduction of the placeholder string in the content item as the trigger to
start the test. We can try this out in the draft stack in integration and production before
running it in the live content item.

We think this is a pragmatic thing to do as a one off, as
it keeps an immportant stakeholder happy, helps us learn about video
content (which is a strategic priority), and helps us learn about the
use case for content AB tests in mainstream content.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -24,31 +24,7 @@ class ContentItemsController < ApplicationController
   def show
     load_content_item
 
-    # TEMPORARY (author: richard.towers, expected end date: November 30 2023)
-    # Content specific AB test for the Find your UTR number page
-    placeholder = "{{ab_test_find_utr_number_video_links}}"
-    if @content_item.base_path == "/find-utr-number" && @content_item.body.include?(placeholder)
-      ab_test = GovukAbTesting::AbTest.new(
-        "find_utr_number_video_links",
-        dimension: 61, # https://docs.google.com/spreadsheets/d/1h4vGXzIbhOWwUzourPLIc8WM-iU1b6WYOVDOZxmU1Uo/edit#gid=254065189&range=69:69
-        allowed_variants: %w[A B Z],
-        control_variant: "Z",
-      )
-      @requested_variant = ab_test.requested_variant(request.headers)
-      @requested_variant.configure_response(response)
-
-      replacement = case @requested_variant.variant_name
-                    when "A"
-                      I18n.t("ab_tests.find_utr_number_video_links.A")
-                    when "B"
-                      I18n.t("ab_tests.find_utr_number_video_links.B")
-                    else
-                      I18n.t("ab_tests.find_utr_number_video_links.Z")
-                    end
-      @content_item.body.sub!(placeholder, replacement)
-    end
-    # /TEMPORARY
-
+    temporary_ab_test_find_utr_page
     set_expiry
 
     if is_service_manual?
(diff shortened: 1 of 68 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -24,6 +24,35 @@ class ContentItemsController < ApplicationController
...
+        #
+        # Any change to the base content, or to the way the content is rendered
+        # could potentially break the B variant of the test, and result in both
+        # variants being the same.
+        # We're aware of this risk, and we're going to be careful in this one off
+        # situation. This is not a sustainable way of doing AB tests in the
+        # future.
+        @content_item.body.sub!(
+          /<li>\s*in\ the\s+<a\ href="[^"]*"><abbr\ title="[^"]+">HMRC<\/abbr>\s+app<\/a>\s*<\/li>/,
+          '<li>in the <a href="https://www.gov.uk/guidance/download-the-hmrc-app"><abbr title="HM Revenue and Customs">HMRC</abbr> app</a> - watch a <a href="https://www.youtube.com/watch?v=LXw9ily9rTo">video about finding your UTR number in the app</a>',
```
richardTowers, line 50: Forgot the closing `</li>` tag here (although technically they're optional in HTML now 😅)

app/controllers/content_items_controller.rb
```diff
@@ -24,6 +24,35 @@ class ContentItemsController < ApplicationController
...
+      if @requested_variant.variant? "VideoLink"
+        # NOTE: this is a fragile way of doing an AB test on content.
+        #
+        # Any change to the base content, or to the way the content is rendered
+        # could potentially break the B variant of the test, and result in both
+        # variants being the same.
+        # We're aware of this risk, and we're going to be careful in this one off
+        # situation. This is not a sustainable way of doing AB tests in the
+        # future.
+        @content_item.body.sub!(
```
hannako, line 48: As you've noted, if HMRC update the text in line 49 things break. We would still be recording variant "VideoLink" but clicks to the video would be reduced (because the video isn't there) - ie our data would be polluted. So I was wondering if there's a way to reduce that/bring it back under our control? Could we read the variant A and variant B text from a yaml file in this application? We could put the logic to fetch that data in the ContentItem::Body module maybe. And then once this AB test logic is live, we could re_publish the content item with an obvious "AB_TEST_MATCHING_STRING_DO_NOT_REMOVE" kind of thing. And then if they want to update either variant, they'd have to make a request to GDS to update it for them. It's still a hack! But maybe a more obvious and visible one?

app/controllers/content_items_controller.rb
```diff
@@ -24,6 +24,35 @@ class ContentItemsController < ApplicationController
   def show
     load_content_item
 
+    # TEMPORARY (author: richard.towers, expected end date: November 30 2023)
+    # Content specific AB test for the Find your UTR number page
+    if @content_item.base_path == "/find-utr-number"
+      ab_test = GovukAbTesting::AbTest.new(
+        "find_utr_number_video_links",
+        dimension: 300, # TODO: which dimension should we use?
+        allowed_variants: %w[HelpText VideoLink],
```
hannako, line 33: I might be getting confused here - but is there not something about having A,B and Z variants? And the control is Z. I'll see if I can dig out some docs.

hannako: I checked with @tarastockford who linked me to [this](https://docs.google.com/document/d/1MAFk0JYh9wWs-bfkmhHv8V9RPpZL3kU7uS_6GLJoed8/edit#heading=h.ig4gd5kwqtxf) and here's some [prior art](https://github.com/alphagov/collections/pull/2711/files#diff-a80f7fd7d691b7ddbcccd6c4da7abf14d1c93fa68d0a42986f4397df30957427R4)

hannako (approved):
LGTM!

PR #2948, opened 2023-10-05
Add user research banner to benefits pages

catalinailie:


app/presenters/content_item/recruitment_banner.rb
```diff
@@ -12,10 +12,34 @@ module RecruitmentBanner
       "/government/organisations/hm-revenue-customs/contact/self-assessment" => SURVEY_URL,
     }.freeze
 
+    BENEFITS_SURVERY_URL = "https://signup.take-part-in-research.service.gov.uk/home?utm_campaign=Content_History&utm_source=Hold_gov_to_account&utm_medium=gov.uk&t=GDS&id=16".freeze
+    BENEFITS_SURVERY_URL = {
```
hannako, line 16: Think you've got a typo here - maybe `BENEFITS_SURVEY_URL_MAPPINGS` ?

app/presenters/content_item/recruitment_banner.rb
```diff
@@ -12,10 +12,34 @@ module RecruitmentBanner
...
+      "/paternity-pay-leave" => BENEFITS_SURVERY_URL,
+      "/child-benefit" => BENEFITS_SURVERY_URL,
+      "/jobseekers-allowance" => BENEFITS_SURVERY_URL,
+      "/universal-credit" => BENEFITS_SURVERY_URL,
+      "/employment-support-allowance" => BENEFITS_SURVERY_URL,
+      "/benefits-calculators" => BENEFITS_SURVERY_URL,
+    }.freeze
+
     def recruitment_survey_url
       user_research_test_url
```
hannako, line 35: If you update this to be `user_research_test_url || benefits_recruitment_survey_url` you can lose https://github.com/alphagov/government-frontend/pull/2948/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR41-R49. That logic would mean that if a page was in both lists, the `user_research_test_url` would be shown. But that's the same as the `if/elsif` logic in the template anyhow. Plus it doesn't actually matter as we don't have a page that's in both lists.

app/presenters/content_item/recruitment_banner.rb
```diff
@@ -12,10 +12,34 @@ module RecruitmentBanner
...
     end
 
+    def benefits_recruitment_survey_url
+      key = content_item["base_path"]
+      BENEFITS_SURVERY_URL[key]
```
hannako, line 40: typo `BENEFITS_SURVEY_URL`

hannako (commented):
Thanks for picking this up and getting it done so speedily!!

app/presenters/content_item/recruitment_banner.rb
hannako: Ah, sorry I didn't check that. We are not supposed to allow different banner text - one of the reason being it makes it harder to implement. https://docs.google.com/document/d/1AqXsJKM9iq6ejJqteCiWHkJu7LYsXLD9icHottxMn-M/edit#heading=h.zi8eyaz9c98g

I'll make a note on the trello card for Ariana

catalinailie: Ohh I didn't know about this. Will make the changes then, thanks

hannako (approved):
LGTM

PR #2959, opened 2023-10-16
Use Time.zone.parse instead of Date.parse in ManualUpdatesPresenter

davidgisbey:
## Description

At the moment, it uses Date.parse which doesn't account for the difference in timing between the timestamp (UTC) and local time (BST).

This causes issues with some updates being a day off if they were published between 00.00 & 00.59.

Time.zone.parse parses the date correctly into the "London" timezone. We still need to convert it back into a date though as it groups on year, then by identical dates, so all updates for a day are grouped.

### Before 

image

### After 

image

## Trello card

https://trello.com/c/idVD4RQT/1794-update-history-shows-the-wrong-date

### DIFF
--- a/app/presenters/content_item/manual_updates.rb
+++ b/app/presenters/content_item/manual_updates.rb
@@ -28,7 +28,7 @@ module ContentItem
     end
 
     def updated_at(published_at)
-      Date.parse(published_at)
+      Time.zone.parse(published_at).to_date
     end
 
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
I'd have probably just used `Time.zone.parse` in the test as well to save changing the date but this will do the job. Nice one

PR #2965, opened 2023-10-20
Remove underscores from AB test name

richardTowers:
nginx no likey underscores in headers - https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#missing-disappearing-http-headers

This was resulting in the HTTP header that Fastly sets for the AB test being dropped by one of the nginxs which sit in front of rails.


anatron (approved):
LGTM 👍

PR #2967, opened 2023-10-23
Add AB test meta tag to /find-utr-number page

richardTowers:
If our AB test is active (see ContentItemsController#temporary_ab_test_find_utr_page) then @requested_variant will be set, and we should use it to add the required <meta> tag to the page.

Technically this code will work for any Answer that sets a @requested_variant field, but I think that's probably okay / what people would want in that situation anyway.


cynthia-anya (approved):
LGTM

PR #3025, opened 2023-12-22
AB tests for Self Assessment videos

davidtrussler:
[Trello](https://trello.com/c/GJ5lt8ij/589-set-up-the-next-round-of-a-b-tests)

This PR adds temporary A/B Tests to 3 pages on government-frontend: 
- https://www.gov.uk/stop-being-self-employed
- https://www.gov.uk/pay-self-assessment-tax-bill/pay-weekly-monthly
- https://www.gov.uk/log-in-file-self-assessment-tax-return

N.B. The tests for the pay-weekly-monthly one are currently failing and need more work. 

|Page|Variation A|Variation B|
|-|-|-|
|File your Self Assessment tax return online|Screenshot 2023-12-22 at 16 04 26|Screenshot 2023-12-22 at 16 04 09|
|Pay your Self Assessment tax bill|Screenshot 2023-12-22 at 18 19 11|Screenshot 2023-12-22 at 18 19 11|
|Stop being self-employed|Screenshot 2023-12-22 at 16 12 03|Screenshot 2023-12-22 at 16 11 52|


package-lock.json
```diff
@@ -0,0 +1,8358 @@
+{
```
MartinJJones, line 1: Please can we remove this file, this shouldn't be needed as we use `yarn.lock` instead.

davidtrussler: Oh yes, not sure how that got there - have removed it. 👍

hannako (approved):
Great work David and Anna! This looks good to me. I checked the pages when you published the placeholder strings on integration and used the chrome extension to toggle variants - it all seemed to be working fine.

Before merging, please could you either combine the commits into a single one, or create one commit per AB test - whichever you prefer. 

Because I've pushed a commit to fix your failing tests, you might want to get another person to give the work a once over before merging

richardTowers (commented):
Mostly looks good to me - it's tempting to suggest refactoring out the duplication (e.g. by pulling the placeholders / test names / base paths / dimensions from a config file), but that's a step towards normalising this kind of work, which I'm still reluctant to take. I prefer the more brutalist approach of just repeating the code (for now at least).

The one thing I'm not sure about is whether it's okay to reuse the same "dimension" (47) for all of the tests. I'm told this doesn't make any difference in the GA4 world anyway, but I'd appreciate a steer from a Performance Analyst on whether it would make their lives easier if we used different dimensions for different tests.

test/controllers/content_items_controller_test.rb
```diff
@@ -409,6 +453,113 @@ class ContentItemsControllerTest < ActionController::TestCase
...
+  end
+
+  test "AB test replaces content on the log-in-file-self-assessment-tax-return page with variant A" do
+    content_item = content_store_has_schema_example("answer", "answer")
+    content_item["base_path"] = "/log-in-file-self-assessment-tax-return"
+    content_item["details"]["body"] = "<li>{{ab_test_sa_video_return_1}}</li>"
+
+    stub_content_store_has_item(content_item["base_path"], content_item)
+
+    request.headers["HTTP_GOVUK_ABTEST_SAVIDEORETURN1"] = "A"
```
richardTowers, line 478: Turns out we've created a `with_variant` helper method which does this in a neater way - we could consider using it here

https://github.com/alphagov/govuk_ab_testing/blob/main/README.md#minitest

davidtrussler:
> Mostly looks good to me - it's tempting to suggest refactoring out the duplication (e.g. by pulling the placeholders / test names / base paths / dimensions from a config file), but that's a step towards normalising this kind of work, which I'm still reluctant to take. I prefer the more brutalist approach of just repeating the code (for now at least).
> 
> The one thing I'm not sure about is whether it's okay to reuse the same "dimension" (47) for all of the tests. I'm told this doesn't make any difference in the GA4 world anyway, but I'd appreciate a steer from a Performance Analyst on whether it would make their lives easier if we used different dimensions for different tests.

Hello @richardTowers We discussed the dimension earlier - see [this thread](https://gds.slack.com/archives/C068F3WJV1D/p1702570556764439) on Slack, so it seems that it is fine to use the same value for all the tests.

richardTowers (approved):
Thanks folks! Happy for this to be merged now - because of the `if ... @content_item.current_part_body.include?(placeholder)` conditional it won't do anything until the placeholder is in place, so we can deploy it everywhere and test in lower environments by updating the content.

PR #3028, opened 2023-12-28
Dry up Recruitment Banner logic

georges1996:

## How
To add a yml file and dry up the code to make it easier to update banners

Only caveat is you will need to add:

`<%= render partial: 'shared/intervention_banner' %>` to the view you want the banner to be shown on 
and add `include RecruitmentBannerHelper` to the needed controller


hannako (commented):
Thanks George - would be good to add tests that the banner is visible on all pages of interest

lib/data/recruitment_banners.yml
```diff
@@ -0,0 +1,80 @@
...
+  # - name: Banner 1
+  #   suggestion_text: "Help improve GOV.UK"
+  #   suggestion_link_text: "Sign up to take part in user research (opens in a new tab)"
+  #   survey_url: https://google.com
+  #   page_paths: 
+  #     - /
+  #     - /foreign-travel-advice
+
+banners:
+  - name: check-employment-status-for-tax Banner
```
hannako, line 12: I'm wondering about the`name` key. It's not being used for anything right? So I suppose it's to help map the banner to the banner request we receive. So in this case they would all be HMRC team banner as per the [Trello description](https://trello.com/c/eV6vhoUa/2300-govuk-user-research-banner-requests-hmrc-team-set-up-m) card.

hannako (approved):
LGTM!

PR #3031, opened 2024-01-03
Update copy for download link for travel advice map

davidgisbey:
## Description

FCDO would like the copy of this link updated to be more specific. Zendesk ticket: https://govuk.zendesk.com/agent/tickets/5495177

## Trello card

https://trello.com/c/acKHf0Md/2281-change-link-title-to-download-a-more-detailed-map

### DIFF
--- a/app/views/shared/_travel_advice_first_part.html.erb
+++ b/app/views/shared/_travel_advice_first_part.html.erb
@@ -4,7 +4,7 @@
     <img src="<%= content_item.map["url"] %>" alt="<%= content_item.map["alt_text"] %>" class="map-image">
     <% if content_item.map_download_url %>
       <figcaption>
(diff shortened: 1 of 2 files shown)


minhngocd (approved):

minhngocd:
All looks good, but the zendesk ticket link seems to be a different one to the one in the trello card

davidgisbey:
So there's 2 places that you can access that PDF. One if via a PDF of a historical edition that you can access via Travel Advice Publisher. And one is on GOV.UK which is rendered via Government Frontend. The Zendesk ticket is referencing the GOV.UK one. The example they use is this page https://www.gov.uk/foreign-travel-advice/mexico.

Maybe i overstepped doing it in both places, but it felt like it was probably useful info for people downloading a superseded edition via Travel Advice Publisher too. I don't think it can hurt anyway.

PR #3048, opened 2024-01-15
[WIP] [DO NOT MERGE] Add tracking to details elements in attachments on form pages

andysellick:

## What
Rework [form pages](https://docs.publishing.service.gov.uk/document-types/form.html) to include tracking on details components that appear inside attachment components.

Specifically, rework the list of documents on the forms pages to use the content item data in `details->attachments` to render the required content, in place of the data in `details->documents`, which is pre-rendered. 

This presents some problems, because the `attachments` partial used here is used elsewhere (consultations, call for evidence) where switching from a pre-rendered chunk of markup is not so straightforward. I've gone for a middle ground where `attachments` now accepts two parameters for the main content - either the pre-rendered markup (to keep the existing pages from breaking) or an array of data to be used in the template with components to generate the page - and updated all the places where the partial is called accordingly.

This change will impact `form` pages as well as some other page types. I'm a bit confused by the naming conventions, `form` pages are handled by `publication` code, but this will also impact `call for evidence` and `consultation` pages.

Example form pages:

- https://www.gov.uk/government/publications/make-a-lasting-power-of-attorney
- https://www.gov.uk/government/publications/application-for-a-vehicle-registration-certificate
- https://www.gov.uk/government/publications/attendance-allowance-claim-form
- https://www.gov.uk/government/publications/self-assessment-tax-return-sa100
- https://www.gov.uk/government/publications/new-year-honours-list-2024

Still to do:

- [ ] attachment thumbnails are not included in the content item outside of the pre-rendered markup, will need to find a way to pass them
- [ ] ...then republish so they come through in the content item
- [ ] fix failing tests
- [ ] apply tracking to the details components, where they appear

## Why
Good question! We need to put tracking on the details components that can (sometimes) appear beneath the list of attachments, and that's very hard to do here on pre-rendered markup - normally we'd just pass some options to the component.

Screenshot 2024-01-22 at 13 59 49

## Visual changes
None, hopefully.

Trello card: https://trello.com/c/xhASj1rp/765-add-tracking-to-details-component

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -11,7 +11,11 @@ class PublicationPresenter < ContentItemPresenter
   end
 
   def documents
-    content_item["details"]["documents"].to_a.join("")
+    docs = content_item["details"]["attachments"].select { |a| a["locale"] == locale }
+    docs.each do |doc|
+      doc.delete("alternative_format_contact_email") if doc["accessible"] == true
+    end
+    docs
   end
 
   def featured_attachments
--- a/app/views/content_items/_attachments.html.erb
+++ b/app/views/content_items/_attachments.html.erb
@@ -1,26 +1,37 @@
-<% attachments ||= [] %>
-<% attachment_details = attachments.filter_map { |id| @content_item&.attachment_details(id) } %>
-<% if legacy_pre_rendered_documents.present? || attachment_details.any? %>
+<%
+  attachments ||= nil
+  attachments_html ||= nil
+  featured_attachments ||= []
+  attachment_details = featured_attachments.filter_map { |id| @content_item&.attachment_details(id) }
+%>
+<% if attachments || attachments_html %>
   <section id="<%= title.parameterize %>">
     <%= render 'govuk_publishing_components/components/heading',
         text: title,
         mobile_top_margin: true %>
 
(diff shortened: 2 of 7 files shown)


andysellick:
@leenagupte @hannako I've updated the PR to hopefully make it a bit clearer, but please shout if you have any questions.

hannako (commented):
First main question, out of scope of your PR but just for my understanding...why have we got duplicate attachment data in the content items?!

Second question - Is it more or less confusing to an unfamiliar (eg 2nd line) dev, to have the `attachments` partial render both types of attachment data? 

I'm wondering if it might be simpler to have two partials, each handling one type. I imagine the long term plan would be to phase out the govspeak version (attachment_html)?

andysellick:
@hannako thanks for the review. Great idea on using an entirely new partial. I've raised a new PR here to take advantage of this approach and break the change into smaller commits: https://github.com/alphagov/government-frontend/pull/3058 (currently less complete than this PR).

Excellent question - yes why do we have duplicate data in the content item? I have no idea, but I think generally we should have data like this rather than pre-rendered HTML, so I'm keen to use it. It'll need some adjusting to make it fully useful (it doesn't contain the URLs for the thumbnails for each attachment) - any idea where it's coming from?

PR #3054, opened 2024-01-22
Fix video tracking

andysellick:

## What
Fix video tracking on AB test variants of:

- https://www.gov.uk/log-in-file-self-assessment-tax-return
- https://www.gov.uk/pay-self-assessment-tax-bill/pay-weekly-monthly

by doing the following:

- replaces the existing markup with the markup of the govspeak component, so that when rendered it will initialise the component JS correctly and include GA4 video tracking
- not the best way of doing our components but limited by how this page is rendered

I can't get the AB variant to work locally so I've not been able to fully test this, although I've inserted the markup into a visible page and it seems to be correct.

Related to https://github.com/alphagov/frontend/pull/3931

## Why
Video tracking is required on these pages.

## Visual changes
Shouldn't be any, although the video may appear slightly differently/better. 

Trello card: https://trello.com/c/rHlNjPxe/775-video-tracking-not-working-on-a-b-test-pages


davidtrussler (approved):
This looks like it does what's required WRT tracking (and is more elegant than the current implementation). Nice! 👍

PR #3055, opened 2024-01-22
Use edition links for Worldwide Organisations

brucebolt:
In https://github.com/alphagov/whitehall/pull/8734 and https://github.com/alphagov/publishing-api/pull/2601, we have switched worldwide organisations to use edition links to fix a bug in editionable worldwide organisations.

This applies the changes needed to the frontend to reflect the lack of multi-level link expansion in edition links.

Whilst this change adds support for edition links, it also retains compatibility with non-edition links. This will allow us to deploy the frontend changes in advance of republishing the content items, preventing any downtime of these pages. The code for backward compatibility will be removed once the content items have all been republished with edition links (all the code to be removed has been marked with a comment so it is easy to identify).

[Trello card](https://trello.com/c/yNJwHw4K)

### DIFF
--- a/app/presenters/worldwide_organisation_presenter.rb
+++ b/app/presenters/worldwide_organisation_presenter.rb
@@ -48,14 +48,18 @@ class WorldwideOrganisationPresenter < ContentItemPresenter
     return unless content_item["links"]["primary_role_person"]
 
     person = content_item.dig("links", "primary_role_person").first
-    current_roles = person.dig("links", "role_appointments").select { |role_app| role_app.dig("details", "current") }
+    current_roles = if person.dig("links", "role_appointments")
+                      person.dig("links", "role_appointments").select { |role_app| role_app.dig("details", "current") } # To be removed once switched to edition links
+                    else
+                      roles_for_person(person["content_id"])
+                    end
 
     {
       name: person["title"],
       href: person["web_url"],
       image_url: person["details"]["image"]["url"],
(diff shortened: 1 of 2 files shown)


callumknights (approved):

app/presenters/worldwide_organisation_presenter.rb
```diff
@@ -136,6 +147,18 @@ def sponsoring_organisations
 
 private
 
+  def contact_for_office(office_content_id)
+    contact_mapping = content_item.dig("details", "office_contact_associations").select { |office_contact_association|
+      office_contact_association["office_content_id"] == office_content_id
+    }.first
+
+    return unless contact_mapping
```
hannako, line 155: I notice that we don't perform this check in [`roles_for_person`](https://github.com/alphagov/government-frontend/blob/c3b86d9de7be84c86ef8ba86ea759450e604a01d/app/presenters/worldwide_organisation_presenter.rb#L157). Could you possibly link me to the schema change in your publishing API PR that determines this difference?

brucebolt: It's possible to have an office without a contact, whereas all the people will have a role (as we're only including those people in the content item).

hannako: Ah nice that makes sense, thanks

hannako (approved):
Looks good! Thanks Bruce!

PR #3058, opened 2024-01-23
Convert publications to local rendering

andysellick:

## What
Rework [form (publication) pages](https://docs.publishing.service.gov.uk/document-types/form.html) to render attachments locally in the application, rather than using the pre-rendered chunk of markup from the content item.

Specifically, rework the list of documents on the forms pages to use the content item data in `details->attachments` to render the required content, in place of the data in `details->documents`, which is pre-rendered. This has involved the creation of a new partial for publication pages, as the existing partial is used by other page types.

Example form pages:

- https://www.gov.uk/government/publications/make-a-lasting-power-of-attorney
- https://www.gov.uk/government/publications/application-for-a-vehicle-registration-certificate
- https://www.gov.uk/government/publications/attendance-allowance-claim-form
- https://www.gov.uk/government/publications/self-assessment-tax-return-sa100
- https://www.gov.uk/government/publications/new-year-honours-list-2024

## Why
Relates to previous work in https://github.com/alphagov/govuk_publishing_components/pull/3820

Reworked version of changes in https://github.com/alphagov/government-frontend/pull/3048

We need to put tracking on the details components that can (sometimes) appear beneath the list of attachments, and that's very hard to do here on pre-rendered markup - normally we'd just pass some options to the component. This work will be carried out in a follow up PR.

Screenshot 2024-01-22 at 13 59 49

## Visual changes
`details->attachments` doesn't include custom thumbnails for PDFs, as previously displayed. We looked at getting this included in the content item, but ultimately decided with consultation from leads that the default thumbnail for PDFs is acceptable. This is for several reasons:

- we believe custom thumbnails for PDFs draw users to click on PDFs over other formats, which we don't want to encourage
- technical difficulty of implementation
- as [reported to us](https://govuk.zendesk.com/agent/tickets/5580280) some PDFs produce a landscape thumbnail, which appears distorted as the attachment component expects it in portrait. While there are other solutions to this problem, removing custom thumbnails removes this problem

With the new data it looks like the attachments display the `Request an accessible format` option more than they used to. This seems to be valid (it happens if you pass an email address to the component via the `alternative_format_contact_email` option) and allowing users this option more often seems sensible, but it raises the question why some PDFs used to have this option and some didn't. 

See example below. Note how before only some of these PDFs had the accessible format option, but now all of them have.

Before | After
--- | ----
Screenshot 2024-04-11 at 10 49 16 | Screenshot 2024-04-11 at 10 49 22

Trello card: https://trello.com/c/xhASj1rp/765-add-tracking-to-details-component

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -11,7 +11,9 @@ class PublicationPresenter < ContentItemPresenter
   end
 
   def documents
-    docs = content_item["details"]["attachments"].select { |a| a["locale"] == locale } if content_item["details"]["attachments"]
+    return [] unless content_item["details"]["attachments"]
+
+    docs = content_item["details"]["attachments"].select { |a| a["locale"] == locale }
     docs.each { |t| t["type"] = "html" unless t["content_type"] }
   end
 
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -11,7 +11,8 @@ class PublicationPresenter < ContentItemPresenter
   end
 
   def documents
-    content_item["details"]["documents"].to_a.join("")
+    docs = content_item["details"]["attachments"].select { |a| a["locale"] == locale } if content_item["details"]["attachments"]
+    docs.each { |t| t["type"] = "html" unless t["content_type"] }
   end
 
   def featured_attachments
--- a/app/views/content_items/_attachments_list.html.erb
+++ b/app/views/content_items/_attachments_list.html.erb
@@ -1,27 +1,21 @@
-<% attachments ||= [] %>
-<% attachment_details = attachments.filter_map { |id| @content_item&.attachment_details(id) } %>
-<% if legacy_pre_rendered_documents.present? || attachment_details.any? %>
-  <section id="<%= title.parameterize %>">
-    <%= render 'govuk_publishing_components/components/heading',
-        text: title,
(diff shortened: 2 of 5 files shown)


app/presenters/publication_presenter.rb
```diff
@@ -11,7 +11,8 @@ def details
   end
 
   def documents
-    content_item["details"]["documents"].to_a.join("")
+    docs = content_item["details"]["attachments"].select { |a| a["locale"] == locale } if content_item["details"]["attachments"]
```
leenagupte: Oh, this is clever, so you only display attachments with the same locale as the page. Nice!

app/views/content_items/_attachments_list.html.erb
```diff
@@ -1,27 +1,21 @@
...
+    text: title,
+    mobile_top_margin: true
+  %>
+  <% if attachment_details.length > 0 %>
+    <% add_gem_component_stylesheet("details") %>
+    <% attachment_details.each do |details| %>
+      <%= render 'govuk_publishing_components/components/attachment', {
+        heading_level: 3,
+        attachment: details,
+        margin_bottom: 6
```
leenagupte, line 17: 😍

leenagupte (approved):
I got the [review app](https://government-frontend-pr-3058.herokuapp.com/government/publications/make-a-lasting-power-of-attorney) going to test it and it looks good! 💯

PR #3076, opened 2024-02-05
Add element with id "Top" to receive focus from "Top of page" links

davidtrussler:
[Trello](https://trello.com/c/gopugt7K/2365-top-of-page-links-dont-move-focus)

This deals with a widespread use of a `<a href=“#Top”>` link within page content which visually returns the user to the top of the page. However, since there is no element with that ID on the page, focus remains on the link so keyboard-only and screen-reader users are unable to navigate from that point.

This change adds a `<span>` element with that ID so that focus is added to the top of the content and these users can continue to navigate the page.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,6 +28,7 @@
     <%= yield :header %>
 
     <main role="main" id="content" class="<%= @content_item.schema_name.dasherize %>" lang="<%= I18n.locale %>">
+      <span id="Top"></span>
       <%= yield :main %>
     </main>
   </div>


Tetrino (approved):

andysellick:
Is this a problem with content containing links to `#top` in govspeak? Seems a shame to have to add an empty element to fix this.

davidtrussler:
> Is this a problem with content containing links to `#top` in govspeak? Seems a shame to have to add an empty element to fix this.

Yes it is. And yes it does. The problem is the huge number of instances of this that currently exist. We have discussed longer term solutions as well but since it's a current accessibility failure we need to get something in place to cover that.

PR #3090, opened 2024-02-13
Implement fallback for empty links in HMRC Manuals

davidtrussler:
[Trello](https://trello.com/c/PqXBvn5k/2377-implement-fallback-for-empty-links-in-hmrc-manuals-m)

This deals with an accessibility fail whereby a link to the amended document is not being displayed, see [this page](https://www.gov.uk/hmrc-internal-manuals/employment-status-manual/updates), under the entry for 5 October 2023. 

This happens because there is no value entered in the data for `title`, which is used as the text content for the link. There is a outstanding suggestion to force this to be a required value so that this does not occur but as a more immediate fix to deal with existing instances of the problem it was decided to not render the link at all in these cases. This retains the current visual appearance of the page (see below) but removes the cause of the accessibility fail (the blank link).

|Description|Screenshot|
|-|-|
|View|Screenshot 2024-02-13 at 12 58 00|
|Markup&nbsp;before|Screenshot 2024-02-21 at 14 51 15|
|Markup&nbsp;after|Screenshot 2024-02-21 at 14 49 00|

### DIFF
--- a/app/views/content_items/manuals/_updates.html.erb
+++ b/app/views/content_items/manuals/_updates.html.erb
@@ -26,9 +26,11 @@
                 <div class="govuk-!-margin-top-3">
                   <% change_notes.each do |change_note_entry| %>
                     <% change_note = change_note_entry.flatten.last %>
-                    <p class="govuk-body">
-                      <%= link_to change_note["title"], change_note["base_path"], class: "govuk-link" %>
-                    </p>
+                    <% if change_note["title"].present? %>
+                      <p class="govuk-body">
+                        <%= link_to change_note["title"], change_note["base_path"], class: "govuk-link" %>
+                      </p>
+                    <% end %>
                     <%= simple_format(change_note["change_note"], class: "govuk-body") %>
                   <% end %>
                 </div>


app/views/content_items/manuals/_updates.html.erb
```diff
@@ -27,7 +27,7 @@
                   <% change_notes.each do |change_note_entry| %>
                     <% change_note = change_note_entry.flatten.last %>
                     <p class="govuk-body">
-                      <%= link_to change_note["title"], change_note["base_path"], class: "govuk-link" %>
+                      <%= link_to change_note["title"].present? ? change_note["title"] : change_note["base_path"], change_note["base_path"], class: "govuk-link" %>
```
leenagupte, line 30: Hmmm. I don't think you want to display the base path as a link though? You'd still want human readable text in the link. Is there anything else available? Or perhaps you'd could use some generic text?

maxgds: I agree with Leena that this looks ugly and will sound horrible to screenreaders. But I appreciate we don't have much to work with. We can't use generic text because that would potentially lead to the same text multiple times on a page pointing to different urls, which would be a different accessibility fail. We could do something like extracting the last part of the url and linking "esm2055a" but that too is largely impenetrable. So another option is to just not put the link in if the title hasn't been supplied. To the sighted user the page will look the same as it does now, but we remove the bug of an empty link.

I recommend getting input from content people on which of the two options would be preferable - ugly url links, or no link at all.

davidtrussler: Yes, I agree with both comments TBH. I'm referring it back to our team to look at from a content point of view. The only counter argument might be that this is not that common an occurrence and we are considering a longer term fix as well along the lines of making `title` a required field so it's always present.

leenagupte: I had a quick look into seeing if the base path could be made into a readable string. It got ugly fast:

```ruby
irb(main):032> base_path = "/hmrc-internal-manuals/employment-status-manual/esm2055a"
=> "/hmrc-internal-manuals/employment-status-manual/esm2055a"
irb(main):033> base_path.split(/[\/]/).drop(2).join(" ").split("-").join(" ").humanize
=> "Employment status manual esm2055a"
```

davidtrussler: Thanks @leenagupte - we're putting this on hold for the minute to further discuss the best way forward but this will be useful for that discussion. 👍

davidtrussler: @leenagupte @maxgds I've updated this PR that we were discussing last week. Since then we have decided to not render a link at all where that link would have blank text content. It's still not a great solution but deemed to be the least worst: it means the affected entries are visually the same as they currently are but removing the link from the markup prevents the accessibility failure.

maxgds (approved):
Seems like a reasonable way to handle it. Cases should hopefully be rare, as as you say, for most users it's not actually a change.

hannako (commented):
Could we add an extra test [in here](https://github.com/alphagov/government-frontend/blob/main/test/integration/hmrc_manual_updates_test.rb#L51-L69)? 

You would need to make a small change to [this](https://github.com/alphagov/publishing-api/blob/022c7bf4049089c17a4905e04e9dea6a4f13bb09/content_schemas/examples/hmrc_manual/frontend/vat-government-public-bodies.json#L82) file which is the example being fetched for the test. eg:

```
{ 
 "change_note": "Updated content",
 "section_id": "VATGPB5390",
 "title": "",
 "published_at": "2015-03-05T15:16:20+00:00",
 "base_path": "/hmrc-internal-manuals/vat-government-and-public-bodies/vatgpb5390"
},
```

leenagupte (approved):
This feels better, not ideal, but better. 
Not to borrow trouble, but I'm assuming that the base_path field is required?

hannako (approved):
As discussed in Slack, let's merge this and we can add a test later.

hannako:
> This feels better, not ideal, but better. Not to borrow trouble, but I'm assuming that the base_path field is required?

@leenagupte The basepath is being added by HMRC Manuals API [here](https://github.com/alphagov/hmrc-manuals-api/blob/main/app/models/publishing_api_manual.rb#L109)

PR #3095, opened 2024-02-15
Add support for worldwide offices embedded in worldwide organisations

brucebolt:
As part of the switch to Worldwide Organisation content items containing the content for their offices, we need to switch the rendering of the Worldwide Organisation and Worldwide Office pages to use these parts.

This makes that switch in a backward compatible way, so we can continue rendering office pages under the existing content items until the office routes are updated to point to the organisation's content item.

In a later PR, we will remove `WorldwideOfficePresenter` and it's associated view, once all the routes for offices have been switched.

Depends on https://github.com/alphagov/whitehall/pull/8844.


jkempster34 (commented):
How does this load the correct template? Don't you need to modify the `content_item_template` method?

jkempster34 (approved):
All good from me! ✅

hannako (approved):
Thanks Bruce

PR #3096, opened 2024-02-15
Update NewsImage to render multiple defaults

mec:
The logic to render an image for a news item is:

- render the details > image if there is one
- render the organisation > default image if there is one
- render a placeholder

Organisation here could be one of two types:

- Organisation, presented in the primary_publishing_organisation link
- WorldwideOrganisation, presented in the worldwide_organisations link

The placeholder is also dependent on which type of document and so which
type of organisation the content links to.

Here we allow the module to render the default image from the worldwide
organisation link, when the document is the appropriate type and one is
available.

Right now this change will not be used as Whitehall currently sets
details > image to the default image for worldwide organisation news
items, but this is set to change as we want this logic out of the
publishing applications and for all related concepts like 'the default
news image' to work the same way.

With this change, once we do turn the replacement in Whitehall off,
everything will continnue to work without change.

### DIFF
--- a/app/presenters/content_item/news_image.rb
+++ b/app/presenters/content_item/news_image.rb
@@ -7,16 +7,14 @@ module ContentItem
   private
 
     def default_news_image
-      if content_item["document_type"].eql? "world_news_story"
-        return first_worldwide_organisation_default_news_image
-      end
+      return first_worldwide_organisation_default_news_image if world_news_story?
 
       first_primary_publishing_organisation_default_news_image
     end
 
     def placeholder_image
       # this image has been uploaded to asset-manager
-      if content_item["document_type"] == "world_news_story"
+      if world_news_story?
         { "url" => "https://assets.publishing.service.gov.uk/media/5e985599d3bf7f3fc943bbd8/UK_government_logo.jpg" }
       else
         { "url" => "https://assets.publishing.service.gov.uk/media/5e59279b86650c53b2cefbfe/placeholder.jpg" }
@@ -31,6 +29,10 @@ module ContentItem
     def first_primary_publishing_organisation_default_news_image
       organisation = content_item.dig("links", "primary_publishing_organisation")
       organisation[0].dig("details", "default_news_image") if organisation.present?
+    end
+
+    def world_news_story?
+      content_item["document_type"] == "world_news_story"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/presenters/content_item/news_image.rb
```diff
@@ -7,17 +7,32 @@ def image
...
+      if world_news_story?
         { "url" => "https://assets.publishing.service.gov.uk/media/5e985599d3bf7f3fc943bbd8/UK_government_logo.jpg" }
       else
         { "url" => "https://assets.publishing.service.gov.uk/media/5e59279b86650c53b2cefbfe/placeholder.jpg" }
       end
     end
+
+    def first_worldwide_organisation_default_news_image
+      worldwide_organisation = content_item.dig("links", "worldwide_organisations")
+      worldwide_organisation[0].dig("details", "default_news_image") if worldwide_organisation.present?
```
brucebolt, line 26: Minor nitpick where I think `first` would be nicer than accessing with a numeric position.  The `first` method still returns `nil` if there's an empty array.
```suggestion
      worldwide_organisation = content_item.dig("links", "worldwide_organisations").first
      worldwide_organisation.dig("details", "default_news_image") if worldwide_organisation.present?
```

app/presenters/content_item/news_image.rb
```diff
@@ -7,17 +7,32 @@ def image
...
+    end
+
+    def first_primary_publishing_organisation_default_news_image
+      organisation = content_item.dig("links", "primary_publishing_organisation")
+      organisation[0].dig("details", "default_news_image") if organisation.present?
```
brucebolt, line 31: Same here as my previous comment.

brucebolt (approved):
Looks good, just a minor nitpick about the syntax to get the worldwide organisation.

jkempster34:
We considered including default news image as a link instead of relying on ActiveRecord callbacks. However:
> 1. That leaves us with inconsistency between the records in Whitehall, and the content item (unless we add additional fields to the content item)
> 2. There is already a [callback pattern](https://github.com/alphagov/whitehall/blob/cee99a9c6a7d59afc72cf6d262af5cd40c2006f2/app/models/featured_image_data.rb#L34) in place to deal with the delay in uploading assets. 

See https://github.com/alphagov/whitehall/pull/8877

PR #3122, opened 2024-03-12
Ensure dashes not read in screenreaders

beccapearce:
Nearly every screen reader (except NVDA of the major ones) reads out the “en dash“ (VoiceOver on macOS says “space“) before each item/link in the “Page contents“ list. Screen reader users found that frustrating to read through.

To solve this we have used a [version of the contents list in the publishing components ](https://github.com/alphagov/govuk_publishing_components/pull/3907/files )which fixes this issue. This has also meant that we have removed some of the custom elements that were applying to the contents list in favour of the standard govuk-publishing-components. This means that there is a change in the appearance of the contents due to this.

Before:

https://github.com/alphagov/government-frontend/assets/17481621/192da476-3fae-4d87-8d6b-daefff553f7d

After:

https://github.com/alphagov/government-frontend/assets/17481621/9079c13b-0523-40e9-ab27-747d54594296

Trello card: https://trello.com/c/prtvfJV5


app/presenters/service_manual_guide_presenter.rb
```diff
@@ -8,7 +8,7 @@ def body
 
   def header_links
     header_links = details.fetch("header_links", {})
-    Array(header_links).map { |h| ActiveSupport::HashWithIndifferentAccess.new(h) }
+    Array(header_links).map { |h| ActiveSupport::HashWithIndifferentAccess.new(h.transform_keys { |k| k == "title" ? "text" : k }) }
```
beccapearce, line 11: I don't think this is the best way of doing this, very open to suggestions that make it clearer, or I can add some comments

davidtrussler:
This looks great to me. 👍 Nice to see a bunch of stuff removed as part of the work. 
The only thing that's slightly concerning is if the tracking is correct. I see that it is now using the tracking built into the component but it seems to give a very different result to the existing links. Saying that I don't really know what I'm looking at with tracking so it might be fine. 🤷

beccapearce:
Yes! Martin and I took a look at the tracking and we _think_ it looked ok when we tested it on integration

davidtrussler:
> Yes! Martin and I took a look at the tracking and we _think_ it looked ok when we tested it on integration

OK, I suspect that's just a function of using the component then. 

The other thing (sorry!) is that it does create a fairly different experience for the user: where it used to retain the list of links in view and scroll the content to it, we now have standard skip links. Should we (or maybe we have) consult with someone on the design side about this? Maybe we'd need to introduce some kind of "back to top" mechanism or some-such. I tend to think that using the component makes for a more consistent experience across all parts of the site but it's quite a big change. (I'm sure we can live with the change from en dash to em dash)

beccapearce:
AH no I don't think we have checked with design, I can do that!

MartinJJones:
I had a further look at the GA events firing after switching to the `contents-list` component, it looks like we need to test the changes on Integration as the GA4 event do not appear to fire on Heroku.

When comparing the difference for the `select_content` event, the only event data that changes is for the section as shown below.

```diff
- Event Data (section): Page contents:
+ Event Data (section): Contents
```

The only other change that I could see relating to Google Analytics events, is that using the `contents-list` component will also fire an "Event" for Universal Analytics as well.

@andysellick @AshGDS sorry to tag you in directly, just want to double-check my understanding here and if there is anything I may be overlooking relating to Google Analytics with this change? thanks

MartinJJones:
Thanks for the further info @andysellick 👍

davidtrussler (approved):
Nice! Sorts out previous concerns with user experience elegantly.
All looks good to me 👍

beccapearce:
While completing this work, we noticed that the functionality in the [sticky module](https://github.com/alphagov/government-frontend/blob/main/app/assets/stylesheets/modules/_sticky.scss ) is similar to that used by [html publications' content lists](https://github.com/alphagov/government-frontend/blob/main/app/assets/stylesheets/views/_html-publication.scss#L19-L26.) Ideally we could combine the two into one place to tidy up the code and reduce duplication. 

We should complete this work as a separate PR since it may cause some changes that should be discussed separately.

PR #3148, opened 2024-04-08
2412 heading sizes in government frontend

davidtrussler:
[Trello](https://trello.com/c/T1JGEeyz/2412-investigate-heading-sizes-in-government-frontend-m-l)

By way of an investigation into heading sizes, and a suggested way of dealing with inconsistencies and accessibility issues, I have made some small suggested changes to heading levels and sizes to make the three manuals headings consistent, based on the design [here](https://www.figma.com/file/KITAANxWkx64lUtH9U3Nlr/Manuals-Headings?type=design&node-id=0-1&mode=design&t=jObeFE1vLMTMICvK-0). 

Initially this brought two of these ("[Content design: planning, writing and managing content](https://www.gov.uk/guidance/content-design/updates)" and "[Employment Status Manual](https://www.gov.uk/hmrc-internal-manuals/employment-status-manual/updates)") into line with the recently updated "[Complete the school census](https://www.gov.uk/guidance/complete-the-school-census/census-dates)" page. 

This became problematic though because the `<h3>` elements on the two pages changed here are part of the accordion component and cannot (or maybe should not) be changed in terms of the font size. My feeling is that it should not be a problem to have different sized `<h3>`s across these pages since some of those are functioning as part of the accordion.

|Before|After|Changes|
|-|-|-|
|Screenshot 2024-04-08 at 11 36 21|Screenshot 2024-04-12 at 14 53 46|Top heading ("Content design: planning, writing and managing content") retained as `<h2>` and font size reduced to 27px<br><br>`<h1>` ("Updates: Content design: planning, writing and managing content") updated (slightly) to align content with page title<br><br>`<h2>` ("2023") reduced font size from 36px to 27px|
|Screenshot 2024-04-08 at 11 43 57|Screenshot 2024-04-12 at 15 05 28|Top heading ("Employment Status Manual") retained as `<h2>` and font size reduced to 27px<br><br>`<h1>` ("Updates: Employment Status Manual") updated to align content with page title<br><br>`<h2>` ("2023") reduced font size from 36px to 27px|
|Screenshot 2024-04-08 at 11 46 23|Screenshot 2024-04-12 at 15 09 19|Top heading ("Complete the school census") changed to `<h2>` and font size reduced to 27px<br><br>`<h1>` ("Census dates") updated to align content with page title<br><br>`<h2>` ("2023 to 2024 census dates") changed from 24px to 27px|

### DIFF
--- a/app/assets/stylesheets/views/_manual.scss
+++ b/app/assets/stylesheets/views/_manual.scss
@@ -142,3 +142,7 @@
     padding-right: govuk-spacing(3);
   }
 }
+
+.test-accordion-heading-s .govuk-accordion__section-heading-text {
+  font-size: 16px;
+}
--- a/app/views/content_items/hmrc_manual_updates.html.erb
+++ b/app/views/content_items/hmrc_manual_updates.html.erb
@@ -9,4 +9,4 @@
   } %>
 <% end %>
 
-<%= render "content_items/manuals/manual_updates_layout" %>
+<%= render "content_items/manuals/manual_updates_layout", type: I18n.t("manuals.hmrc_manual_type") %>
--- a/app/views/content_items/hmrc_manual_updates.html.erb
+++ b/app/views/content_items/hmrc_manual_updates.html.erb
@@ -2,7 +2,7 @@
 <% content_for :header do %>
   <%= render "content_items/manuals/header", {
     content_item: @content_item,
-    heading_level: 2,
+    heading_level: 1,
     margin_bottom: 6,
     green_background: true,
     type: I18n.t("manuals.hmrc_manual_type"),
--- a/app/views/content_items/hmrc_manual_updates.html.erb
+++ b/app/views/content_items/hmrc_manual_updates.html.erb
@@ -2,7 +2,7 @@
 <% content_for :header do %>
   <%= render "content_items/manuals/header", {
     content_item: @content_item,
-    heading_level: 1,
+    heading_level: 2,
     margin_bottom: 6,
     green_background: true,
     type: I18n.t("manuals.hmrc_manual_type"),
--- a/app/views/content_items/manual_section.html.erb
+++ b/app/views/content_items/manual_section.html.erb
@@ -20,7 +20,6 @@
           <div class="govuk-!-margin-bottom-3">
             <%= render "govuk_publishing_components/components/heading", {
               text: item[:heading][:text],
-              font_size: "m",
               margin_bottom: 1,
               id: item[:heading][:id],
             } %>
--- a/app/views/content_items/manual_section.html.erb
+++ b/app/views/content_items/manual_section.html.erb
@@ -1,7 +1,9 @@
 <% add_view_stylesheet("manual") %>
 <% content_for :header do %>
   <%= render "content_items/manuals/header", {
-    content_item: @content_item, heading_level: 1, margin_bottom: 6,
+    content_item: @content_item,
+    heading_level: 2,
+    margin_bottom: 6
   } %>
   <%= render partial: "content_items/manuals/breadcrumbs" %>
 <% end %>
--- a/app/views/content_items/manual_updates.html.erb
+++ b/app/views/content_items/manual_updates.html.erb
@@ -1,9 +1,7 @@
 <% add_view_stylesheet("manual") %>
 <% content_for :header do %>
   <%= render "content_items/manuals/header", {
-    content_item: @content_item,
-    heading_level: 2,
-    margin_bottom: 6
+    content_item: @content_item, heading_level: 2, margin_bottom: 6
   } %>
 <% end %>
 
--- a/app/views/content_items/manual_updates.html.erb
+++ b/app/views/content_items/manual_updates.html.erb
@@ -1,7 +1,7 @@
 <% add_view_stylesheet("manual") %>
 <% content_for :header do %>
   <%= render "content_items/manuals/header", {
-    content_item: @content_item, heading_level: 2, margin_bottom: 6,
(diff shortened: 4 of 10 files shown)


MartinJJones (commented):
Nice work on this, the changes look good to me 👍 

I agree with your point on the differences between the h3 headings sizes used in the accordion and in govspeak, there is also an issue in the gem relating to the govspeak heading sizes - https://github.com/alphagov/govuk_publishing_components/issues/3576

maxgds:
@davidtrussler @MartinJJones 
Sorry to undermine this work but I’m afraid it isn’t correct to move the H1 to the piece of content that you have. The H1 on the page should be the most specific thing about it, and ideally it should be unique across the site (though this is often impossible to achieve). The H1 and the page title (ie what shows in the tab, and is marked up with the <title> tag) should align. By making the H1 “Employment status manual” you are no longer matching the title, and if you were to match the title then the page would fail WCAG because both these pages would appear to be the same one:
https://www.gov.uk/hmrc-internal-manuals/employment-status-manual
https://www.gov.uk/hmrc-internal-manuals/employment-status-manual/updates

Effectively the manual is a book called “Employment Status Manual” but the page is the chapter about updates and the chapter title is more important. The H1 is one of the ways to orient screenreader users and things like search engines and disambiguate the pages. We generally try to start headings with the h1 first, but it’s not a WCAG fail if we don’t do that - see example 2 here:
https://www.w3.org/WAI/tutorials/page-structure/headings/

I don't know if you might need to revisit the other piece of work referenced in light of this info.

There's no problem with applying relevant heading sizes, eg govuk-heading-m to the headings, though.

I think that although the accordion doesn't currently accept a font size change, that's only because we haven't needed to before. We can extend that component to add the necessary class to the span, leading to something like 
<span class="govuk-accordion__section-heading-text govuk-heading-s" id="default-id-3060d499-heading-1"> - I think this lies within the remit of what the design system advocates when it says "Structure section headings with the rest of the page", I think it has to both semantically and visually align with its surrounding content. However I'm not sure if this is necessary if you aren't moving the H1. Happy to chat through this and figure out next steps with you.

davidtrussler:
> @davidtrussler @MartinJJones Sorry to undermine this work but I’m afraid it isn’t correct to move the H1 to the piece of content that you have. The H1 on the page should be the most specific thing about it, and ideally it should be unique across the site (though this is often impossible to achieve). The H1 and the page title (ie what shows in the tab, and is marked up with the <title> tag) should align. By making the H1 “Employment status manual” you are no longer matching the title, and if you were to match the title then the page would fail WCAG because both these pages would appear to be the same one: https://www.gov.uk/hmrc-internal-manuals/employment-status-manual https://www.gov.uk/hmrc-internal-manuals/employment-status-manual/updates
> 
> Effectively the manual is a book called “Employment Status Manual” but the page is the chapter about updates and the chapter title is more important. The H1 is one of the ways to orient screenreader users and things like search engines and disambiguate the pages. We generally try to start headings with the h1 first, but it’s not a WCAG fail if we don’t do that - see example 2 here: https://www.w3.org/WAI/tutorials/page-structure/headings/
> 
> I don't know if you might need to revisit the other piece of work referenced in light of this info.
> 
> There's no problem with applying relevant heading sizes, eg govuk-heading-m to the headings, though.
> 
> I think that although the accordion doesn't currently accept a font size change, that's only because we haven't needed to before. We can extend that component to add the necessary class to the span, leading to something like - I think this lies within the remit of what the design system advocates when it says "Structure section headings with the rest of the page", I think it has to both semantically and visually align with its surrounding content. However I'm not sure if this is necessary if you aren't moving the H1. Happy to chat through this and figure out next steps with you.

@maxgds Thanks for that - makes sense. I've updated this a bit (reflected in the screenshots and stuff in the "changes" column) to hopefully address it in line with what you are saying, if I have understood correctly. In essence this is to change the top heading to an `<h2>` and update the content of the `<h1>` to more closely align it content-wise with the page title. The latter might be something for content to check but it should at least be unique in each case I think.

maxgds:
> @maxgds Thanks for that - makes sense. I've updated this a bit (reflected in the screenshots and stuff in the "changes" column) to hopefully address it in line with what you are saying, if I have understood correctly. In essence this is to change the top heading to an `<h2>` and update the content of the `<h1>` to more closely align it content-wise with the page title. The latter might be something for content to check but it should at least be unique in each case I think.

Thanks for the reworking. I think it is good that we've brought consistency to the way the H1s are shown, but it might need to be checked with content people to ensure that they're happy with the change to the Census dates format. Although we use a "-" in the `<title>` tag I think the ":" is probably more appropriate for the `<H1>` - again I'd suggest checking with a content designer. Ask if we might be better to put at ":" into the `<title>` or if the mismatch that exists is ok.

The font size changes outside of the "banner" (for want of a better term) are what was most important here. I can see you've adjusted the banner heading font to match. In discussion with Anika she felt that what was inside the banner was different enough from the body copy that it didn't need to change size, and now you've changed it I think it does look a bit offbalanced. I'd check that change with a designer and see what they prefer. What's important is that the structure of the page is comprehensible both for screenreader users and sighted users - you can instinctively tell that the banner content is different from the rest of the page, almost a chunk of metadata, and the structure from there onwards flows with the sizes making sense, while the screenreader user gets a similar understanding from the h2 preceding the h1, and the heading levels being appropriately labelled.

hannako:
I'm going to close this PR for now, and we will pick it up as a fresh ticket in the future if the issue is raised.

KludgeKML:
Closing because Government-Frontend is being prepared for retirement.

PR #3157, opened 2024-04-17
Correctly display 'request accessible format' option

andysellick:

## What
- attachments on publications should only show this option when 'accessible' is false and email address is supplied
- including tests accordingly

Follows on from https://github.com/alphagov/government-frontend/pull/3058 where this problem was introduced, see that PR for example pages.

## Why
Issue raised in zendesk: https://govuk.zendesk.com/agent/tickets/5807086

## Visual changes
Fewer attachments should now display this option.


KludgeKML (approved):
LGTM (But it'll want you to lint it)

leenagupte (commented):
Andy, the code looks good, but as Keith say's you need to run linter and fix up.

`bundle exec rubocop -A` should hopefully fix it.

PR #3159, opened 2024-04-18
Fix publication attachments

andysellick:

## What
- simplify and fix the code to display attachments on publication/form pages
- move most of the code into the presenter, to reduce ambiguity
- pass only one thing to the attachments_list component
- update tests

Follows on from the work done in https://github.com/alphagov/government-frontend/pull/3058

## Why
Recent changes to rendering of publication attachments were not fully correct, as reported in zendesk ticket https://govuk.zendesk.com/agent/tickets/5807151 and elsewhere.

Specifically:

- attachments on https://www.gov.uk/government/publications/further-education-workforce-data-collection should not be marked as inaccessible, so the option to request an accessible format should not be shown
- attachments on https://www.gov.uk/government/publications/dvsa-vision-to-2030 were not marked as HTML where they should be

## Visual changes
See 'why'.

### DIFF
--- a/app/presenters/publication_presenter.rb
+++ b/app/presenters/publication_presenter.rb
@@ -10,10 +10,14 @@ class PublicationPresenter < ContentItemPresenter
     content_item["details"]["body"]
   end
 
+  def attachments_for_components
+    documents.select { |doc| featured_attachments.include? doc["id"] }
+  end
+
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM

PR #3163, opened 2024-04-19
Use govuk publishing components content list to prevent hyphens being read out by screenreaders

beccapearce:
## What
Use the `contents-list` from `govuk-publishing-component` to render the list of corporate information pages for worldwide organisations.

## Why
The contents list component has a fix for an issue where nearly every screen reader  reads out the hyphen as either “dash“ or “hyphen“ before each item/link in the Corporate information list.  Screen reader users found that frustrating to read through, so this should improve the experience.

## Screenshots
| Before | After |
|--------|--------|
| image | image | 

Trello card: https://trello.com/c/pVeGF0WE

### DIFF
--- a/app/presenters/worldwide_organisation_presenter.rb
+++ b/app/presenters/worldwide_organisation_presenter.rb
@@ -112,7 +112,7 @@ class WorldwideOrganisationPresenter < ContentItemPresenter
     ordered_cips.map do |cip|
       {
         text: cip["title"],
-        url: cips.find { |cp| cp["content_id"] == cip["content_id"] }["web_url"],
+        href: cips.find { |cp| cp["content_id"] == cip["content_id"] }["web_url"],
       }
     end
   end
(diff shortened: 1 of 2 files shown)


test/integration/worldwide_organisation_test.rb
```diff
@@ -52,6 +52,7 @@ class WorldwideOrganisationTest < ActionDispatch::IntegrationTest
 
   test "renders the navigational corporate information pages" do
     setup_and_visit_content_item("worldwide_organisation")
+    puts page.body
```
andysellick, line 55: Debug?

andysellick: Oh sorry, this is still in draft, apologies. Change looks good though 👍

beccapearce: Ready now! Sorry I was a bit too keen with making it a PR 😅

MartinJJones (commented):
Nice work on this Becca!

I've added a comment for 1 small thing to look at.

I think it could also be worth running past a Content Designer for a quick look at the changes, as we now include "Contents" above the links. If we want to avoid this, I think we could always switch out the `contents_list` component with the `list` component instead - https://components.publishing.service.gov.uk/component-guide/list

beccapearce:
Closing this PR in favour of another one that uses the list component to solve the same issue. 
By using the list instead of the content-list we avoid adding a confusing `contents` heading to the list

PR #3164, opened 2024-04-19
Prevent hyphens being read out by screen readers

beccapearce:
## What
Use the `list` from `govuk-publishing-component` to render the list of corporate information pages for worldwide organisations.

## Why
The  list component means we can remove the hyphens that were rendered previously. These were read out as either “dash“ or “hyphen“ before each item/link in the Corporate information list.  Screen reader users found that frustrating to read through, so this should improve the experience.

## Screenshots
| Before | After |
|--------|--------|
| image | image | 

Trello card: https://trello.com/c/pVeGF0WE

### DIFF
--- a/app/presenters/worldwide_organisation_presenter.rb
+++ b/app/presenters/worldwide_organisation_presenter.rb
@@ -99,21 +99,19 @@ class WorldwideOrganisationPresenter < ContentItemPresenter
   end
 
   def show_corporate_info_section?
-    corporate_information_pages.any? || secondary_corporate_information.present?
+    corporate_information_pages.present? || secondary_corporate_information.present?
   end
 
   def corporate_information_pages
     cips = content_item.dig("links", "corporate_information_pages")
-    return [] if cips.blank?
+    return if cips.blank?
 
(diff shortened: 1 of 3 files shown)


app/presenters/worldwide_organisation_presenter.rb
```diff
@@ -109,12 +109,14 @@ def corporate_information_pages
...
-        url: cips.find { |cp| cp["content_id"] == cip["content_id"] }["web_url"],
-      }
+    links = []
+    ordered_cips.each do |cip|
+      text = cip["title"]
+      link = cips.find { |cp| cp["content_id"] == cip["content_id"] }["web_url"]
+      hyperlinked_text = "<a href =#{link} class=govuk-link>#{text}</a>"
+      links << hyperlinked_text
     end
+    links
```
hannako, line 119: Should we follow the pattern for this set higher up in the file, lines 27-35?

beccapearce: YES!

hannako (approved):
LGTM!

PR #3169, opened 2024-04-24
Address deprecation warning from minitest

beccapearce:
Addressing the deprecation warning:

```
DEPRECATED: Use assert_nil if expecting nil. This will fail in Minitest 6.
```

Using assert_equal nil will be deprecated in minitest 6. We should address this deprecation so upgrading will be easier when the time comes, as well as to keep our test outputs nice and clear.


CodeSonia (approved):
LGTM 🙌

PR #3181, opened 2024-05-10
Use components for consultation attachments

andysellick:

## What
Output attachments in consultation pages using the data from the content item, rather than the pre-rendered markup from the content item, and apply GA4 tracking to those attachments that contain Details elements.

This follows on from the work to convert publications in the same manner: https://github.com/alphagov/government-frontend/pull/3058

Attachments on consultation pages can occur in three places, so this is a bit more complex than publications, although some of the code and tests have been copied from that previous work.

Example pages to test with:

- [concluded consultation](https://www.gov.uk/government/consultations/setting-the-grade-standards-of-new-gcses-in-england-2017-2018) that displays documents in all three places
- [another concluded consultation](https://www.gov.uk/government/consultations/improving-the-ssros-response-to-contract-specific-queries-on-the-regulatory-framework-for-single-source-defence-contracts) with documents in two places
- [consultation in progress](https://www.gov.uk/government/consultations/domestic-maritime-decarbonisation-the-course-to-net-zero-emissions) with documents in two places

Note: an interesting side point on margin - there's some code for logic around a margin to appear beneath boxes that appear at the top of in-progress consultations ('We are analysing your feedback') so that spacing is only applied if there's stuff beneath it. Pretty sure I've preserved the logic but worth checking.

## Why
This work is for two reasons:

- to convert the rendering of attachments to using components so that tracking can be added to Details elements within attachments (for the 'Request an accessible format' element)
- to generally move away from the use of pre-rendered markup passed through content items

## Visual changes
Mostly should be no visual changes, however this change does remove custom thumbnails for PDFs (as previously removed for publications). This has been agreed and widely discussed around the previous work.

### DIFF
--- a/app/presenters/consultation_presenter.rb
+++ b/app/presenters/consultation_presenter.rb
@@ -52,20 +52,50 @@ class ConsultationPresenter < ContentItemPresenter
     content_item["details"]["final_outcome_detail"]
   end
 
-  def final_outcome_documents
-    content_item["details"]["final_outcome_documents"]&.join("")
+  # Download the full outcome, top of page
+  def final_outcome_attachments_for_components
+    documents.select { |doc| final_outcome_attachments.include? doc["id"] }
+  end
+
+  # Feedback received, middle of page
+  def public_feedback_attachments_for_components
+    documents.select { |doc| public_feedback_attachments.include? doc["id"] }
+  end
+
+  # Documents, bottom of page
+  def documents_attachments_for_components
+    documents.select { |doc| featured_attachments.include? doc["id"] }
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM (I wonder if we should extract the presenter stuff into a mixin if it's going to be largely the same in all the presenters, but perhaps we wait until they're all done and then refactor)

andysellick:
Thanks @KludgeKML yes definitely something to consider once all the formats have been converted. There's certainly a lot of duplication here now, particularly in the tests.

PR #3190, opened 2024-05-17
Remove publisher metadata component

andysellick:

## What / why
Remove the 'published metadata' component: https://govuk-government-frontend.herokuapp.com/component-guide/publisher_metadata

- appears to not be in use
- functionality can be reproduced using the metadata component from `govuk_publishing_components` if we need it

## Visual changes
None. This is what the component looks like (for some reason styles don't work properly for it in the government-frontend component guide).

Screenshot 2024-05-17 at 14 20 15


kevindew:
Nice work @andysellick I think you also have phrases in the locale files to remove for this (I searched for publisher_metadata)

There is also this line to adjust: https://github.com/alphagov/government-frontend/blob/882723a158ab355d9f65fdf21783e33f14d1a384/app/views/components/docs/important_metadata.yml#L4

I think other search matches to publisher_metadata were unrelated.

andysellick:
Thanks @kevindew have updated 👍

kevindew (approved):
Looks good to me 👍

PR #3192, opened 2024-05-17
Fix the '+(xx) others' link losing focus after activation

matthillco:
## What/Why

From [Trello](https://trello.com/c/jrwE1oz3/2588-using-link-loses-focus-with-jaws-s-m):
_When using the ‘+ 12 others’ link with JAWS, the focus is lost. Using the tab key gets you to the first inserted link. But when navigating ‘in context’, using arrow keys to read through the page, the focus for JAWS users gets moved back to the start of the page which causes them to lose their place._ 

## More info
This was reported as an issue in JAWS. After testing, I also observed this as a problem in other screen-readers. However, this issue was potentially a problem for _all_ users as it did not focus a relevant element after activation. It would be confusing for non-screen-reader users who use the keyboard to navigate.

The behaviour has now changed:
* Anchor element is replaced with a button for improved accessibility
* After activation, focus the first link in the now-visible dropdown

Styling modified:
* Button inherits parent styles to appear like a link
* Text size increased to 16px to match surrounding elements

I've tested this extensively in Assistiv Labs using JAWS, NVDA and Talkback and the behaviour now seems consistent:

* Use tab or up/down arrows to move through elements in screen reader view
* When the _+XX others_ text is focused, user activates it with enter or space
* Other languages are displayed and the _+XX others_ text is removed
* The first language in the revealed list now has focus

### DIFF
--- a/app/assets/javascripts/modules/hide-other-links.js
+++ b/app/assets/javascripts/modules/hide-other-links.js
@@ -9,7 +9,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.hiddenElementContainer = this.createHiddenElementContainer()
     this.shownElements = []
     this.hiddenElements = []
-    this.showLink = document.createElement('a')
+    this.showLink = document.createElement('button')
   }
 
   HideOtherLinks.prototype.init = function () {
@@ -51,7 +51,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.showLink.classList.add('show-other-content', 'govuk-link')
     this.showLink.innerHTML = linkText
-    this.showLink.href = '#'
+    this.showLink.setAttribute('aria-expanded', 'false')
+    this.showLink.setAttribute('aria-controls', 'other-content')
 
     this.showLink.addEventListener('click', this.showHiddenLinks.bind(this))
 
@@ -71,6 +72,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   HideOtherLinks.prototype.createHiddenElementContainer = function () {
     var showHide = document.createElement('span')
     showHide.classList.add('other-content')
+    showHide.id = 'other-content'
 
     return showHide
   }
@@ -79,8 +81,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     event.preventDefault()
 
     this.hiddenElementContainer.style.display = ''
+    this.hiddenElementContainer.querySelectorAll('a')[0].focus()
     this.showLink.remove()
-    this.hiddenElementContainer.focus()
   }
 
   Modules.HideOtherLinks = HideOtherLinks
--- a/app/assets/stylesheets/views/_worldwide-organisation.scss
+++ b/app/assets/stylesheets/views/_worldwide-organisation.scss
@@ -6,9 +6,14 @@
 .js-enabled {
   .worldwide-organisation-header {
     .show-other-content {
-      @include govuk-font(14);
+      all: inherit;
+      display: inline;
       margin-left: .3em;
       white-space: nowrap;
+      color: $govuk-link-colour;
+      cursor: pointer;
+      @include govuk-link-common;
+      @include govuk-font(16);
     }
   }
 }


app/assets/stylesheets/views/_worldwide-organisation.scss
```diff
@@ -6,9 +6,13 @@
 .js-enabled {
   .worldwide-organisation-header {
     .show-other-content {
-      @include govuk-font(14);
+      all: inherit;
+      display: inline;
       margin-left: .3em;
```
andysellick, line 11: Suggestion: adding `cursor: pointer` would make the mouse hover behaviour consistent with the other links here.

AshGDS (commented):
Looks good, but I have a couple questions

- The button disappears when the content is expanded - is this correct? Could this confuse users into not knowing if the button has been expanded, as it no longer exists after interacting with it? 
- Should the button have an `aria-controls` on it as well, like our accordion component does?

app/assets/javascripts/modules/hide-other-links.js
```diff
@@ -79,8 +79,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     event.preventDefault()
 
     this.hiddenElementContainer.style.display = ''
+    this.hiddenElementContainer.querySelectorAll('a')[0].focus()
```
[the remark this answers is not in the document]
matthillco: Do you mean using index [0] to target the first anchor is brittle? I'm unclear of the benefit of adding a `js-` class to the button to fix this. Perhaps a better solution is to target the first anchor by class name, as they all have a `govuk-link` class. The reason I was avoiding that in case of the possibility that these anchors might not have valid classes, but checking their origin the class is assigned in the presenter so that seems safe to use.

matthillco:
@AshGDS, thanks for the comments

> * The button disappears when the content is expanded - is this correct? Could this confuse users into not knowing if the button has been expanded, as it no longer exists after interacting with it?

I agree, it's a strange pattern. However, the requirement for this work is just to fix a problem with lost focus once the button is removed. As there is a task on the Trello card to review this pattern for a refactor, I think it's best not to confuse the issue as part of this work.

> * Should the button have an `aria-controls` on it as well, like our accordion component does?

I'm not sure. Does that create a meaningfully different experience for a screen-reader user than what they get without it? My accessibility knowledge on that is not extensive. Happy to be guided on this if you can provide more info, thanks.

app/assets/javascripts/modules/hide-other-links.js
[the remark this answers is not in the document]
andysellick: Oh I got confused. Ignore my original point - the solution I suggested wasn't valid.

app/assets/stylesheets/views/_worldwide-organisation.scss
matthillco: Good spot, thanks, I've now added that.

matthillco:
@AshGDS I've added `aria-controls` as you suggested, thanks.

app/assets/javascripts/modules/hide-other-links.js
```diff
@@ -51,7 +51,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.showLink.classList.add('show-other-content', 'govuk-link')
     this.showLink.innerHTML = linkText
-    this.showLink.href = '#'
+    this.showLink.setAttribute('aria-expanded', 'false')
+    this.showLink.setAttribute('aria-controls', 'other-content')
```
AshGDS, line 55: Thanks -  though `aria-controls` needs an element ID, rather than a CSS class

AshGDS (changes requested):
Thanks - a small change needed but looking good

app/assets/javascripts/modules/hide-other-links.js
matthillco: Ah, I forgot to add the ID assignment before pushing that up. Fixed now, at line 75.

matthillco: @AshGDS I forgot to tag you, if you could re-check the above that would be great thanks.

AshGDS (approved):

PR #3202, opened 2024-05-28
Fix missing headings and replace subpage breadcrumb with back link

matthillco:
## What

The main content of the page doesn’t have a heading. There is the H1 of the manual header and then no other heading until the footer. Other pages within manuals have a heading before the lead paragraph. [Trello](https://trello.com/c/O2VfbDGm/2622-page-missing-main-heading-m-l)

## Why

These pages currently fail on accessibility and need to be fixed according to the findings of the DAC audit.

## How

* Replaced the landing-page breadcrumb with H2
* Replaced the sub-page breadcrumb with the back link component
* Added a "Contents" H2 above the document list
* Rewrote tests accordingly

## Before/After: Missing Headings

image

## Before After: Replace breadcrumb with back link (Section)

image

## Before After: Replace breadcrumb with back link (Updates)

image

### DIFF
--- a/app/views/content_items/manuals/_manual_layout.html.erb
+++ b/app/views/content_items/manuals/_manual_layout.html.erb
@@ -1,4 +1,8 @@
 <% content_for :main do %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: t("manuals.summary_title"),
+    margin_bottom: 4
+  } %>
   <div id="manuals-frontend" class="manuals-frontend-body">
 
   <%= machine_readable_metadata(schema: :article) %>
@@ -14,6 +18,11 @@
         <%= raw(@content_item.body) %>
       <% end %>
     <% end %>
+
+    <%= render "govuk_publishing_components/components/heading", {
+      text: t("manuals.contents_title"),
+      margin_bottom: 4
+    } %>
 
     <%= yield %>
 
--- a/app/views/content_items/manuals/_manual_section_layout.html.erb
+++ b/app/views/content_items/manuals/_manual_section_layout.html.erb
@@ -1,6 +1,10 @@
 <% show_contents ||= false %>
 
 <% content_for :main do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    text: t("manuals.breadcrumb_contents"),
+    href: @content_item.base_path
+  } %>
   <div id="manuals-frontend" class="manuals-frontend-body">
     <% if show_contents %>
       <%= render "govuk_publishing_components/components/contents_list", {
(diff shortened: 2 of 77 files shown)


KludgeKML (approved):
LGTM!

app/views/content_items/manuals/_manual_layout.html.erb
```diff
@@ -1,4 +1,8 @@
 <% content_for :main do %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: I18n.t("manuals.summary_title"),
```
unoduetre, line 3: I suppose "I18n." is unnecessary here?

app/views/content_items/manuals/_manual_layout.html.erb
```diff
@@ -15,8 +19,13 @@
       <% end %>
     <% end %>
 
+    <%= render "govuk_publishing_components/components/heading", {
+      text: I18n.t("manuals.contents_title"),
```
unoduetre, line 23: Please, see my comment about I18n above.

Also, should a test be written for this one?

app/views/content_items/manuals/_manual_section_layout.html.erb
```diff
@@ -1,6 +1,10 @@
 <% show_contents ||= false %>
 
 <% content_for :main do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    text: I18n.t("manuals.breadcrumb_contents"),
```
unoduetre, line 5: Please, see my comment about I18n above.

app/views/content_items/manuals/_manual_updates_layout.html.erb
```diff
@@ -1,4 +1,8 @@
 <% content_for :main do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    text: I18n.t("manuals.breadcrumb_contents"),
```
unoduetre, line 3: Please, see my comment about I18n above.

unoduetre (commented):
I assumed the style changes have been discussed with the previous reviewer, so I didn't focus on those, but if not, would it make sense to discuss with a frontend dev?

app/views/content_items/manuals/_manual_layout.html.erb
CodeSonia: > Please, see my comment about I18n above.
> 
> Also, should a test be written for this one?



> Please, see my comment about I18n above.
> 
> Also, should a test be written for this one?



> Please, see my comment about I18n above.
> 
> Also, should a test be written for this one?



> Please, see my comment about I18n above.
> 
> Also, should a test be written for this one?

Thanks @unoduetre, I've added a test case for the contents_title and have added that on both hmrc_manual_test.rb & manual_test.rb. Let me know if that's ok?


It looks like the I18n was used previously, not sure if we should leave it just in case if we wanted to add more translation perhaps in the future (looks like it's used elsewhere too). Let me know what you think

unoduetre: What I meant is that the part `I18n.` can be removed, leaving only the `t(...)` part. So instead of writing `I18n.t(xyz)` one can write `t(xyz)`.

unoduetre (approved):
Looks good to me.

PR #3208, opened 2024-06-04
Change publishing outcome pages text

andysellick:

## What
Changes the heading on consultations and calls for evidence pages from 'Download the full outcome' to 'Read the full outcome'.

Example pages:

- https://www.gov.uk/government/consultations/pre-employment-checks-for-health-and-care-volunteers

## Why
Text doesn't always make sense, specifically because some of the things being linked to are web pages, and some are PDFs (which for the most part just open in the browser rather than being downloaded).

Note that this change has been made only in the English translation for these pages. This is partly because getting all the translations is hard and not having them in one language shouldn't block the change for other languages, and also because we think there aren't any calls for evidence not in English. There are potentially some consultations in other languages though, so more work may need to be done.

## Visual changes

Before | After
------ | ------
Screenshot 2024-06-04 at 09 50 42 | Screenshot 2024-06-04 at 09 50 48

Trello card: https://trello.com/c/GfUczqKP/733-request-to-change-terminology-on-publishing-outcome-now-pages-are-in-html

### DIFF
--- a/app/presenters/consultation_presenter.rb
+++ b/app/presenters/consultation_presenter.rb
@@ -52,7 +52,7 @@ class ConsultationPresenter < ContentItemPresenter
     content_item["details"]["final_outcome_detail"]
   end
 
-  # Download the full outcome, top of page
+  # Read the full outcome, top of page
   def final_outcome_attachments_for_components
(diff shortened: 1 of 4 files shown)


hannako (approved):
Thanks Andy!

PR #3219, opened 2024-06-07
Use components for calls for evidence

andysellick:

## What
Output attachments in call for evidence pages using the data from the content item, rather than the pre-rendered markup from the content item, and apply GA4 tracking to those attachments that contain Details elements.

This follows on from the work to convert publications in the same manner: https://github.com/alphagov/government-frontend/pull/3058

Attachments on consultation pages can occur in two places, so this is a bit more complex than publications, although some of the code and tests have been copied from that previous work.

Example pages to test with:

- [open call for evidence](https://www.gov.uk/government/calls-for-evidence/registering-historic-classic-rebuilt-vehicles-and-vehicles-converted-to-electric-call-for-evidence)
- [call for evidence outcome](https://www.gov.uk/government/calls-for-evidence/future-policy-framework-for-power-with-carbon-capture-usage-and-storage-ccus-call-for-evidence)
- [closed call for evidence](https://www.gov.uk/government/calls-for-evidence/restricting-the-generation-of-surplus-funds-from-traffic-contraventions)

## Why
This work is for two reasons:

- to convert the rendering of attachments to using components so that tracking can be added to Details elements within attachments (for the 'Request an accessible format' element)
- to generally move away from the use of pre-rendered markup passed through content items

## Visual changes
None, except PDF custom thumbnails are intentionally no longer shown.


KludgeKML (approved):
LGTM

PR #3269, opened 2024-07-17
Do not attempt to show siblings if there aren't any

brucebolt:
When a parent manual has been withdrawn and redirected, we don't get the value for `child_section_groups` as the redirect content item doesn't contain this value.

Therefore not showing any siblings items at all when this is the case, as we can't be sure what the correct parent actually is for this manual section.

This also resolves an issue where there are no breadcrumbs given for an item.

[Trello card](https://trello.com/c/LXBvfN2L)


beccapearce (approved):
This looks good to me! Thank you for doing it Bruce! 🤩

PR #3270, opened 2024-07-17
Use type=module for application.js

AshGDS:
## What / Why
- Use type=module for this application's JS file
- To coincide with https://github.com/alphagov/govuk_publishing_components/pull/4111
- Has [DO NOT MERGE] as we want the above PR live on `static` first before this is merged.
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- To accurately test this you will need to use local static that is using a local version the `govuk_publishing_components` branch linked above. If you test without this, this app's `application.js` seems to crash.
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers

### DIFF
--- a/app/views/shared/_webchat.html.erb
+++ b/app/views/shared/_webchat.html.erb
@@ -34,4 +34,4 @@
   </span>
 </p>
 <% # This is inline in the source however slimmer will optimize this. %>
-<%= javascript_include_tag "webchat", integrity: false %>
+<%= javascript_include_tag "webchat", integrity: false, type: "module" %>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,7 +164,7 @@ GEM
     govuk_personalisation (0.16.0)
       plek (>= 1.9.0)
       rails (>= 6, < 8)
-    govuk_publishing_components (41.1.1)
+    govuk_publishing_components (42.0.0)
       govuk_app_config
       govuk_personalisation (>= 0.7.0)
       kramdown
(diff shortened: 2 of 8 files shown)


app/views/shared/_webchat.html.erb
```diff
@@ -34,4 +34,4 @@
   </span>
 </p>
 <% # This is inline in the source however slimmer will optimize this. %>
-<%= javascript_include_tag "webchat", integrity: false %>
+<%= javascript_include_tag "webchat", integrity: false, type: "module" %>
```
MartinJJones, line 37: Looks good to me as well, I tested one of the pages where webchat is used - https://government-frontend-pr-3282.herokuapp.com/government/organisations/hm-passport-office/contact/hm-passport-office-webchat

Still functions as expected and is included in the standardx checks.

MartinJJones (approved):
Changes look good to me 👍

MartinJJones (commented):
The code changes look to me 👍 

The JS error in the browser relates to the closest polyfill, further info here - https://github.com/alphagov/smart-answers/pull/6846#pullrequestreview-2198973334

I'm happy to give this another look and approve when the new version of the gem is released and included in this PR as this should then resolve the issue.

MartinJJones (changes requested):
The code changes look to me 👍 

The JS error in the browser relates to the closest polyfill, further info here - https://github.com/alphagov/smart-answers/pull/6846#pullrequestreview-2198973334

I'm happy to give this another look and approve when the new version of the gem is released and included in this PR as this should then resolve the issue.

---
Used request changes to avoid anyone helpfully merging

MartinJJones (approved):

PR #3278, opened 2024-07-22
Remove org sort by title from html publications

KludgeKML:
We're currently sorting organisations by their title for the header, but this erases information from the content item, namely the ordering of organisations. Until now that hasn't been useful information (because it was the database natural ordering of the organisations involved). But now it's the lead organisations followed by the supporting organisations, so we shouldn't resort. This will bring the order of organisations in HTML attachments into line with the order presented in the item they're attached to.

Before Merging:

- [x] Ensure that the [republish task](https://docs.publishing.service.gov.uk/manual/republishing-content.html#to-republish-all-documents-of-a-specific-type) has been run for: Consultations, and Calls for Evidence in production.

https://trello.com/c/7d3ERApJ/23-lead-organisation-logos-in-html-publications

### DIFF
--- a/app/presenters/html_publication_presenter.rb
+++ b/app/presenters/html_publication_presenter.rb
@@ -33,8 +33,7 @@ class HtmlPublicationPresenter < ContentItemPresenter
   end
 
   def organisations
-    orgs = content_item["links"]["organisations"] || []
-    orgs.sort_by { |o| o["title"] }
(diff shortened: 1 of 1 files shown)


beccapearce (approved):
Looks good! Thanks for doing it and thanks for the clear PR description! 😄

PR #3296, opened 2024-08-09
Align phase banner on service manual homepage

MartinJJones:
## What
- Align phase banner on service manual homepage
- Add new `service_manual_homepage?` presenter method

## Why
When on the service_manual_homepage, the phase banner will be wrapped in a div element with the `govuk-width-container` class to align it with the rest of the content on the page.

The service manual homepage layout is different from the rest of the service manual as it contains an app-hero banner that needs to span the full width of the page with a blue background, but we still want the phase banner to be contained.

### Service manual homepage

| Before | After |
| --- | --- |
| Screenshot 2024-08-09 at 13 56 22 | Screenshot 2024-08-09 at 13 56 58 |

### DIFF
--- a/app/presenters/content_item_presenter.rb
+++ b/app/presenters/content_item_presenter.rb
@@ -97,6 +97,10 @@ class ContentItemPresenter
     false
   end
 
+  def service_manual_homepage?
+    false
+  end
+
   def render_guide_as_single_page?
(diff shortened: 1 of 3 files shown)


CodeSonia (approved):
Thanks for fixing this 🌟

PR #3303, opened 2024-08-19
Display GOV.UK Chat promo on certain pages

jackbot:
**Note: this is a draft and an incomplete implementation. I'm just looking for feedback for the approach taken.**

We want to show a promo to the new GOV.UK Chat application on certain pages on GOV.UK. It's a CTA pointing users to the chat app.

The list of URLs we want the promo to appear on is quite short and cover only `answer` and `guide` content types from this codebase.

The CTA is to be shown in the sidebar on the right of the relevant pages.

Screenshot 2024-08-27 at 09 49 23

### DIFF
--- a/app/views/shared/_sidebar_navigation.html.erb
+++ b/app/views/shared/_sidebar_navigation.html.erb
@@ -1,5 +1,9 @@
 <% content_item = @content_item.content_item.parsed_content %>
 
 <div class="govuk-grid-column-one-third">
+  <% if show_govuk_chat_promo?(@content_item.base_path) %>
+    <%= render "govuk_publishing_components/components/chat_entry", { border_top: true } %>
+  <% end %>
+
   <%= render 'govuk_publishing_components/components/contextual_sidebar', content_item: content_item %>
 </div>
--- a/app/helpers/govuk_chat_promo_helper.rb
+++ b/app/helpers/govuk_chat_promo_helper.rb
@@ -1,0 +1,17 @@
+module GovukChatPromoHelper
+  GOVUK_CHAT_PROMO_BASE_PATHS = %w[
+    /business-support-helpline
+    /company-tax-returns
+    /corporation-tax
+    /pay-corporation-tax
+    /pay-vat
+    /search-for-trademark
+    /set-up-business
+    /submit-vat-return
+    /write-business-plan
+  ].freeze
+
+  def show_govuk_chat_promo?(base_path)
+    ENV["GOVUK_CHAT_PROMO_ENABLED"] == "true" && GOVUK_CHAT_PROMO_BASE_PATHS.include?(base_path)
+  end
+end
--- a/test/helpers/govuk_chat_promo_helper_test.rb
+++ b/test/helpers/govuk_chat_promo_helper_test.rb
@@ -1,0 +1,19 @@
+require "test_helper"
+
+class GovukChatPromoHelperTest < ActionView::TestCase
+  test "show_govuk_chat_promo? when configuration disabled" do
+    assert_not show_govuk_chat_promo?(GOVUK_CHAT_PROMO_BASE_PATHS.first)
+  end
+
+  test "show_govuk_chat_promo? when base_path not in configuration" do
+    ClimateControl.modify GOVUK_CHAT_PROMO_ENABLED: "true" do
+      assert_not show_govuk_chat_promo?("/non-matching-path")
+    end
+  end
+
+  test "show_govuk_chat_promo? when base_path is in configuration" do
+    ClimateControl.modify GOVUK_CHAT_PROMO_ENABLED: "true" do
+      assert show_govuk_chat_promo?(GOVUK_CHAT_PROMO_BASE_PATHS.first)
+    end
+  end
+end
--- a/test/integration/govuk_chat_promo_test.rb
+++ b/test/integration/govuk_chat_promo_test.rb
@@ -1,0 +1,15 @@
+require "test_helper"
+
+class GovukChatPromoTest < ActionDispatch::IntegrationTest
+  test "renders GOV.UK chat promo for matching content type and base path" do
+    ClimateControl.modify GOVUK_CHAT_PROMO_ENABLED: "true" do
+      setup_and_visit_a_page_with_specific_base_path("answer", GovukChatPromoHelper::GOVUK_CHAT_PROMO_BASE_PATHS.first)
+
+      assert page.has_css?(".gem-c-chat-entry")
+    end
+  end
+
+  def schema_type
+    "answer"
+  end
+end


config/application.rb
```diff
@@ -115,6 +115,18 @@ class Application < Rails::Application
     # Path within public/ where assets are compiled to
     config.assets.prefix = "/assets/government-frontend"
 
+    config.govuk_chat_promo_base_paths = %w[
```
kevindew, line 118: I think it'd be good to also have a config option here about whether it is enabled that is controlled by a env var. Then we could merge and ship without the env var in place and just need to use that to turn it on and off in production.

app/views/shared/_sidebar_navigation.html.erb
```diff
@@ -1,5 +1,7 @@
 <% content_item = @content_item.content_item.parsed_content %>
 
 <div class="govuk-grid-column-one-third">
+  <%= yield :govuk_chat_promo %>
```
kevindew, line 4: Is the banner consistently here in also usages of this app? (I have some recollection of other placings but they may all be in collections)

kevindew (commented):
Approach looks great to me, looks like @jon-kirwan isn't far away from merging the component in so that we can see the final result: https://github.com/alphagov/govuk_publishing_components/pull/4151

Not sure if you and him want to converge to try out his component before he merges it to rule out any issues?

config/application.rb
jackbot: Good call; I've made it configurable with an env var.

app/views/shared/_sidebar_navigation.html.erb
jackbot: Yeah it's in the same place on both content type pages. The last 2 screenshots in the description show the placement on an answer and a guide page respectively.

kevindew:
Nice one, this looks good enough to me that we could proceed to write tests for it.

test/helpers/govuk_chat_promo_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "test_helper"
+
+class GovukChatPromoHelperTest < ActionView::TestCase
```
jackbot, line 3: This test covers the two early-returns in the helper method, but not the actual render method, which I'm struggling to get working as I'm not very familiar with Minitest.

It seems that if the test actually calls the `render` method in the helper, an exception is thrown:

```
Minitest::UnexpectedError:         NoMethodError: undefined method `form_with_generates_remote_forms=' for ActionView::Base:Class
            app/helpers/govuk_chat_promo_helper.rb:9:in `render_govuk_chat_promo'
            test/helpers/govuk_chat_promo_helper_spec.rb:34:in `block (2 levels) in <class:GovukChatPromoHelperTest>'
            test/support/govuk_chat_promo.rb:6:in `enable_govuk_chat_promo'
            test/helpers/govuk_chat_promo_helper_spec.rb:24:in `block in <class:GovukChatPromoHelperTest>'
```

test/support/govuk_chat_promo.rb
```diff
@@ -0,0 +1,11 @@
+module GovukChatPromo
```
jackbot, line 1: I tried using CodeClimate to set this env var, but because we're using that env var to set a Rails config variable, CodeClimate will only work if it's used in a `setup` / `teardown` combo. If you just try to set the variable within the test itself, the Rails app has already been loaded and so the change to the env var doesn't then set the config.

AFAICT, you can only have one `setup` / `teardown` call per test class, so it's tricky to test both behaviours (promo enabled / promo disabled) in one test class. An option would be to define multiple classes in the test file, but I couldn't see anywhere else in the codebase where that was done, so felt a bit off to me.

Another option would be to define multiple test files each containing a single class, but that felt like a lot of new files just to test the two different behaviours.

Again, my lack of Minitest knowledge doesn't really help me here!

kevindew (commented):
Thanks @jackbot - nice job with the minitest fight. After chatting with @leenagupte I've made a few suggestions that are related to how we could simplify things a bit. They were a little surprised at how many files are affected, which is my fault as I told them it'd not be much more complex than:

```
<% if SHOW_CHAT_PROMO_BASE_PATHS.include?(content_item.base_path) %>
  <%= render "govuk_publishing_components/chat_promo" %>
<% end %>
```

Let me know what you think, but I think in general they'll make it mostly easier to test and reduce places with chat code.

test/integration/answer_test.rb
```diff
@@ -33,4 +33,12 @@ class AnswerTest < ActionDispatch::IntegrationTest
     setup_and_visit_content_item("answer")
     assert_not page.has_css?(".gem-c-single-page-notification-button")
   end
+
+  test "renders GOV.UK chat promo for matching base path" do
```
jackbot, line 37: Wasn't really sure where to put this test now that we could technically render the promo on any content type page now. There doesn't seem to be any "integration test for all content types" test file anywhere, as far as I could see.

test/helpers/govuk_chat_promo_helper_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  test "show_govuk_chat_promo? when configuration disabled" do
+    assert_not show_govuk_chat_promo?(GOVUK_CHAT_PROMO_BASE_PATHS.first)
```
kevindew, line 5: Aside: It surprises me that this constant is in scope, I wonder how that works. I didn't think it'd be included when methods are included.

test/integration/answer_test.rb
kevindew: Yeah it doesn't look like there is one. Closest thing seems to be meta_tags test.

I imagine we've got a choice between

1) no test, consider the integration code sufficiently simple that it doesn't need a test
2) create a simple govuk_chat_promo_test in integration tests.

I think the most responsible thing to do is the latter with us being prepared to remove it if the team don't want it to keep chat surface area smaller.

I guess a thing to consider is that, for when we have the component installed, the assertion will change from `has_text?("Try GOV.UK Chat")` to `has_selector(".gem-c-chat-entry")` so would be a little more contrived as we shouldn't couple the test to the component copy.

kevindew (commented):
Great stuff, just a note on the integration test

test/integration/answer_test.rb
leenagupte: Hello. I agree that 2, creating a separate test file is the best option. Keeping the tests separate will make them easier to remove later without needing to unpick them from the rest of the test suite.

jackbot: Thanks both, I've created a separate test file that just tests the happy path; I think the unit tests for the helper cover all the rest of the paths but let me know if you disagree.

I've also removed the temporary view because we can now render the actual component.

leenagupte:
@jackbot I think it's best to check with the frontend community about how the stacking of components works and the best options for solving this problem. Perhaps @andysellick or @AshGDS could help?

AshGDS:
If I'm understanding the `flex` proposal correctly, would this swap the order of the `Related content` and `Collection` sections on the sidebar? If so, this will impact analytics (which relies on indexes listing what position the sections are in lists)

kevindew:
Thanks @jackbot, so if I understand correctly the designs dictate a different ordering for desktop and mobile. I think it'd be best to check with our designers first on this. Let's first rule out that the changing order isn't a mistake. It seems unnecessarily complex having order different in mobile so I think we should push for having just one that is consistent for both.

AshGDS:
No worries, thanks @jackbot - an issue with ordering elements with `flex` is that the order of the DOM is preserved, which creates an accessibility issue.

See here for more information: https://adropincalm.com/blog/accessibility-issues-using-order-in-grid-and-flex/

leenagupte (commented):
I've added one tiny inline comment about the test file name, but other than that it looks good to me 👍 

I've added the environment variable to the review app in Heroku to try it out.

Strangely enough the first link https://government-frontend-pr-3303.herokuapp.com/business-support-helpline redirects to https://www.gov.uk/business-support-service for me. Not sure what's happening there. 

All the other pages look great though.

test/helpers/govuk_chat_promo_helper_spec.rb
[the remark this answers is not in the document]
jackbot: Oops, good spot! Rspec muscle memory...

leenagupte (approved):
👍

PR #3304, opened 2024-08-20
Add user research banner

hannako:
As requested by MOD via our normal processes
https://trello.com/c/2RgPDQf8/2800-govuk-user-research-banner-requests-mod

|Before|After|
|-|-|
|Screenshot 2024-08-20 at 12 52 38|Screenshot 2024-08-20 at 12 52 06|

Trello: https://trello.com/c/2RgPDQf8/2800-govuk-user-research-banner-requests-mod
Review app: https://government-frontend-pr-3304.herokuapp.com/guidance/medals-campaigns-descriptions-and-eligibility

### DIFF
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -3,6 +3,7 @@
     schema: :faq
   ) %>
 <% end %>
+<%= render 'shared/intervention_banner' %>
 
 <%= render 'shared/email_subscribe_unsubscribe_flash', { title: @content_item.title_and_context[:title] } %>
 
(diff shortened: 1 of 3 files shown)


app/views/content_items/detailed_guide.html.erb
```diff
@@ -3,6 +3,7 @@
     schema: :faq
   ) %>
 <% end %>
+<%= render 'shared/intervention_banner' %>
```
leenagupte, line 6: Should this be in a guard clause? So it doesn't need to be added and removed everytime?
e.g. https://github.com/alphagov/frontend/blob/735e8c9e1d5bcab90994651009437e9127e37b58/app/views/shared/_intervention_banner.html.erb#L1

leenagupte: I think the field values should be added too?

KludgeKML (approved):
LGTM

leenagupte (commented):
A couple of inline comments.

app/views/content_items/detailed_guide.html.erb
leenagupte: Sorry, I missed that it's in it's own partial 🤦‍♀️

PR #3310, opened 2024-08-27
Replace organisation logos with metadata

hannako:
This commit is a spike to explore how easy/hard it would be to replace organisation logos on HTML attachments with a call to the metadata component instead:
https://components.publishing.service.gov.uk/component-guide/metadata#on_a_dark_background_without_padding

Some organisations have custom logos, but we only display the logo if it will be the only logo on the page. If it will sit alongside other logos, we hide the custom logo. This behaviour is weird and unexpected.

|Before|After|
|-|-|
|Screenshot 2024-08-27 at 17 17 41|Screenshot 2024-08-27 at 17 18 22|

[Live site](https://gov.uk/government/publications/keepers-of-time-ancient-and-native-woodland-and-trees-policy-in-england/keepers-of-time-ancient-and-native-woodland-and-trees-policy-in-england)
[Review app](https://government-frontend-pr-3310.herokuapp.com//government/publications/keepers-of-time-ancient-and-native-woodland-and-trees-policy-in-england/keepers-of-time-ancient-and-native-woodland-and-trees-policy-in-england)

Trello: https://trello.com/c/3jez4Zg9/2799-understand-technical-design-and-content-implications-of-removing-org-logos-from-html-attachments

### DIFF
--- a/app/views/content_items/html_publication.html.erb
+++ b/app/views/content_items/html_publication.html.erb
@@ -9,31 +9,23 @@
   content_for :simple_header, true
 %>
 
-<% if @content_item.organisations %>
-  <div class="publication-external">
-    <ul class="organisation-logos">
-      <% @content_item.organisations.each do |organisation| %>
-        <% logo_attributes = @content_item.organisation_logo(organisation) %>
-        <% next unless logo_attributes %>
-        <li class="organisation-logos__logo">
-          <%= render 'govuk_publishing_components/components/organisation_logo', logo_attributes %>
-        </li>
-      <% end %>
-    </ul>
-  </div>
-<% end %>
 
-<%= render 'govuk_publishing_components/components/inverse_header', {} do %>
-  <%= render 'govuk_publishing_components/components/title',
-        title: @content_item.title,
-        context: I18n.t("content_item.schema_name.#{@content_item.format_sub_type}", count: 1),
+<div class="govuk-!-margin-top-5">
+  <%= render 'govuk_publishing_components/components/inverse_header', {} do %>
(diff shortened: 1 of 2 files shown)


app/views/content_items/html_publication.html.erb
```diff
@@ -9,31 +9,23 @@
...
-  <%= render 'govuk_publishing_components/components/title',
-        title: @content_item.title,
-        context: I18n.t("content_item.schema_name.#{@content_item.format_sub_type}", count: 1),
-        inverse: true,
-        margin_bottom: 0,
-        margin_top: 3
-  %>
-  <p class="publication-header__last-changed"><%= @content_item.last_changed %></p>
-<% end %>
+<div class="govuk-!-margin-top-5">
```
MartinJJones, line 13: Nice, I like the approach of using a utility class from the design system here, avoids adding custom CSS / stylesheet and easy to understand what it does just from looking at the view template 👍

MartinJJones: Just to provide some further info on this, I noticed a similar issue with spacing when reviewing another PR in government-frontend where we also use utility classes to apply spacing to each view template - https://github.com/alphagov/government-frontend/pull/3338

After further investigation I think we can make an improvement in government-frontend by adding a CSS class to the `main` element as recommend by the page layout section of the design system. I have created a Trello card with further info on this and added it to the backlog - https://trello.com/c/Gawqbmsl/1307-page-layout-in-government-frontend-does-not-follow-the-design-system-guidance

hannako:
I don't think we're going to go with this spike, so closing for now

PR #3317, opened 2024-09-02
Update URL in GOV.UK Chat promo helper

jackbot:
The original URL is actually a redirect, so the promo wouldn't render
after the browser followed the redirect. This updates the helper to
specify the actual URL.


leenagupte (approved):
Looks good: https://government-frontend-pr-3317.herokuapp.com/business-support-service 🎉

PR #3331, opened 2024-09-16
[DO NOT MERGE] Testing branch for collapsing columns on print

matthillco:
## DO NOT MERGE 

This is a branch for testing the work in the components gem to add a new print class that collapses the grid columns when printing. This should not be merged as it's only required for local testing and grabbing screenshots.


hannako:
Closing as assume this branch has served its purpose

PR #3337, opened 2024-09-19
Apply 'no columns' print style to Guide/Detailed Guide

matthillco:
## What
Add the new `gem-print-columns-none` print class to the layouts in Guides and Detailed Guides, to collapse columns on print. [Trello](https://trello.com/c/J1FtkA7Q/344-add-generic-print-styles-to-guide-and-detailed-guide)

## Why
When printing various page types, the 2/3rds + 1/3rds column layout causes issues with cropped content and a generally poor printing experience because the full page isn't used. By selectively collapsing floated columns, we can avoid most of these issues and improve the printing experience by using the full width of the page.

### Visual Changes: GUIDE (Portrait)
| Before    | After |
| -------- | ------- |
|image|image|

### Visual Changes: GUIDE (Landscape)
| Before    | After |
| -------- | ------- |
|image|image|

### Visual Changes: DETAILED GUIDE (Portrait)
| Before    | After |
| -------- | ------- |
|image|image|

### Visual Changes: DETAILED GUIDE (Landscape)
| Before    | After |
| -------- | ------- |
|image|image|

### DIFF
--- a/app/views/content_items/detailed_guide.html.erb
+++ b/app/views/content_items/detailed_guide.html.erb
@@ -7,7 +7,7 @@
 
 <%= render 'shared/email_subscribe_unsubscribe_flash', { title: @content_item.title_and_context[:title] } %>
 
-<div class="govuk-grid-row">
+<div class="govuk-grid-row gem-print-columns-none">
   <div class="govuk-grid-column-two-thirds">
     <%= render 'govuk_publishing_components/components/title', @content_item.title_and_context %>
   </div>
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Looks good to me 👍 

Slightly off topic, is the contrast on these numbers good enough for print? 
image

Also why are my links when I print blue, and yours grey? Is that just your printer settings set to greyscale?

matthillco:
> Slightly off topic, is the contrast on these numbers good enough for print?

Nice spot, yep I have this in my backlog of other fixes that still need doing, this one slipped through the net.

> Also why are my links when I print blue, and yours grey? Is that just your printer settings set to greyscale?

That's curious, links should be set to black when printing, we should compare our setups to find out the differences.

PR #3338, opened 2024-09-19
Fix breadcrumbs WCAG fails in government-frontend (move breadcrumbs out of main)

unoduetre:

## What

What

goverment-frontend places breadcrumbs inside the <main> container in some places, creating a WCAG fail. I’ve investigated all the templates in the repo and this problem is present in [all the page templates that import the breadcrumbs](https://github.com/search?q=repo%3Aalphagov%2Fgovernment-frontend+breadcrumbs+language%3AHTML%2BERB&type=code&l=HTML%2BERB), apart from /app/views/layouts/application.html.erb:

* app/views/histories/1_horse_guards_road.html.erb
* app/views/histories/11_downing_street.html.erb
* app/views/histories/king_charles_street.html.erb
* app/views/histories/10_downing_street.html.erb
* app/views/histories/lancaster_house.html.erb
* app/views/content_items/service_manual_service_standard.html.erb; example page: /service-manual/service-standard
* app/views/content_items/service_manual_guide.html.erb; example page: /service-manual/measuring-success/using-data-to-improve-your-service-an-introduction
* app/views/content_items/service_manual_topic.html.erb; example page: /service-manual/communities

Move the breadcrumbs out of the main container in each of the impacted templates.

There is one more template using the breadcrumbs component: /app/views/content_items/manuals/_manual_section_layout.html.erb
This is used e.g. in 

## Why

Having the breadcrumbs inside the main container is a WCAG fail.

[Trello card](https://trello.com/c/TVbg8g55/2857-fix-breadcrumbs-wcag-fails-in-government-frontend-m-l)


app/views/content_items/service_manual_topic.html.erb
```diff
@@ -5,11 +5,12 @@
...
 
-  <%= render "govuk_publishing_components/components/breadcrumbs", {
-    breadcrumbs: @content_item.breadcrumbs,
-    collapse_on_mobile: true
-  } %>
+<% content_for :header do %>
+  <%= render partial: "content_items/service_manual_breadcrumbs" %>
+<% end %>
 
+<div class="govuk-width-container">
```
unoduetre, line 12: I compared this file to the two similar files: `service_manual_service_standard.html.erb` and `service_manual_guide.html.erb` and it seems this page lacks the govuk-width-container div which is required by the design system. I've added it. This causes a change in the left and right margins, but it makes it consistent with other pages.

app/views/histories/10_downing_street.html.erb
```diff
@@ -1,17 +1,9 @@
+<% content_for :header do %>
```
unoduetre, line 1: This change causes a visual difference, because after moving breadcrumbs out of main, they are in different parts of the page. In the previous version the bottom margin of breadcrumbs collapsed with the top margin of the title. But as those are now separated by many levels of element structure, some elements of which do things like `clear`, the margins cannot collapse anymore.

But this makes it more consistent with the component system, as the margins are now the default margins of the components. 

If we want to stick to the old look, I'd need to apply a custom CSS instead of using component defaults, which defeats the purpose of having the component system.

unoduetre: EDIT:

I've added a separate commit which mitigates the visual changes without creating a custom CSS. Now the changes are only visible on the mobile and the space is slightly smaller, so it takes less space on the mobile screen, and it also keeps the current look without any visual changes on the desktop.

MartinJJones: Good spot on the visual difference, this has highlighted something that we can improve across all templates in `government-frontend` and likely other frontend applications as well.

Looking at the guidance from the Design System, we should add the missing `govuk-main-wrapper` class to the `main` element - https://design-system.service.gov.uk/styles/layout/#add-vertical-space

`whitehall` has a good example which conditionally loads the `govuk-main-wrapper--l` if there are no breadcumbs or back-link component - https://github.com/alphagov/whitehall/blob/e6bdcf39d1a91a0552c5327ccbb42b097aaaa19d/app/views/layouts/design_system.html.erb#L33

This would fix the visual changes mentioned here and avoid the need to add margin-top to the title, but it could cause issues in other templates, so further testing may be required, this may be best done as a separate pull request, I will create a Trello card for this detailing the issue and the fix.

unoduetre: Do you think anything else needs to be done in this PR, or it is fine and the changes you mentioned would go to the PR you mentioned?

unoduetre: I discussed with @MartinJJones and he had a nice idea how to fix this, which I have now applied, so the problem is now resolved.

MartinJJones: Nice work, thanks for spotting the issue and fixing 👍

MartinJJones (commented):
Nice work!, the changes look good to me overall.

I think that we should still move the breadcrumbs component before the `main` element in the [`_main_section_layout` template](https://github.com/alphagov/government-frontend/blob/aeeb9a5da9cca27feb28322bef89f6b6eef8a2ad/app/views/content_items/manuals/_manual_section_layout.html.erb#L4-L14 ), this will follow the design system guidance and fix the accessibility issue by allowing users to jump straight to the main content.

Another issue spotted when reviewing this is that the breadcrumbs also have the same aria-label value, I will create a new Trello card to capture this issue so it can be fixed.

unoduetre:
> I think that we should still move the breadcrumbs component before the `main` element in the [`_main_section_layout` template](https://github.com/alphagov/government-frontend/blob/aeeb9a5da9cca27feb28322bef89f6b6eef8a2ad/app/views/content_items/manuals/_manual_section_layout.html.erb#L4-L14), this will follow the design system guidance and fix the accessibility issue by allowing users to jump straight to the main content.
> 


I've moved the breadcrumbs out of main, so this should now be fixed.

MartinJJones (approved):
Changes look good to me, nice work 👍

PR #3348, opened 2024-09-27
Handle nested divs in first_item_content

CodeSonia:
## What
[Trello card](https://trello.com/c/lT8fhx5Q/2806-display-content-lists-on-publications-even-when-there-are-fewer-than-3-h2s-m-l)

This PR addresses an issue where the content list was not displayed on some publication pages due to the way nested content elements were handled in the `show_contents_list?` method.

## Context:
The `show_contents_list?` method uses various conditions to determine whether to display a content list. These rules were first introduced in PR #719.

Recently, a Zendesk ticket reported that the content list was missing from the [DVSA Earned Recognition page](https://www.gov.uk/government/publications/dvsa-earned-recognition-it-or-software-you-need/vehicle-operator-it-systems-and-software-that-work-with-dvsa-earned-recognition) page, resulting in an empty block of space on the left side of the publication. 

The issue was traced to the method's ability to handle deeply nested content inside `div` elements.

### Debugging:
During investigation, the following issues were identified on that page:
```
- first_item_content.length = 0 was inaccurately evaluated as 0
- first_item_has_long_content? = false
- first_item_content = "" was empty
```

However, this is not accurate, because the content includes multiple headings and subsections:
- An `h2` section for "Driver system providers"
- Multiple `h3` subsections like "A", "B", and "C", each containing `p` elements with contact details, all wrapped inside nested `div` tags.

The `show_contents_list?` method wasn't picking up this content due to the nested structure. As a result, the conditions returned `false`, causing the content list to be omitted.

### Solution:

I updated the `first_item_content` method in `contents_list.rb` to handle nested `divs` and introduced other methods to extract valid elements `p`, `ul`, and `ol` inside a nested `div` element, ensuring that all content is considered.

### Tests:

Several test cases have been added:
1. Test case for long content in nested elements: It shows the content list when the first item contains long content inside nested `div` elements.
2. Test case for short content in nested elements: Ensures that the content list is not shown when the first item contains short content, even if it is nested inside `divs`.
3. Deep div nesting: This tests that the method correctly extracts text from deeply nested divs while ignoring allowed elements

### Testing:

The [page](https://government-frontend-pr-3348.herokuapp.com/government/publications/dvsa-earned-recognition-it-or-software-you-need/vehicle-operator-it-systems-and-software-that-work-with-dvsa-earned-recognition) now correctly renders the content list:
- first_item_content.length: 2817
- first_item_has_long_content? true
- first_item_content now captures the actual content.

## Why
Without this fix, pages with only two content items and no recognised long content could fail to render the content list, leaving an empty whitespace that negatively impacts user experience.

## Visual Changes
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -46,15 +46,33 @@ module ContentItem
 
     def first_item_content
       element = first_item
-      first_item_text = ""
-      allowed_elements = %w[p ul ol]
+      return "" unless element
 
-      until element.name == "h2"
-        first_item_text += element.text if element.name.in?(allowed_elements)
+      content = ""
+
+      until element.nil? || element.name == "h2"
+        content += extract_content_from_element(element)
         element = element.next_element
-        break if element.nil?
       end
-      first_item_text
+
+      content.strip
+    end
+
+    def extract_content_from_element(element)
+      case element.name
+      when "div"
+        extract_nested_content(element)
+      when "p", "ul", "ol"
+        element.text
+      else
+        ""
+      end
+    end
+
+    def extract_nested_content(element)
+      element.children.each_with_object("") do |child, content|
+        content << extract_content_from_element(child)
+      end
     end
 
     def first_item_character_count
--- a/test/presenters/content_item/contents_list_test.rb
+++ b/test/presenters/content_item/contents_list_test.rb
@@ -172,6 +172,42 @@ class ContentItemContentsListTest < ActiveSupport::TestCase
     assert_not @contents_list.show_contents_list?
   end
 
+  test "#show_contents_list? returns true if the first item contains long content from nested elements" do
+    class << @contents_list
+      def body
+        "<h2 id='one'>One</h2>
+        <div>
+          <p>#{Faker::Lorem.characters(number: 200)}</p>
+          <ul>
+            <li>#{Faker::Lorem.characters(number: 150)}</li>
+            <li>#{Faker::Lorem.characters(number: 150)}</li>
+          </ul>
+        </div>
+        <h2 id='two'>Two</h2>
+        "
+      end
+    end
+    assert @contents_list.show_contents_list?
+  end
+
+  test "#show_contents_list? returns false if the first item does not contain long content from nested elements" do
+    class << @contents_list
+      def body
+        "<h2 id='one'>One</h2>
+      <div>
+        <p>#{Faker::Lorem.characters(number: 50)}</p>
+        <ul>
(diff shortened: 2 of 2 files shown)


app/presenters/content_item/contents_list.rb
```diff
@@ -47,14 +47,19 @@ def first_item_has_long_content?
     def first_item_content
       element = first_item
       first_item_text = ""
-      allowed_elements = %w[p ul ol]
-
-      until element.name == "h2"
-        first_item_text += element.text if element.name.in?(allowed_elements)
+      allowed_elements = %w[p ul ol div]
```
unoduetre, line 50: `div` is checked separately in line 53, so adding it here doesn't change anything and is confusing, so it can be removed from this line. So link 55 can be also changed to not exclude it.

app/presenters/content_item/contents_list.rb
```diff
@@ -47,14 +47,19 @@ def first_item_has_long_content?
...
+
+      until element.nil? || element.name == "h2"
+        if element.name == "div"
+          element.children.each do |child|
```
unoduetre, line 54: It only checks a single level of divs, so it won't find paragraphs inside nested divs. It might solve the immediate problem with the particular document, but the problem will reappear when someone adds a document with more deeply nested divs.

CodeSonia: Yeah that makes sense - I'll change this so I look for more deeply nested divs as like you said I only considered the first level 😅

app/presenters/content_item/contents_list.rb
```diff
@@ -46,15 +46,33 @@ def first_item_has_long_content?
...
+      when "div"
+        extract_nested_content(element)
+      when "p", "ul", "ol"
+        element.text
+      else
+        ""
+      end
+    end
+
+    def extract_nested_content(element)
```
hannako, line 72: This is some nice fancy code!! I think we should check that this is behaviour we actually want though first. I'm wondering if it might result in longer than expected contents lists as we nibble through every element on the page. Do you happen to have an example of a page with very nested content so we can look at the before an after?

hannako:
@CodeSonia I think this PR can be closed right? Could you link to the replacement when you close it.

CodeSonia:
> @CodeSonia I think this PR can be closed right? Could you link to the replacement when you close it.

Yes that's correct! Once I've got the tests cases written for doc collections and PR description, I'll close this!

CodeSonia:
This PR can be closed as the tests have been refactored and a new PR has been generated here: https://github.com/alphagov/government-frontend/pull/3361

PR #3314, opened 2024-08-29
Add HMRC's UR banner to required pages

hannako:
As requested by HMRC via our normal processes
https://trello.com/c/1Pln7K5R/2809-govuk-user-research-banner-requests-hmrc-m

Review apps

- https://government-frontend-pr-3314.herokuapp.com/government/collections/tax-compliance-detailed-information
- https://government-frontend-pr-3314.herokuapp.com/government/collections/hm-revenue-and-customs-compliance-checks-factsheets
- https://government-frontend-pr-3314.herokuapp.com/difficulties-paying-hmrc
- https://government-frontend-pr-3314.herokuapp.com/tax-help
- https://government-frontend-pr-3314.herokuapp.com/get-help-hmrc-extra-support
- https://government-frontend-pr-3314.herokuapp.com/guidance/voluntary-and-community-sector-organisations-who-can-give-you-extra-support
- https://government-frontend-pr-3314.herokuapp.com/tax-appeals
- https://government-frontend-pr-3314.herokuapp.com/guidance/how-to-get-a-review-of-an-hmrc-decision
- https://government-frontend-pr-3314.herokuapp.com/guidance/tax-disputes-alternative-dispute-resolution-adr

PR #3351, opened 2024-10-01
Take down HMRC's UR banner

hannako:
Reverts the banner added in https://github.com/alphagov/government-frontend/pull/3314


lib/data/recruitment_banners.yml
```diff
@@ -8,18 +9,3 @@
   #     - /
   #     - /foreign-travel-advice
 
-banners:
```
hannalaakso, line 11: Just wanted to double check whether we'd want to retain the empty `banners` here? It looks like it might have been previously kept there when there were no banners in the file: https://github.com/alphagov/government-frontend/blob/0f3357ee06cf69df2de0366ef30fbe7495ad4db3/lib/data/recruitment_banners.yml#L11

hannalaakso: Oh I think I missed that, apologies! Will approve the PR now.

hannalaakso (approved):

PR #3363, opened 2024-10-14
Minor refactor of contents list logic

hannako:
Pair programming effort with @CodeSonia 

## What

This PR contains a minor reorganisation of the existing code with no user facing changes

The show_contents_list? method is hard to understand, and bugs have surfaced here.

The special logic around showing contents lists on MOJ manual sections only should only be called by the manual section presenter class.


beccapearce (approved):
Looks much better! 🌟

PR #3361, opened 2024-10-11
Display content lists on publications irrespective off h2 count

CodeSonia:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #3366, opened 2024-10-14
Refactor contents list logic

hannako:
The original implementation could result in the `show_contents_list?` method returning nil
rather than true or false. Which is unexpected, and has been a barrier
to changing when and where we display contents lists ([trello](https://trello.com/c/lT8fhx5Q/2806-display-content-lists-on-publications-even-when-there-are-fewer-than-3-h2s-m-l)) because of confusing
test failures (https://github.com/alphagov/government-frontend/pull/3361).

### DIFF
--- a/app/presenters/content_item/contents_list.rb
+++ b/app/presenters/content_item/contents_list.rb
@@ -1,9 +1,9 @@
 module ContentItem
   module ContentsList
-    CHARACTER_LIMIT = 415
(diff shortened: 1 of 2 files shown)


CodeSonia (approved):
This just reads so much better! 🙌

beccapearce (approved):
This all looks good! Easier to get than before

PR #3371, opened 2024-10-17
Handle incorrectly UTF-8 encoded query and cookie url

unoduetre:

## What

Stop sentry alerting on the “invalid byte sequence in UTF-8 errors”.

As discussed with @hannako, we will use a a solution based on an external gem called `rack-utf8_sanitizer` instead of developing our own gem. This solution will then be used in all our apps by copy-pasting the configuration of this gem.

## Why

When security scanning of our site takes place we are inundated with these errors in our slack channel. These errors are not serving any purpose since there is no issue to be solved. Therefore we should remove them to keep all our alerts actionable.

We often get exceptions in Sentry coming from a security checking bot that attempts some attacks by using strings with special characters. All of those "attacks" fail, but they produce exceptions which we would like to ignore, so they do not clutter Sentry. On the other hand we don’t want to ignore exceptions of the same class coming from other sources.

[Trello card](https://trello.com/c/FeOwxE2x/2938-silently-handle-invalid-byte-sequence-in-utf-8-errors-l)


config/application.rb
```diff
@@ -128,5 +129,15 @@ class Application < Rails::Application
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
     # config.active_record.raise_in_transactional_callbacks = true
+
+    # Protect from "invalid byte sequence in UTF-8" errors,
+    # when a query or a cookie is a string with incorrect UTF-8 encoding.
+    config.middleware.insert_before(
+      Rack::Head,
+      Rack::UTF8Sanitizer,
+      sanitizable_content_types: [],
```
hannako, line 138: Why are you doing `sanitizable_content_types: []`? 

Are you trying to clear the [defaults](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L69-L74)? 

If so why? Also I'm not sure setting it to [] will do anything will it??

config/application.rb
```diff
@@ -128,5 +129,15 @@ class Application < Rails::Application
 
     # Do not swallow errors in after_commit/after_rollback callbacks.
     # config.active_record.raise_in_transactional_callbacks = true
+
+    # Protect from "invalid byte sequence in UTF-8" errors,
+    # when a query or a cookie is a string with incorrect UTF-8 encoding.
+    config.middleware.insert_before(
+      Rack::Head,
+      Rack::UTF8Sanitizer,
```
hannako, line 137: I'm not familiar with adding custom middleware, so I'm just 100% going from the [docs](https://github.com/whitequark/rack-utf8_sanitizer?tab=readme-ov-file#strategies) and they seem to recommend a different structure to the code you've got here.  

Specifically it says `Note: The exception handling middleware must be inserted before Rack::UTF8Sanitizer`

lib/sanitiser/strategy.rb
```diff
@@ -0,0 +1,18 @@
+module Sanitiser
+  class Strategy
+    class SanitisingError < StandardError; end
+
+    class << self
+      def call(input, sanitize_null_bytes: false)
```
hannako, line 6: Sorry I don't understand how `sanitize_null_bytes` can ever be set to `true` ?

But can't we just pass in `sanitize_null_bytes: true` in the application config file, as it recommends [here](https://github.com/whitequark/rack-utf8_sanitizer?tab=readme-ov-file#sanitizing-null-bytes) and allow the default error and message to be raised?

lib/sanitiser/strategy.rb
```diff
@@ -0,0 +1,18 @@
...
+        input
+          .force_encoding(Encoding::ASCII_8BIT)
+          .encode!(Encoding::UTF_8)
```
hannako, line 9: Isn't this also just the default?

hannako (commented):
Sorry Mat I'm not really following this code so I've asked a lot of questions!

If you're confident in these settings it might be helpful to split the commits and explain the options you're passing in, with links to documentation.

I'm wondering if some of the complexity you've added is because you're trying to raise a single error type, so you only have to add that single exception in app-config? But looking at https://github.com/alphagov/govuk_app_config/blob/7b8d4536bf8f54a65694d3e17048d77d60355951/lib/govuk_app_config/govuk_error/configuration.rb#L31-L32 I think it's fine to add more than one error type from the same middleware to this file. So keeping the config in the frontend apps simple is maybe more important.

config/application.rb
unoduetre: Yes, it clears the defaults.

I'm preventing [this function](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L117) from running, which would sanitize POST form data if not disabled.

Yes, setting it to [] works. [1](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L19) [2](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L120).

hannako: Thanks for linking to the code. Does government-frontend POST any data?

lib/sanitiser/strategy.rb
unoduetre: [sanitize_null_bytes](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L23) is an option which can be [potentially specified](https://github.com/whitequark/rack-utf8_sanitizer/tree/main?tab=readme-ov-file#sanitizing-null-bytes). I didn't specify it, so it won't be set, but it safer for the strategy to support it to be bullet proof against someone setting this option at some point in the future. The strategy is based on the original strategy from the gem which you can [see here](https://github.com/whitequark/rack-utf8_sanitizer/blob/f2bcf865432a1470c30de3519df780a5b39cf9a2/lib/rack/utf8_sanitizer.rb#L47). I just changed the exception raised.

lib/sanitiser/strategy.rb
unoduetre: Please, see my comment above and the link to the strategy there.

config/application.rb
unoduetre: For government-frontend we only have one POST route: POST /*path(.:format)                 content_items#service_sign_in_options, but there are more in e.g. frontend and I wanted to use the same setup for all appls. I also wanted to apply the minimal change that resolves the issue, and sanitizing POST data adds unrelated additional stuff.

hannako: It seems a bit hacky to pass in `[]` as a work around to prevent sanitize_rack_input from running, don't you think? It could easily stop working, and I'm not sure we'd know. If we really need to skip it, then at the very least I think we need an inline comment to explain this workaround. 

But in https://github.com/alphagov/transition/pull/1618 they've just gone with the default, and that app has lots of POST routes, so is this a valid concern?

unoduetre: It's more of a question about what you want. Form data sanitisation is simply unrelated to the orginal problem, but if you think it's a nice to have feature, I think we can enable it. It's just another change, so maybe I can create a separate PR for removing that line, so the change can be easily reverted just in case anything goes wrong?

hannako: That's a good idea 👍

lib/sanitiser/strategy.rb
hannako: Why have you created this custom of strategy? Is it for the reason I've suggested [here](https://github.com/alphagov/government-frontend/pull/3371#pullrequestreview-2389192308) ?

unoduetre: We need a custom error so we can safely ignore it. The default strategy raises e.g. EncodingError which is [one of a standard ruby exceptions](https://ruby-doc.org/3.3.5/EncodingError.html). We cannot ignore all EncodingErrors, because they can come from different places and for different reasons. So we should only ignore those that relate to UTF-8 encoding issues in query parameters (and as discussed earlier, in post data). We shouldn't ignore all EncodingErrors, as those might be related to some unrelated problems which we will miss, if we ignore all of them.

hannako (approved):
Thanks for explaining Mat!

Shall we put this branch on integration and see if we can trigger some of the new errors in our tech channel before we merge https://github.com/alphagov/govuk_app_config/pull/402 ?

unoduetre:
> Thanks for explaining Mat!
> 
> Shall we put this branch on integration and see if we can trigger some of the new errors in our tech channel before we merge [alphagov/govuk_app_config#402](https://github.com/alphagov/govuk_app_config/pull/402) ?

I've already tested it on integration. Here is the error produced (which will be ignored after the govuk_app_config PR is merged) for the URL: "http://government-frontend.dev.gov.uk/update-uk-visas-immigration-account-details?%AD":
https://govuk.sentry.io/issues/5998976313/events/bd2880d035fb458aa757a3d63606ec4f/?environment=integration&environment=integration-blue-aws&environment=integration-eks&query=SanitisingError&referrer=previous-event&statsPeriod=90d&stream_index=0 

If you want to test more, I can deploy it again.

hannako:
> > Thanks for explaining Mat!
> > Shall we put this branch on integration and see if we can trigger some of the new errors in our tech channel before we merge [alphagov/govuk_app_config#402](https://github.com/alphagov/govuk_app_config/pull/402) ?
> 
> I've already tested it on integration. Here is the error produced (which will be ignored after the govuk_app_config PR is merged) for the URL: "https://integration.publishing.service.gov.uk/update-uk-visas-immigration-account-details?%AD": https://govuk.sentry.io/issues/5998976313/events/bd2880d035fb458aa757a3d63606ec4f/?environment=integration&environment=integration-blue-aws&environment=integration-eks&query=SanitisingError&referrer=previous-event&statsPeriod=90d&stream_index=0
> 
> If you want to test more, I can deploy it again.

Great! No need to test any more feel free to merge tomorrow 👍

unoduetre:
I need to recreate this PR as there are some issues with the tests.

PR #3404, opened 2024-11-04
Improve print styles for right-to-left (RTL) content.

matthillco:
## What
Improvements to components and layouts so that right-to-left (RTL) content renders correctly when printing*. This is part of the work to improve print styles more generally. [Trello](https://trello.com/c/6E5UoTiK/335-rtl-print-styles-have-been-considered-and-implemented-where-required)

_Note:_ This PR depends on https://github.com/alphagov/govuk_publishing_components/pull/4365 and https://github.com/alphagov/collections/pull/3835 and the _After_ screenshots below include those dependecies.

_* Although this PR is specifically about print styles, some fixes also apply to the screen rendering due to problems originating in the screen styles._

## Why
Some pages are broken for foreign languages that read right to left, with styles not set correctly for the changed direction.

### Publications Page
| Before    | After |
| -------- | ------- |
| image | image  |

### News Article
| Before    | After |
| -------- | ------- |
| image | image |

### Speech
| Before    | After |
| -------- | ------- |
| image | image |

### DIFF
--- a/app/assets/stylesheets/components/_figure.scss
+++ b/app/assets/stylesheets/components/_figure.scss
@@ -17,7 +17,7 @@
   width: 100%;
 
   @include govuk-media-query($from: tablet) {
-    float: left;
+    float: inline-start;
     width: 50%;
   }
 }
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
LGTM 👍

PR #3409, opened 2024-11-08
Unsubscribe button on document collection pages

hannako:
## What

Fix two issues with the email subscription signup process on Document Collection pages.

#### [First commit](https://github.com/alphagov/government-frontend/pull/3409/commits/dbb589261a7aded797c4888cc59127b42f5cd3c5): Display a flash message to confirm subscription. 

This should have been added in https://github.com/alphagov/government-frontend/pull/2535

|Before|After|
|-|-|
|Screenshot 2024-11-12 at 22 28 09|Screenshot 2024-11-12 at 22 26 38|

#### [Second commit](https://github.com/alphagov/government-frontend/pull/3409/commits/c53a9076621824936caea4c7c779df23a2706851): Fix unsubscribe button

Allow users with a gov.uk account to unsubscribe from a document collection via the single page notification button. See commit message for full explanation and links to documentation. 

[Trello](https://trello.com/c/qCUfi1lj/2917-bug-single-page-notification-button-unsubscribe-button-broken-on-document-collections-for-users-with-a-govuk-account)

### DIFF
--- a/app/views/content_items/document_collection.html.erb
+++ b/app/views/content_items/document_collection.html.erb
@@ -3,6 +3,9 @@
     schema: :article
   ) %>
 <% end %>
+
+<%= render 'shared/email_subscribe_unsubscribe_flash', { title: @content_item.title_and_context[:title] } %>
+
 <%= render 'shared/intervention_banner' %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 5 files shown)


KludgeKML:
Commit 3: I kind of forget what the equivalence is, but I think controller tests are the olde-world version of request tests, which is probably reasonable for this test case (we're not testing the actual user interaction in the test, which would make it an integration/system test, just what is returned for a specific request - ie given an account or not, does the signup control in the returned page point to the account or magic link endpoint?)

KludgeKML (approved):
LGTM

PR #3414, opened 2024-11-11
Remove important metadata component

jon-kirwan:
## What

Remove important metadata component

## Why

It can be replaced with the [metadata block](https://components.publishing.service.gov.uk/component-guide/metadata) component.

## Anything else

See https://github.com/alphagov/govuk_publishing_components/pull/4383


AshGDS (approved):
LGTM - and thanks for providing example URLs 👍 I assume the `important-metadata` class is needed for testing?

jon-kirwan:
> LGTM - and thanks for providing example URLs 👍 I assume the `important-metadata` class is needed for testing?

Thanks for the review, @AshGDS! That's right. Since there are often two metadata blocks on the same page now - one with 'published' information and the other containing 'important' metadata - the tests need a way to differentiate between them.

PR #3416, opened 2024-11-12
Fix disappearing Youtube video

andysellick:

## What / why
- on this page https://www.gov.uk/government/publications/electronic-travel-authorisation-eta-help-videos/what-is-an-eta-video there is only an embedded video, which does not appear on mobile
- this seems to be due to the `.offset-empty-contents-list` class, which floats the main column right if there is no contents list on the left to push it across
- however below desktop this styling persists, when all the grid columns should have collapsed
- somehow if there is only a Youtube video present in this floated element, it produces an element with zero width, so the video is effectively invisible
- adjusting the class to only apply on desktop seems to fix the issue

Have checked this on some other HTML publications, both with and without contents lists on the left, and the fix seems to not have impacted anything else. I suspect the page where this problem occurs is something of an edge case - if there is any other content with the video it forces a width on the element and all is fine.

## Visual changes

Before | After
------ | ------
Screenshot 2024-11-12 at 11 04 04 | Screenshot 2024-11-12 at 11 04 09


matthillco (approved):
Looks good to me.

PR #3417, opened 2024-11-12
Add flash success banner to document collection pages

hannako:
This was missing from https://github.com/alphagov/government-frontend/pull/2535


hannako:
Incorporated into https://github.com/alphagov/government-frontend/pull/3409

PR #3447, opened 2024-11-28
Render news articles using GraphQL

brucebolt:
This switches the rendering of news articles to use Publishing API's GraphQL endpoint, only when the feature flag is enabled.

The code here (and in the associated Publishing API PR: https://github.com/alphagov/publishing-api/pull/3008) has been written in such a way that the least number of changes have been needed in this app, as this is only a proof of concept for now.  If we decide to use GraphQL for everything going forward, we should revisit the structure of content returned by GraphQL.

Also note that GraphQL powered queries rely on an initial request to Content Store, since we need to know the content type before deciding whether to use GraphQL.

Furthermore it only renders one level of parent taxon.  We will have follow-up work to change how the taxonomy is presented by GraphQL, as it's recursive nature doesn't fit the current implementation.

[Trello card](https://trello.com/c/RMkMvY8b)

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -136,15 +136,45 @@ private
   end
 
   def load_content_item
+    @content_item = if use_graphql?
+                      graphql_response = Services
+                        .publishing_api
+                        .graphql_content_item(Graphql::NewsArticleQuery.new(content_item_path).query)
+
+                      if graphql_response["schema_name"] == "news_article"
+                        PresenterBuilder.new(
+                          graphql_response,
+                          content_item_path,
+                          view_context,
+                        ).presenter
+                      else
+                        load_content_item_from_content_store
+                      end
+                    else
+                      load_content_item_from_content_store
+                    end
+  end
+
+  def load_content_item_from_content_store
     content_item = Services.content_store.content_item(content_item_path)
 
     content_item["links"]["ordered_related_items"] = ordered_related_items(content_item["links"]) if content_item["links"]
 
-    @content_item = PresenterBuilder.new(
+    PresenterBuilder.new(
       content_item,
       content_item_path,
       view_context,
     ).presenter
+  end
+
+  def use_graphql?
+    if params.include?(:graphql) && params[:graphql] == "true"
+      true
+    elsif params.include?(:graphql) && params[:graphql] == "false"
+      false
+    else
+      Features.graphql_feature_enabled?
+    end
   end
 
   def ordered_related_items(links)
--- a/test/fixtures/graphql/news_article.json
+++ b/test/fixtures/graphql/news_article.json
@@ -1,0 +1,74 @@
+{
+  "data":
+  {
+    "edition":
+    {
+      "base_path": "/government/news/announcement",
+      "description": "Summary of the news",
+      "details": {
+        "body": "Some text",
+        "political": "true"
+      },
+      "document_type": "news_story",
+      "first_published_at": "2016-12-25T01:02:03+00:00",
+      "links": {
+        "available_translations": [
+          {
+            "base_path": "/government/news/announcement",
+            "locale": "en"
+          }
+        ],
+        "government": [
+          {
+            "details": {
+              "current": false
+            },
+            "title": "A government"
+          }
+        ],
+        "organisations": [
+          {
+            "base_path": "/organisation-1",
+            "title": "An organisation"
+          }
+        ],
+        "people": [
+          {
+            "base_path": "/person-1",
+            "title": "A person"
+          }
+        ],
+        "taxons": [
+          {
+            "base_path": "/taxon-2",
+            "content_id": "8fa1fa2f-bcce-4cde-98fb-308514c35c63",
+            "title": "Taxon 2",
+            "links": {
+              "parent_taxons": [
+                {
+                  "base_path": "/taxon-1",
+                  "content_id": "9de167ce-6c4f-40b1-a45e-e3160d75556e",
+                  "title": "Taxon 1"
+                }
+              ]
+            }
+          }
+        ],
+        "topical_events": [
+          {
+            "base_path": "/topical-event-1",
+            "title": "A topical event"
+          }
+        ],
+        "world_locations": [
+          {
+            "base_path": "/world-location-1",
+            "title": "A world location"
+          }
+        ]
+      },
+      "schema_name": "news_article",
+      "title": "Generic news article"
+    }
+  }
+}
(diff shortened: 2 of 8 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -136,16 +136,41 @@ def set_guide_draft_access_token
...
     content_item = Services.content_store.content_item(content_item_path)
 
-    content_item["links"]["ordered_related_items"] = ordered_related_items(content_item["links"]) if content_item["links"]
-
-    @content_item = PresenterBuilder.new(
-      content_item,
-      content_item_path,
-      view_context,
-    ).presenter
+    @content_item = if content_item["schema_name"] == "news_article" && (Features.graphql_feature_enabled? || params.include?(:graphql))
```
yndajas, line 153: **thought:** query params

If the query param was `graphql=false` 👀

app/controllers/content_items_controller.rb
```diff
@@ -136,16 +136,41 @@ def set_guide_draft_access_token
...
   end
 
+  class GdsApi::Response
+    def update_response_body
+      updated_response = JSON.parse(@http_response.body).dig("data", "edition").deep_transform_keys(&:underscore)
+      @http_response = RestClient::Response.create(
+        updated_response.to_json,
+        @http_response.net_http_res,
+        @http_response.request,
+      )
```
yndajas, line 146: **question:** documentation

I'm not sure where this instance variable comes from or what the impact of reassigning it is - is there some documentation anywhere/can this be documented? (I can see what you're trying to do, but I'm struggling to say "yes, it clearly does that")

test/fixtures/graphql/news_article.json
```diff
@@ -0,0 +1,72 @@
...
+          {
+            "basePath": "/taxon-1",
+            "contentId": "8fa1fa2f-bcce-4cde-98fb-308514c35c63",
+            "title": "Taxon 1",
+            "links": {
+              "parentTaxons": [
+                {
+                  "basePath": "/taxon-2",
+                  "contentId": "9de167ce-6c4f-40b1-a45e-e3160d75556e",
+                  "title": "Taxon 2"
```
yndajas, line 49: **nitpick:** example data

The parent's number suggests it was created after it's child 🐓🥚

yndajas (commented):
Question about the `@http_response` instance variable; otherwise nothing significant

app/controllers/content_items_controller.rb
```diff
@@ -136,16 +136,41 @@ def set_guide_draft_access_token
     @content_item.draft_access_token = params[:token]
   end
 
+  class GdsApi::Response
+    def update_response_body
```
mike3985, line 140: +1 for making another change to the API adapters to have them do the `dig`ging and keys-transforming for us, if that would enable us to avoid this monkey patch?

app/controllers/content_items_controller.rb
mike3985, line 153: I'd like to see this conditional given a name. The schema_name check might be a separate concept to the combination of feature flag and query param, too.

mike3985: I was thinking about this same thing again the other day.

It was when I was debugging the broken Ministers Index page in integration. The feature flag being on currently means always use GraphQL. The feature flag is on in integration, so I'd've had to switch it off (a helm charts pull request) to test the page with Content Store instead.

Would changing this behaviour so that the feature flag just influenced the default make things too complicated?
* if there's no query param set, check the feature flag
* if query param is set, do what it says

brucebolt: > * if there's no query param set, check the feature flag
> * if query param is set, do what it says

That seems like a good approach to me.

app/controllers/content_items_controller.rb
brucebolt: This is being assigned in `GdsApi::Response`'s [initialiser](https://github.com/alphagov/gds-api-adapters/blob/5bbcd887b054126c28a8584baffed2875ee86d3c/lib/gds_api/response.rb#L98).  This code basically just adds a method to that class.

test/fixtures/graphql/news_article.json
brucebolt: Nice spot.  Have updated this.

app/controllers/content_items_controller.rb
brucebolt: That's a good idea.  I'll look into this.

app/controllers/content_items_controller.rb
brucebolt: I've now moved this out of the application and into gds-api-adapters: https://github.com/alphagov/gds-api-adapters/pull/1310.

app/controllers/content_items_controller.rb
brucebolt: I've moved the feature flag/query string logic into a method caled `serve_response_from_graphql?`.

app/controllers/content_items_controller.rb
```diff
@@ -137,17 +137,55 @@ def set_guide_draft_access_token
...
+                                             .graphql_content_item(Graphql::SchemaNameQuery.new(content_item_path).query)
+
+                      if schema_name_result["schema_name"] == "news_article"
+                        load_content_item_from_graphql
+                      else
+                        load_content_item_from_content_store
+                      end
+                    else
+                      load_content_item_from_content_store
+                    end
```
yndajas, line 152: **thought:** two requests issue

I suppose if we were to switch over to GraphQL for everything, we might be able to remove the schema name query and do what I suggested in the meeting with Richard and Martyn: have a single big query that specifies what we want for each type using the `...Type` syntax. We'd have to weigh up whether one big query is better than one very small one and then the normal schema-specific one

I think this could work with both the current more strongly typed implementation and the Edition god type alternative. With the latter, we could have types that don't themselves define any fields but do at least map document types to GraphQL types, like you've already done for news articles: https://github.com/alphagov/publishing-api/blob/main/app/graphql/types/news_article_type.rb. Then whichever `...Type` section of the query matches would be executed

JonathanHallam: Would it be gross to change this to just be 
``` 
 @content_item = use_graphql? ? load_content_item_from_graphql : load_content_item_from_content_store
 ```
 and then at the top of the `load_content_item_from_graphql` method we call a new method that checks for the content type and either carries on using graphql or just returns `load_content_item_from_content_store`
 
 I know that it hides a bit of functionality but it would make this bit of code look nicer

brucebolt: I've reduced this down to one GraphQL query now.  We'll need to think about the query a bit more once any changes to the typing in Publishing API have been made and when we add GraphQL support for further document types rendered by Government Frontend.

JonathanHallam (approved):
Thanks Bruce

catalinailie (approved):
Looks good to me

PR #3467, opened 2024-12-06
Hide email signup button from pages that do not have an :en locale

hannako:
## What
Hide the single page notification button from pages that are not published in English. 

## Why
GOV.UK does not support email subscriptions to foreign language content. The single page notification button was added to all pages (English and foreign language). The consequence has been users signing up to an email subscription that will [never match a content change](https://github.com/alphagov/email-alert-service/blob/main/email_alert_service/models/major_change_message_processor.rb#L23-L26).

We _could_ make the button work so that clicking the button on eg a welsh page results
in the user signing up to the equivalent content in English. But we decided against this for the following reasons:

1. There is no indication that there is a user need for that behaviour
2. The code to fetch the english base-path would be added to the button component in the
gem. The logic for this button is already very complicated, and adding further complexity
feels like a bad idea.
3. Not all our publishing applications create foreign content with a link back to the english
version. So if the button was added to other content types in the future it wouldn't work as
expected eg https://www.gov.uk/api/content/budd-dal-plant-16-19
4. Hiding the signup button is the same approach used [elsewhere](https://github.com/alphagov/collections/blob/ea307473313aa5138e6db6dd16eb26f1bf453bfd/app/views/world_location_news/show.html.erb#L56-L86).

|Before|After|
|--|--|
|Screenshot 2024-12-06 at 16 50 52|Screenshot 2024-12-06 at 16 50 21|

[Review app](https://government-frontend-pr-3467.herokuapp.com/government/publications/investigation-of-building-control-professionals.cy)

[Trello](https://trello.com/c/24HBupfM/3107-hide-single-page-notification-button-on-welsh-pages)

### DIFF
--- a/app/views/shared/_document_collections_email_signup.html.erb
+++ b/app/views/shared/_document_collections_email_signup.html.erb
@@ -1,4 +1,4 @@
-<% if @content_item.taxonomy_topic_email_override_base_path.present? %>
+<% if @content_item.show_email_signup_link? %>
   <div
     data-module="ga4-link-tracker"
     data-ga4-link='{ "event_name": "navigation", "type": "subscribe", "index_link": 1, "index_total": 1, "section": "Top" }'
--- a/app/views/shared/_document_collections_email_signup.html.erb
+++ b/app/views/shared/_document_collections_email_signup.html.erb
@@ -10,7 +10,7 @@
       margin_bottom: 6
     } %>
   </div>
-<% else %>
+<% elsif @content_item.display_single_page_notification_button? %>
   <%= render 'shared/single_page_notification_button', {
     content_item: @content_item,
     skip_account: @has_govuk_account ? "false" : "true"
--- a/app/views/shared/_single_page_notification_button.html.erb
+++ b/app/views/shared/_single_page_notification_button.html.erb
@@ -1,24 +1,26 @@
-<%
-  default_ga4_data_attributes = {
-    module: "ga4-link-tracker",
-    ga4_link: {
-      event_name: "navigation",
-      type: "subscribe",
-      index_link: 1,
-      index_total: 2,
-      section: "Top"
+<% if @content_item.display_single_page_notification_button? %>
+  <%
+    default_ga4_data_attributes = {
+      module: "ga4-link-tracker",
+      ga4_link: {
+        event_name: "navigation",
+        type: "subscribe",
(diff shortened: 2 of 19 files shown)


app/views/shared/_document_collections_email_signup.html.erb
```diff
@@ -14,5 +14,5 @@
   <%= render 'shared/single_page_notification_button', {
     content_item: @content_item,
     skip_account: @has_govuk_account ? "false" : "true"
-    } %>
+    } if @content_item.display_single_page_notification_button? %>
```
unoduetre, line 17: I suppose this `if ...` can be put in place of `else` in line 13 (as `elsif`) which I think would make it faster to read?

```
<% elsif @content_item.display_single_page_notification_button? %>
```

app/views/shared/_single_page_notification_button.html.erb
```diff
@@ -21,4 +21,4 @@
   margin_bottom: 6,
   button_location: "top",
   skip_account: skip_account,
-} if @content_item.has_single_page_notifications? %>
+} if @content_item.display_single_page_notification_button? %>
```
unoduetre, line 24: I suppose you could put the whole contents of this file inside an `if` instead of using the `if` here, as the lines 1-15 don't need to be executed and I suppose it would make it a bit more readable, as it would be clear from the beginning that those things only need to be executed under the condition?

```ruby
<% if @content_item.display_single_page_notification_button? %>
...
<% end %>
```

test/presenters/publication_presenter_test.rb
```diff
@@ -74,9 +74,11 @@ def schema_name
     assert_equal(presented.national_applicability[:wales][:alternative_url], "http://wales.gov.uk/topics/statistics/headlines/housing2012/121025/?lang=en")
   end
 
-  test "presents the single page notification button" do
-    presented = presented_item("statistics_publication")
-    assert presented.has_single_page_notifications?
+  test "displays the single page notification button on English pages" do
```
unoduetre, line 77: It would be good to also add at least one test somewhere which checks that, if the locale isn't english, the button isn't displayed.
This applies to both the sign-up button and to the single page notification button (so at least 2 tests).

unoduetre (commented):
I've only added some comments which I think would improve the readbility and one comment about adding some tests.

PR #3480, opened 2024-12-11
Remove integration test for the UR banner

hannako:
This test will fail once we pass the [expiration date](https://github.com/alphagov/govuk_web_banners/blob/main/config/govuk_web_banners/recruitment_banners.yml#L46-L47) of the banner in the gem. 
We could add a setup and tear down to set the time to be within the banner
start and end date, but this will only result in a passing test whilst the config remains
in the gem.  If someone removes the config from the gem (clearing up after the banner has expired)
and forgets to remove this test, then deploys of this application would be blocked.
This seems bad, and somewhat defeats the purpose of having the code in a gem.


unoduetre (approved):
Looks good to me.

PR #3500, opened 2024-12-31
Update GraphQL query to use generic edition type

brucebolt:
Publishing API was updated in https://github.com/alphagov/publishing-api/pull/3063 to have a single edition type for most content, instead of a specific type for news articles.

Updating the query used by this application in making such requests to Publishing API.

[Trello card](https://trello.com/c/MSjuzc9M)


JonathanHallam (approved):
Cracking, cheers

PR #3516, opened 2025-01-15
Memoize manual section presenter

richardTowers:
We're currently calling the main method, and then looping over the result. In each loop, we call the method again so that we can get the length. See manual_section.html.erb, which does roughly:

    @content_item.main.each.with_index do |item, index|
     # ...
     {
       data_attributes: {
         ga4_event: {
           # ...
           index: index,
           index_total: @content_item.main.length,

... the effect of this is that we run the main method n times, where n is the number of `<h2>` elements in the body. This results in us spending a lot of time parsing the same document and running the same CSS / XPath selectors on it, which adds up to a lot of time.

Memoizing the result of the main method shaves several seconds off the time to render www.gov.uk/guidance/complete-the-school-census/data-items-2024-to-2025 on my machine.

The flamegraph before these changes (using the wonderful rack-mini-profiler) shows manual_section.html.erb taking over 2 seconds to render:

flamegraph before changes

Note that `ManualSectionPresenter#main` is called many times.

After the changes in this PR, this reduces to 186ms:

flamegraph after changes

These flamegraphs also identified a further ~70ms we could save by memoizing the intro method, so I've done that too.

I did look at some other optimizations (e.g. traversing the nokogiri document manually, instead of using css / xpaths), but once it's memoized those only save microseconds so it's not worth it.

### DIFF
--- a/app/presenters/manual_section_presenter.rb
+++ b/app/presenters/manual_section_presenter.rb
@@ -11,9 +11,9 @@ class ManualSectionPresenter < ContentItemPresenter
   end
 
   def intro
-    @intro ||= begin
-      return nil unless details["body"]
+    return nil unless details["body"]
 
+    @intro ||= begin
       intro = Nokogiri::HTML::DocumentFragment.parse(details["body"])
 
       # Strip all content following and including h2
@@ -29,9 +29,9 @@ class ManualSectionPresenter < ContentItemPresenter
   end
 
   def main
-    @main ||= begin
-      return nil unless details["body"]
+    return nil unless details["body"]
 
+    @main ||= begin
       document = Nokogiri::HTML::DocumentFragment.parse(details["body"])
 
       # Identifies all h2's and creates an array of objects from the heading and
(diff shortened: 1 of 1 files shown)


hannako (approved):
How do you do that rack-mini-profiler flame diagram thing? I smell a tech fortnightly presentation!!

Maybe we could also help by adding 

```
length = @content_item.main.length
```
[here](https://github.com/alphagov/government-frontend/blob/749b29c182ad1a586ef943a358acf13b57afee0a/app/views/content_items/manual_section.html.erb#L40) in the template. Also maybe some of that template logic should be in the presenter? One for another time!

richardTowers:
Definitely a tech fortnightly talk coming yeah. It's pretty easy though actually:

- Install the rack-mini-profiler and stackprof gems
- Disable garbage collection for the controler action in question
- Visit the page and append `?pp=flamegraph` to the URL

https://github.com/alphagov/government-frontend/commit/9d87d6ad6d979b27eb928cf048131f4fe2842115

Previously I've used rbspy for these things, which is way more faff to set up.

PR #3537, opened 2025-01-27
Remove case_study doc type

georges1996:

### DIFF
--- a/test/presenters/content_item_presenter_test.rb
+++ b/test/presenters/content_item_presenter_test.rb
@@ -44,19 +44,19 @@ class ContentItemPresenterTest < PresenterTestCase
   end
 
   test "available_translations sorts languages by locale with English first" do
-    translated = govuk_content_schema_example("case_study", "translated")
+    translated = govuk_content_schema_example("detailed_guide", "translated_detailed_guide")
     presenter = create_presenter(ContentItemPresenter, content_item: translated)
-    assert_equal %w[en ar es], (presenter.available_translations.map { |t| t[:locale] })
+    assert_equal %w[en cy], (presenter.available_translations.map { |t| t[:locale] })
   end
 
   test "available_translations returns native locale names using native_language_name_for" do
-    translated = govuk_content_schema_example("case_study", "translated")
+    translated = govuk_content_schema_example("detailed_guide", "translated_detailed_guide")
     presenter = create_presenter(ContentItemPresenter, content_item: translated)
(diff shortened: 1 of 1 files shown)


test/integration/meta_tags_test.rb
```diff
@@ -2,35 +2,35 @@
 
 class MetaTagsTest < ActionDispatch::IntegrationTest
   test "correct meta tags are displayed for pages" do
-    case_study = GovukSchemas::RandomExample.for_schema(frontend_schema: "news_article") do |random|
+    news_article = GovukSchemas::RandomExample.for_schema(frontend_schema: "news_article") do |random|
```
deborahchua, line 5: This should be `detailed_guide` but you've already added this in the removal of news article rendering in https://github.com/alphagov/government-frontend/pull/3618/commits/cbf103ea17f1022e738afb1cd024621a59bb5431 but I'm guessing this PR would be merged first?

test/presenter_test_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 class PresenterTestCase < ActiveSupport::TestCase
   def create_presenter(presenter_class,
-                       content_item: schema_item("case_study"),
+                       content_item: schema_item("news_article"),
```
deborahchua, line 5: This also should be updated to `detailed_guide`

test/controllers/content_items_controller_test.rb
```diff
@@ -219,14 +219,14 @@ class ContentItemsControllerTest < ActionController::TestCase
   end
 
   test "renders translated content items in their locale" do
-    content_item = content_store_has_schema_example("case_study", "translated")
+    content_item = content_store_has_schema_example("news_article", "news_article_news_story_translated_arabic")
```
deborahchua, line 222: This will also need to be updated to a different document type

test/integration/meta_tags_test.rb
KludgeKML: Now that news_article is done, we should probably update here just for these two tests. I'll update it.

test/controllers/content_items_controller_test.rb
KludgeKML: Yeah, I think I'd rather leave this one until I do news_articles. There aren't any more examples with translations, so it would be useful to put if off for another two weeks!

KludgeKML:
Closing as we're just going to archive the app as is now that all document types have been moved away from it.

PR #3538, opened 2025-01-27
Update banner component to reflect current usage

hannalaakso:
## What
Fix some of the tech debt in the banner component and update the docs to reflect how it is now used on gov.uk. 

We decided to remove the large text variant where when only text is provided, the font size gets increased. This is because the variant isn't being used anywhere at the moment. (We're generally trying to reduce complexity in apps, instead of adding further unnecessary options; there's also been some discussion about how the gem inverse header component could in the future be made more flexible and replace this banner component so we want to make that easier where possible instead of bloating the app component.)

The changes are probably easiest to review commit by commit.

There is a minor change to the design regarding the padding (see https://github.com/alphagov/government-frontend/pull/3538/commits/a11a21a54c9f688ed026af8808086750a437ed6a) which has been reviewed by our designer.

## Why
The way the component is being used had drifted over time to be somewhat different from its original purpose. The design had also changed over time as the grid layout mentioned in the docs was no longer part of the component. As a result, the component code contained some redundant variables and the naming of others was not clear.

## Before

Screenshot 2025-01-29 at 19 31 55

## After

Screenshot 2025-02-12 at 17 51 19

Fixes https://trello.com/c/cNgD83QR/3174-update-the-banner-app-component-in-government-frontend

### DIFF
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -17,13 +17,13 @@
   <% if title %>
     <h2 class="app-c-banner__title"><%= title %></h2>
   <% end %>
-  <p class="app-c-banner__desc">
+  <p class="app-c-banner__text">
     <%= text %>
   </p>
 <% end %>
 <%= tag.section(**component_helper.all_attributes) do %>
   <%= content_block %>
   <% if aside %>
-    <p class="app-c-banner__desc"><%= aside %></p>
+    <p class="app-c-banner__text"><%= aside %></p>
   <% end %>
 <% end %>
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -15,7 +15,7 @@
 %>
 <%= tag.section(**component_helper.all_attributes) do %>
   <% if title %>
-    <h2 class="app-c-banner__title"><%= title %></h2>
+    <h2 class="app-c-banner__title govuk-heading-m"><%= title %></h2>
   <% end %>
   <p class="app-c-banner__text">
     <%= text %>
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -13,16 +13,13 @@
   component_helper.add_data_attribute({ ga4_track_links_only: "" })
   component_helper.add_data_attribute({ ga4_link: { event_name: "navigation", type: "callout" } })
 %>
-<% content_block = capture do %>
+<%= tag.section(**component_helper.all_attributes) do %>
   <% if title %>
     <h2 class="app-c-banner__title"><%= title %></h2>
   <% end %>
   <p class="app-c-banner__text">
     <%= text %>
   </p>
-<% end %>
-<%= tag.section(**component_helper.all_attributes) do %>
-  <%= content_block %>
   <% if secondary_text %>
     <p class="app-c-banner__text"><%= secondary_text %></p>
   <% end %>
--- a/app/views/components/_banner.html.erb
+++ b/app/views/components/_banner.html.erb
@@ -7,7 +7,6 @@
 
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("app-c-banner")
-  component_helper.add_class("app-c-banner--aside") if aside
   component_helper.add_aria_attribute({ label: "Notice" })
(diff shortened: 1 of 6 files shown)


MartinJJones (commented):
The changes look good to me overall, nice work!, just one other thing that is worth looking at.

When testing the different variations in the banner app component, the `app-c-banner__text` includes `19px` of top margin, this means that when only text is provide to the component, there is more space at the top than expected.

This could be solved several ways, by removing margin-top in the CSS or adding the `govuk-body` class for example.

app/views/components/_banner.html.erb
```diff
@@ -2,30 +2,25 @@
...
-  component_helper.add_class("app-c-banner--aside") if aside
   component_helper.add_aria_attribute({ label: "Notice" })
   component_helper.set_lang("en")
   component_helper.add_data_attribute({ module: "ga4-link-tracker" })
   component_helper.add_data_attribute({ ga4_track_links_only: "" })
   component_helper.add_data_attribute({ ga4_link: { event_name: "navigation", type: "callout" } })
 %>
-<% content_block = capture do %>
+<%= tag.section(**component_helper.all_attributes) do %>
   <% if title %>
```
MartinJJones, line 17: Just wondering if `title` should be optional 🤔 

At present the `title` attribute appears to always be set for the banner component in the [call_for_evidence](https://github.com/alphagov/government-frontend/blob/8799336fd3c96dac7c7e53cce3cd8a522beb17da/app/views/content_items/call_for_evidence.html.erb#L105) and [consultation](https://github.com/alphagov/government-frontend/blob/8799336fd3c96dac7c7e53cce3cd8a522beb17da/app/views/content_items/consultation.html.erb#L132) view templates.

If it is optional, then we may want to keep one of the examples in the docs where only `text` is provided.

app/views/content_items/consultation.html.erb
```diff
@@ -119,15 +119,15 @@
     <% consultation_desc = capture do %>
       <%= @content_item.description %>
       <% if @content_item.held_on_another_website? %>
-        <p class="govuk-!-margin-top-2">
```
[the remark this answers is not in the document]
hannalaakso: Thanks for the suggestion @MartinJJones. I actually did just that at first but then undid the commit as it looked really odd to be using a component class outside the component in a page template, I thought it might be more confusing than helpful. 

But I can see the upsides of the approach, especially if the component changes in the future like you say. So if you think, coming to this from the outside, that it would make sense then happy to implement it.

app/views/components/_banner.html.erb
hannalaakso: Good point, I was going to on the approach that `title` would be always required as that's how the component is being used right now. But happy to make it optional, I'll add an example for it too 🙂 

> this means that when only text is provide to the component, there is more space at the top than expected.

I think this relates to the same point about right now the title being required. But I'll fix the spacing now that we want to make the title optional so that only text can be passed through 👍

hannalaakso: For anyone else reading, after discussing this IRL, we've decided to actually make the component enforce the current usage more closely by requiring both the title and text attributes to be passed to it. This reflects how the component is currently used. If the component needed to become more flexible and work for more use cases, we should consider moving it to the gem. (We're generally trying to reduce complexity in apps, instead of adding further unnecessary options; there's also been some discussion about how the gem inverse header component could in the future be made more flexible and replace this banner component so we want to make that easier where possible instead of allowing more use cases and potentially bloating the app component further.)

app/views/components/_banner.html.erb
```diff
@@ -2,30 +2,26 @@
...
-    <%= text %>
-  </p>
-<% end %>
-<%= tag.section(**component_helper.all_attributes) do %>
-  <%= content_block %>
-  <% if aside %>
-    <p class="app-c-banner__desc"><%= aside %></p>
+<% if title.present? && text.present? %>
+  <%= tag.section(**component_helper.all_attributes) do %>
+      <h2 class="app-c-banner__title govuk-heading-m"><%= title %></h2>
```
MartinJJones, line 19: Nitpick (non-blocking): looks like the indentation on the h2 element should be aligned with the p element

MartinJJones (approved):
Nice work on updating the component, the code and documentation is a lot easier to follow and definitely reflects how it used 👍

PR #3557, opened 2025-02-13
Replace title component with heading component

JamesCGDS:

## What
This PR replaces instances of the title component with the heading component.

## Why
[Trello card](https://trello.com/c/ELvrt0ra/467-replace-title-with-heading-in-government-frontend)

## Visual changes
None - I tested every page listed in the [README](https://trello.com/c/ELvrt0ra/467-replace-title-with-heading-in-government-frontend) and several variations listed in the government-frontend root. I didn't notice any visual differences.


JamesCGDS:
@KludgeKML I've added you to this PR as I've made a few changes to some presenters - hope that's alright.

app/presenters/field_of_operation_presenter.rb
```diff
@@ -6,7 +6,7 @@ class FieldOfOperationPresenter < ContentItemPresenter
   def title_and_context
```
KludgeKML: I think the change to this file might want to be fixed up with the earlier commit where all the other presenter changes were.

KludgeKML:
I think you should definitely fixup those calls where [:title] rather than [:text] is still used, although it might be a better idea to just use the content_item title, if one is available?

The use points to here - the title is just passed to a GA4 component, so probably the tests don't fail because that component is happy with a nil title?

https://github.com/alphagov/government-frontend/blob/8799336fd3c96dac7c7e53cce3cd8a522beb17da/app/views/shared/_email_subscribe_unsubscribe_flash.html.erb#L13

app/views/content_items/_body_with_related_links.html.erb
```diff
@@ -1,9 +1,13 @@
 <% content_for :simple_header, true %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= render 'govuk_publishing_components/components/title',
-      title: @content_item.title %>
+  <div class="govuk-grid-column-two-thirds responsive-top-margin">
```
andysellick, line 4: A lot of the problems with spacing introduced by switching to the heading component can be fixed by adding the `govuk-main-wrapper` class to the `main` element in the application layout. It's something we've not added for some reason but is supposed to be part of doing Design System layout. Apologies, I should have made you aware of this earlier.

You'll probably need to do something extra to override this class in some instances (for example on HTML publication pages, where the extra spacing isn't needed) but should save having to manually add spacing to each individual page type.

andysellick (changes requested):

AshGDS (commented):
Looks good to me, just had some questions:

- There's a small visual difference between the top spacing changes on some of the pages now, is this expected?
- `responsive-top-margin` is added to the HTML and then removed in future commits - can this be squashed so it's never added to places it gets removed from? (I understand it being separated now makes the PR easier to follow, though I guess if it's never eventually used it should ideally be removed from the commit history)

PR #3565, opened 2025-02-19
Add ability to render sub facets as metadata

minhngocd:
Sub-facets are introduced as part of Nested Facet work on Specialist Publisher. The new structure introduced allows users to define one level of nesting for facet values. These also need to be surfaced as part of metadata.

https://trello.com/c/4RJusHld/3374-nested-facets-add-new-document
https://trello.com/c/OA4zE8m0/3430-ipo-finder-request

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -28,6 +28,7 @@ class SpecialistDocumentPresenter < ContentItemPresenter
     super.tap do |m|
       facets_with_friendly_values.each do |facet|
         m.merge!(facet["name"] => value_or_array_of_values(facet["values"]))
+        m.merge!(facet["sub_facet_name"] => value_or_array_of_values(facet["sub_facet_values"])) if facet["sub_facet_values"].present?
       end
     end
   end
@@ -123,6 +124,11 @@ private
       values = [facet_values[facet["key"]]].compact.flatten
       facet["values"] = friendly_facet_values(facet, values)
 
+      if facet["nested_facet"]
+        sub_facet_values = [facet_values[facet["sub_facet_key"]]].compact.flatten
+        facet["sub_facet_values"] = friendly_sub_facet_text(facet, sub_facet_values)
+      end
+
       facet
     end
   end
@@ -156,6 +162,24 @@ private
                  values,
                  facet["filterable"])
   end
+
+  def friendly_sub_facet_text(facet, sub_facet_values)
+    return sub_facet_values if facet["allowed_values"].blank?
+
+    sub_facet_allowed_values = facet["allowed_values"].map { |main_facet|
+      main_facet["sub_facets"]&.map do |sub_facet|
+        {
+          "label" => [main_facet["label"], sub_facet["label"]].join(" - "),
+          "value" => sub_facet["value"],
+        }
+      end
+    }.compact.flatten
+
+    facet_blocks(facet["sub_facet_name"],
+                 facet["sub_facet_key"],
+                 sub_facet_allowed_values,
+                 sub_facet_values,
+                 facet["filterable"])
   end
 
   # The facet value is hyphenated, map this to the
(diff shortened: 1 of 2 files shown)


ryanb-gds (commented):
As discussed I don't love this - I'd probably be more inclined to duplicate some of the logic rather than try to make the facet blocks function work for both facets and sub-facets. I'm struggling to understand what the code does just from a quick read. But I appreciate that's a subjective call and I'm not very familiar with the code.

So I'm happy enough but we'd better get a second opinion from somebody on the #govuk-navigation-tech team. It might well be that due to their greater familiarity with the code they're able to follow it comfortably.

unoduetre:
I agree with Ryan. The current code isn't ideal. For example all of those facet methods are private, but in the end in `important_metadata` all they do is just constructing a hash of names and values, so it's not necessary to keep separate facet hashes and modify them. I think it would be easier to simply extract keys, values etc. instead.

The concerns (not suggested methods!) are the following:

- extract keys from facets
- extract a list of allowed values and their labels, per key (if there is such a list)
- extract keys from subfacets (the new one)
- ignore keys which do not have a matching value (but always show first published at and internal notes)
- check if the value is allowed in case there is a list of allowed values, and if not, send a notification
- convert values to proper forms (includes substituting the label (or a link) from allowed values for the value in case there is a list of allowed values)

Please, do not treat this as exhaustive, compare with the existing code.

hannako:
Hi Minno - thanks for opening the PR - there are some nice improvements - I like the refactor to create the `friendly_facet_values ` method 🏅 

It looks like Mat has given you some detailed pointers. My main thought was that it might be easier to create a new `sub_facets_with_friendly_values` method rather than trying to extend `facets_with_friendly_values` as that's brought in some complexity that's tricky to follow. 

I got a bit lost with the changes to the `facet_blocks` method, so maybe that change could be split out into a separate commit somehow. 

Happy to take a look again once you're ready.

minhngocd:
@hannako @unoduetre Thanks for the review. 

I've done what @hannako suggested and split the refactoring and the implementation into two separate commits, hopefully this makes it clearer what I'm trying to achieve. 

Regarding implementation and extracting a separate method block to deal only with sub facets, I think would be slightly inefficient. 

Some pseudo code; the data structure we get from the facet schema looks something like this for one facet:
```
facets: [{
   key: main-facet-key
   name: Main Facet
   sub_facet_key: sub-facet-key
   sub_facet_name: Sub Facet
   allowed_values: [{
     label: Main Value 1
     value: main-value-1
   }, {
     label: Main Value 2
     value: main-value-2
     sub_facets: [{
       label: Sub 1
       value: main-value-2-sub-1
     }, {
       label: Sub 2
       value: main-value-2-sub-2
     }]]
   }]
 }]
```

There are other facets that may not be nested. The metadata tagged on the content item: 
```
{ main-facet-key: [main-value-1, main-value-2], sub-facet-key: [main-value-2-sub-2] }
```

When the metadata loads, as the code currently exists, it only loads keys relating to main facets and does not understand sub-facets, therefore values under `sub-facet-key` is dropped. 

We need to check for every facet whether or not it has sub facets, so we can add them to the metadata list, therefore I thought it was appropriate to extract the list of sub-facet keys and values if they exist and append them to the list of metadata. 

@unoduetre I wasn't sure what your suggestions applied to. Happy to discuss further on a call if needed.

minhngocd:
Further to discussions, looking into refactoring this area of the code, but it looks to be a bigger refactoring piece that maybe should not belong in this PR. I'll raise as a separate PR. 

Unsure whether this should be blocking on this piece of functionality going through.

unoduetre:
@minhngocd Could you talk to @hannako about this? Thanks!

hannako:
@ryanb-gds @minhngocd @unoduetre 

Thanks for splitting the commits Minno! And for suggesting some refactoring @unoduetre 

Unfortunately @ryanb-gds I'm also not familiar with this code, so in its current form I'm still struggling to follow the changes enough to approve.

I don't want to be a blocker for the roll out of a bigger piece of work. So I think either split the commits further, and consider duplicating some code for readability to make the current PR easier to review. Or apply the refactors you and Mat chatted about in your call, and then add support for nested facets as a second PR. Whichever is the least work for you is fine with me.

PR #3570, opened 2025-02-24
Revert spacing changes to service manual homepage

MartinJJones:
## What

Revert spacing changes to service manual homepage

## Why

The `govuk-main-wrapper` class adds the correct spacing to the view templates and follows the design system guidance.

When used on the service-manual and service-toolkit homepage, it adds extra unwanted spacing at the top of the blue banner. To fix this, the new `exclude_main_wrapper_class` method has been used to remove the `govuk-main-wrapper` class from these pages.

### After


JamesCGDS (approved):
LGTM, good spot 👍

PR #3610, opened 2025-03-13
Change nested facet lookup logic

lauraghiorghisor-tw:
The backend has been remodelled so that it sends a single facet with a nested structure through (and of type "nested"), as opposed to two distinct type "text" facets marked as "nested". That nested facet logic was originally introduced in commit 3e9785c3 / PR https://github.com/alphagov/government-frontend/pull/3574.

These changes adjust the extraction logic, but preserve the behaviour:
- the labels read "Main Facet - Sub Facet"
- the sub-facet links include the Main Facet value.

The `sub_facet_key` and `sub_facet_name` are still sent along, as well as the `main_facet_label` and `main_facet_label` references in the sub-facet, so the only required change is to duplicate the call to the `friendly_facet_value` for the sub-facet key.

[Trello](https://trello.com/c/EeZOA3iW/3534-spike-replicating-taxons-approach-for-nested-facets)

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -27,11 +27,17 @@ class SpecialistDocumentPresenter < ContentItemPresenter
   def important_metadata
     super.tap do |m|
       facets_with_values.each do |facet|
-        facet_values_for_content_item = [content_item_metadata[facet["key"]]].flatten
-        friendly_facet_values = friendly_facet_values(facet, facet_values_for_content_item)
-        m.merge!(facet["name"] => value_or_array_of_values(friendly_facet_values))
+        m.merge!(friendly_facet_values_for_key("key", facet))
+        m.merge!(friendly_facet_values_for_key("sub_facet_key", facet)) if facet["type"] == "nested"
       end
     end
+  end
+
+  def friendly_facet_values_for_key(key, facet)
+    facet_values_for_content_item = [content_item_metadata[facet[key]]].flatten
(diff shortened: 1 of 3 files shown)


app/presenters/specialist_document_presenter.rb
```diff
@@ -138,14 +144,18 @@ def friendly_facet_date(dates)
...
 
-    values.map { |value| friendly_facet_label(facet, value) }
+    values.map { |value| friendly_facet_label(facet, key, value) }
   end
 
-  def friendly_facet_label(facet, value)
-    allowed_value = facet["allowed_values"].detect { |av| av["value"] == value }
+  def friendly_facet_label(facet, key, value)
+    allowed_value = if key == "key"
+                      facet["allowed_values"].detect { |av| av["value"] == value } if key == "key"
```
minhngocd, line 155: redundant if statement here

lauraghiorghisor-tw: 😶‍🌫️

minhngocd:
Some thoughts about separating sub-facet handling as an entirely separate process, now that the structure allows us to deal with them separately from the top. But since this is temporary code to release nested facets for the IPO finder, and Specialist Document rendering code will soon be removed from government-frontend as it is migrated over to frontend. Not sure we need to put in too much effort in refactoring.

hannako (approved):
Approving on the basis that the code is temporary until specialist document rendering is ported to frontend. If it wasn't, it might have been nice to move the facet parsing into its own class or module! 

Tagging @leenagupte for a heads up.

PR #3639, opened 2025-03-28
Make blue metadata box optional

GDSNewt:
This change will make rendering the blue metadata box at the top of specialist documents dependent on a value defined in the schema of each finder. 

Specialist publisher frontend rendering may be eventually moved to alphagov/frontend, at the moment that only applies to [license specialist documents](https://www.gov.uk/find-licences/licence-to-abstract-and-or-impound-water-northern-ireland). The approach to this feature should be broadly the same, so eventually migrating over there shouldn't be too arduous. 

The tests in this PR are dependent on example schemas in the publishing API repo and so may brake CI until that is merged. 

**Associated PRs**

1.[ Publishing-API](https://github.com/alphagov/publishing-api/pull/3241)
2. [Specialist-Publisher
](https://github.com/alphagov/specialist-publisher/pull/3076)
[Trello](https://trello.com/c/spUCgJtQ/3539-make-it-possible-to-hide-the-blue-box-on-specialist-documents)

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -85,6 +85,10 @@ class SpecialistDocumentPresenter < ContentItemPresenter
     }
   end
 
+  def show_metadata_block?
+    content_item.dig("links", "finder", 0, "details", "show_metadata_block")
+  end
+
 private
 
   def nested_headers
--- a/app/views/content_items/specialist_document.html.erb
+++ b/app/views/content_items/specialist_document.html.erb
@@ -21,7 +21,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if @content_item.important_metadata.any? %>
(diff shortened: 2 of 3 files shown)


app/presenters/specialist_document_presenter.rb
```diff
@@ -85,6 +85,10 @@ def images
     }
   end
 
+  def show_metadata_block
```
ChrisBAshton, line 88: Nitpick: I'd rename this

```suggestion
  def show_metadata_block?
```

ChrisBAshton (approved):
LGTM 👍

app/presenters/specialist_document_presenter.rb
```diff
@@ -85,6 +85,10 @@ def images
     }
   end
 
+  def show_metadata_block?
+    content_item.dig("links", "finder", 0, "details", "show_metadata_block")
```
leenagupte, line 89: We should probably have added this value to the `Finder` model, and then could have referred to it wirh `finder.show_metablock`

PR #3645, opened 2025-04-04
Add conditional check for contents list component

GDSNewt:
This change will make rendering the table of contents component in specialist documents dependent on a value defined in the schema of each finder. 

Specialist publisher frontend rendering may be eventually moved to alphagov/frontend, at the moment that only applies to [license specialist documents](https://www.gov.uk/find-licences/licence-to-abstract-and-or-impound-water-northern-ireland). I have raised a PR for frontend replicating the behaviour of this PR, in anticipation of this migration. 

Tests in this PR depend on example schemas from the Publishing API repository, which may cause CI failures until the related changes are merged.

**Associated PRs**

[Publishing API](https://github.com/alphagov/publishing-api/pull/3265)
[Frontend](https://github.com/alphagov/frontend/pull/4740)
[Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/3083)

[Trello](https://trello.com/c/2eakEGOo/3570-make-it-possible-to-hide-the-table-of-contents-on-specialist-documents)

- [ ] #govuk-patterns-and-pages have been notified of this change

### DIFF
--- a/app/presenters/specialist_document_presenter.rb
+++ b/app/presenters/specialist_document_presenter.rb
@@ -89,6 +89,10 @@ class SpecialistDocumentPresenter < ContentItemPresenter
     content_item.dig("links", "finder", 0, "details", "show_metadata_block")
   end
 
+  def show_table_of_contents_list?
+    content_item.dig("links", "finder", 0, "details", "show_table_of_contents_list")
+  end
+
 private
 
   def nested_headers
--- a/app/views/content_items/specialist_document.html.erb
+++ b/app/views/content_items/specialist_document.html.erb
@@ -31,7 +31,9 @@
       <% end %>
     <% end %>
 
-    <%= render "govuk_publishing_components/components/contents_list_with_body", contents: @content_item.contents do %>
+    <% if @content_item.show_table_of_contents_list? %>
+      <%= render "govuk_publishing_components/components/contents_list_with_body", contents: @content_item.contents do %>
+    <% end %>
 
       <% if @content_item.protected_food_drink_name? %>
         <img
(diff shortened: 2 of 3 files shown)


app/presenters/specialist_document_presenter.rb
```diff
@@ -89,6 +89,10 @@ def show_metadata_block?
     content_item.dig("links", "finder", 0, "details", "show_metadata_block")
   end
 
+  def hide_contents_list?
+    content_item.dig("links", "finder", 0, "details", "hide_contents_list")
```
leenagupte, line 93: Question: why was this defined as "hide_contents_list" and the metadata one defined as "show_metadata_block"? Why not "show_contents_list"?

app/views/content_items/specialist_document.html.erb
```diff
@@ -31,7 +31,9 @@
       <% end %>
     <% end %>
 
-    <%= render "govuk_publishing_components/components/contents_list_with_body", contents: @content_item.contents do %>
+    <% unless @content_item.hide_contents_list? %>
```
leenagupte, line 34: This backwards logic is a bit painful. I think the method should be renamed to `show_contents_list?` so the check can be truthy.

app/presenters/specialist_document_presenter.rb
leenagupte: Regardless I'd change this method to be `show_contents_list?` and return `true` if `content_item.dig("links", "finder", 0, "details", "hide_contents_list")` is `nil` or `false`.

leenagupte (commented):
I think having two attributes that show or hide a part of the page that named inconsistently will get confusing. I've added an inline comment that explains this in more detail.

leenagupte:
@GDSNewt This PR can be closed as specialist document rendering has moved to frontend.

PR #3650, opened 2025-04-08
Hide updates link

beccapearce:
# 📣 On-going work to be aware of

- [x] #govuk-patterns-and-pages have been notified of this change

____

This removes the see all updates link in two scenarios:
1. On stats announcement pages where the announcement has been cancelled
2. On standard content items where there are no previous updates.

We are removing the links since it is strange to have the link if there is no section the page can link to. It could be confusing for the user that the link does nothing and seem like a bug.

[Trello card](https://trello.com/c/2RemO0mt/3403-hide-see-all-updates-unless-there-is-a-change-history)

### DIFF
--- a/app/presenters/content_item/metadata.rb
+++ b/app/presenters/content_item/metadata.rb
@@ -4,15 +4,19 @@ module ContentItem
     include Updatable
 
     def metadata
-      {
-        from:,
-        first_published: published,
-        last_updated: updated,
-        see_updates_link: true,
-        part_of:,
-        direction: text_direction,
-        other: {},
-      }
+      metadata =
+        {
+          from:,
+          first_published: published,
+          last_updated: updated,
+          part_of:,
+          direction: text_direction,
+          other: {},
+        }
+      if has_change_history?
+        metadata[:see_updates_link] = true
+      end
+      metadata
     end
 
     def important_metadata
@@ -20,13 +24,14 @@ module ContentItem
     end
 
     def publisher_metadata
-      metadata = {
-        from:,
-        first_published: published,
-        last_updated: updated,
-      }
+      metadata =
+        {
+          from:,
+          first_published: published,
+          last_updated: updated,
+        }
 
-      unless pending_stats_announcement?
+      if has_change_history?
         metadata[:see_updates_link] = true
       end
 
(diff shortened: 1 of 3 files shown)


app/presenters/content_item/metadata.rb
```diff
@@ -26,7 +26,7 @@ def publisher_metadata
         last_updated: updated,
       }
 
-      unless pending_stats_announcement?
+      unless pending_stats_announcement? || cancelled_stats_announcement?
         metadata[:see_updates_link] = true
```
hannako, line 30: I was assuming we wouldn't check the document type at all anymore and could move the conditional closer to the source. 

ie if `see_updates_link` is true, the [component](https://github.com/alphagov/govuk_publishing_components/blob/5749bd3e2feccac509c978eb2f6e348ab32b5a90/app/views/govuk_publishing_components/components/_metadata.html.erb#L70) just add's an anchor tag to link to an element that has an ID `full-publication-update-history`.

`full-publication-update-history` ID will added to in the [published_dates](https://github.com/alphagov/govuk_publishing_components/blob/5749bd3e2feccac509c978eb2f6e348ab32b5a90/app/views/govuk_publishing_components/components/_published_dates.html.erb#L13) component if `history.any?` is true.

The `history` array is being passed to the component in this application in all these places: https://github.com/search?q=repo%3Aalphagov%2Fgovernment-frontend++history%3A+%40content_item.history&type=code

Here is the `history` method [defined in the Updatable module](https://github.com/alphagov/government-frontend/blob/e547c42d1c93bfc5a6175e519d113fab1c7c05b5/app/presenters/content_item/updatable.rb#L11)

And that's checking what's in `content_item[change_history]` https://github.com/alphagov/government-frontend/blob/e547c42d1c93bfc5a6175e519d113fab1c7c05b5/app/presenters/content_item/updatable.rb#L28

So really the conditional shouldn't be checking document type, it should be checking if there's a `change_history`.

That `change_history` method is already included in this module (As `Updatable` is included) but it's private, so you'll have to tweak it a bit. 


https://github.com/alphagov/government-frontend/blob/e547c42d1c93bfc5a6175e519d113fab1c7c05b5/app/presenters/content_item/updatable.rb#L28

hannako (approved):

PR #3700, opened 2025-06-05
Hardcode serve HMPO webchat contact content item

minhngocd:
Contacts Admin is being retired. This is the only page on there that can not be supported in its new home : Specialist Publisher, because of the javascript webchat that is available on the contact page. 

Checking content store, this also seems to be the only available `contact` schema type that is being served. This is why we are confident in making changes to ContactPresenter without affecting anything else. 

This presenter, along with this whole controller can be deleted once we are able to support webchat config in Specialist Publisher and Frontend, or find a different approach to serve this contact page. 

Once we merge this in we can unpublish it from Contact admin, and publish the special route from Publishing API to serve up this page instead and retire Contacts Admin entirely.

https://trello.com/c/QrArPjwt/3675-move-hm-passport-office-out-of-contacts-admin

# 📣 On-going work to be aware of

- [ ] #govuk-patterns-and-pages have been notified of this change

____


ChrisBAshton (commented):
Keen to get this over the line so that we can retire Contacts Admin - but I'm a little uncomfortable with the idea of a hardcoded content item. I think we can be much bolder in deleting code we're confident is no longer being used, and simplifying the webchat rendering. (Apologies - I know I did say initially that it would be good to change as little of the frontend code as possible to get this working - but deleting hundreds of lines rather than adding hundreds of lines seems like a good improvement in hindsight).

I've put together a POC PR [here](https://github.com/alphagov/government-frontend/pull/3701) to give an idea of where I think this should go 🙏

PR #3727, opened 2025-06-23
Countryside stewardship grants request

GDSNewt:
Necessary to facilitate the change made in this [Specialist Publisher PR](https://github.com/alphagov/specialist-publisher/pull/3147)

[Trello](https://trello.com/c/UXXwOxKQ/3728-specialist-finder-edit-request-countryside-stewardship-grants)


ryanb-gds (approved):
Assuming the test passes when [Publishing API](https://github.com/alphagov/publishing-api/pull/3351) is merged, I think this seems reasonable. Ideally the webchat thing would be a separate PR but appreciate we're up against it here

PR #3815, opened 2025-09-04
[DO NOT MERGE] Remove gone route

AshGDS:
It has been migrated to frontend. I've left the plain 410 response in just in case.

Related to https://github.com/alphagov/frontend/pull/4917 and https://github.com/alphagov/router/pull/582


KludgeKML:
Closing as we're just going to archive the app as is now that all document types have been moved away from it.

